PR #38, opened 2023-05-18
feat: make tasks importable from IPFS

DavidMinarsch:


packages/valory/services/mech/service.yaml
```diff
@@ -84,8 +84,9 @@ type: skill
         use_termination: ${USE_TERMINATION:bool:false}
         validate_timeout: 1205
         agent_mech_contract_address: ${AGENT_MECH_CONTRACT_ADDRESS:str:0xFf82123dFB52ab75C417195c5fDB87630145ae81}
-        openai_api_key: ${OPENAI_API_KEY_0:str:null}
         reset_period_count: ${RESET_PERIOD_COUNT:int:1000}
+        file_hash_to_tools: ${FILE_HASH_TO_TOOLS_1:dict:{}}
```
DavidMinarsch, line 88: How can we specify a default here?

packages/valory/services/mech/service.yaml
```diff
@@ -126,8 +127,9 @@ type: skill
         use_termination: ${USE_TERMINATION:bool:false}
         validate_timeout: 1205
         agent_mech_contract_address: ${AGENT_MECH_CONTRACT_ADDRESS:str:0xFf82123dFB52ab75C417195c5fDB87630145ae81}
-        openai_api_key: ${OPENAI_API_KEY_1:str:null}
         reset_period_count: ${RESET_PERIOD_COUNT:int:1000}
+        file_hash_to_tools: ${FILE_HASH_TO_TOOLS_4:dict:{}}
```
DavidMinarsch, line 131: need to fix index

packages/valory/skills/task_execution_abci/behaviours.py
```diff
@@ -70,26 +73,33 @@ def __init__(self, **kwargs: Any) -> None:
     def async_act(self) -> Generator:
         """Do the act, supporting asynchronous execution."""
 
+        if not self.context.params.all_tools:
+            all_tools = {}
+            for file_hash, tools in self.context.params.file_hash_to_tools:
+                tool_py = yield from self.get_from_ipfs(file_hash)
+                if tool_py is None:
+                    self.context.logger.error(f"Failed to get the tool {tool} with file_hash {file_hash} from IPFS!")
```
DavidMinarsch, line 81:
```suggestion
                    self.context.logger.error(f"Failed to get the tools {tools} with file_hash {file_hash} from IPFS!")
```

packages/valory/services/mech/service.yaml
0xArdi: Having this as a dict here might not work. A list of lists approach might have to be used.

packages/valory/skills/task_execution_abci/skill.yaml
```diff
@@ -120,6 +119,14 @@ models:
       tendermint_max_retries: 5
       tendermint_p2p_url: localhost:26656
       tendermint_url: http://localhost:26657
+      file_hash_to_tools:
+        xyz:
```
0xArdi, line 123: if you would later like to use a key that is not xyz, I think the framework will complain

packages/valory/services/mech/service.yaml
0xArdi: any reason why these would be different for different agents?

DavidMinarsch: No, actually no

packages/valory/services/mech/service.yaml
DavidMinarsch: Fixed

packages/valory/skills/task_execution_abci/skill.yaml
DavidMinarsch: Fixed

packages/valory/skills/task_execution_abci/models.py
```diff
@@ -44,9 +45,13 @@ class Params(BaseParams):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         """Initialize the parameters object."""
 
-        self.openai_api_key = self._ensure("openai_api_key", kwargs, str)
+        self.api_keys: Dict[str, str] = json.loads(self._ensure("api_keys", kwargs, str))
```
0xArdi, line 48:
```suggestion
        self.api_keys: Dict[str, str] = json.loads(self._ensure("api_keys_json", kwargs, str))
```

packages/valory/skills/task_execution_abci/models.py
```diff
@@ -44,9 +45,13 @@ class Params(BaseParams):
...
+        self.file_hash_to_tools: Dict[str, List[str]] = json.loads(self._ensure("file_hash_to_tools", kwargs, str))
```
0xArdi, line 49:
```suggestion
        self.file_hash_to_tools: Dict[str, List[str]] = json.loads(self._ensure("file_hash_to_tools_json", kwargs, str))
```

packages/packages.json
```diff
@@ -1,14 +1,14 @@
 {
     "dev": {
-        "connection/eightballer/websocket_client/0.1.0": "bafybeicqkljjzvdl4v7pzbmsfd3al7uv7563x3qa5gmnzxeuoxgtmolpta",
```
DavidMinarsch, line 3: We made substantial alterations so making it a Valory package and keeping copyright notice of eightballer intact.

Adamantios (approved):

PR #40, opened 2023-05-19
Fix resubmit logic

Adamantios:
- Adds missing dependencies (90977ce364cb2967c10dbb77989ff5bcdbf76573).
- Adds missing required ledger cosmos for p2p connection (1880fa75715b16de43a5945dc3b166c9b4d0c513).
- Performs a minor fix. Basically, all the reconnect retries were being used even if the connection was established on the first try (55b15a5f2878f8a89367515f3c318801edd4a52a).
- Makes the `SubscriptionBehaviour` a `SimpleBehaviour` instead of a `OneShotBehaviour` in order to be able to re-subscribe (4be364e63ba6543d17e77dc254aa1ff6dbc05d5f).
- Adds re-subscription in a hacky manner on the behaviour side (4be364e63ba6543d17e77dc254aa1ff6dbc05d5f).
- Adds logic to make sure that we look for the missed blocks too when re-subscribing (4be364e63ba6543d17e77dc254aa1ff6dbc05d5f && f0285a5b44baae7d0328fab5928673e2939fd94b).

Next steps:
- [x] Figure out the [TODO](https://github.com/valory-xyz/mech/blob/4be364e63ba6543d17e77dc254aa1ff6dbc05d5f/packages/eightballer/skills/contract_subscription/behaviours.py#L92-L103). -> The easiest way to do this would be to keep track of the block number that is returned from `eth_getTransactionReceipt` on the handler. Done in f0285a5b44baae7d0328fab5928673e2939fd94b.
- [ ] Refactor all the hacky code -> create contract for the calls and move stuff to the handler. The behaviour is potentially not necessary.


packages/eightballer/skills/contract_subscription/behaviours.py
```diff
@@ -20,41 +20,98 @@
...
+                )
+            self.context.logger.info(
+                "Getting parts that were missed while disconnected."
+            )
+            self.context.shared_state[DISCONNECTION_POINT] = None
+            return
+
+        if not is_connected and disconnection_point is None:
+            self._subscription_required = True
+            # TODO think how to do this. We cannot get the block number while rpc connection is dropped,
```
0xArdi, line 92: Not clear to me why? Cant you get the current block and extrapolate that by some number?

0xArdi (approved):

packages/eightballer/skills/contract_subscription/behaviours.py
Adamantios: As mentioned in the description, I plan to do this by keeping track of the block number that is returned from `eth_getTransactionReceipt` on the handler. I plan to set the shared state's `disconnection_point` with `tx_receipt["blockNumber"]`.

Adamantios: This was implemented on f0285a5b44baae7d0328fab5928673e2939fd94b.

PR #48, opened 2023-05-25
chore: initial steps

DavidMinarsch:


packages/valory/connections/websocket_client/connection.py
```diff
@@ -86,6 +86,8 @@ async def connect(self) -> None:
                 if self._thread is None:
                     self._thread = Thread(target=asyncio.run, args=(self._run(),))
                     self._thread.start()
+                else:
```
DavidMinarsch, line 89: Check Yuri

packages/valory/skills/contract_subscription/behaviours.py
```diff
@@ -59,7 +59,24 @@ def act(self) -> None:
         is_connected = cast(WebSocketClient, self._ws_client_connection).is_connected
         disconnection_point = self.context.shared_state.get(DISCONNECTION_POINT, None)
```
DavidMinarsch, line 61: I rearranged this and added logging to simplify understanding of the control flow

packages/valory/connections/websocket_client/connection.py
solarw: @DavidMinarsch  what about to use more async friendly lib for websockets to avoid threads usage?

is current websocket lib mandatory?

Adamantios: @solarw I tried to use `websockets` but we cannot have it because of dependency issues.

While trying to use it I got:
```
Failed to establish WebSocket connection: As of 3.10, the *loop* parameter was removed from Lock() since it is no longer necessary
```

therefore I had to update the `websockets` version. However, I could not because of dependencies:
```
Because open-aea-ledger-ethereum (1.33.0) depends on web3 (5.25.0)
 and web3 (5.25.0) depends on websockets (>=9.1,<10), open-aea-ledger-ethereum (1.33.0) requires websockets (>=9.1,<10).
So, because mech depends on both websockets (^11.0.3) and open-aea-ledger-ethereum (1.33.0), version solving failed.
```

DavidMinarsch:
Closed in favour of #52

PR #49, opened 2023-05-25
Evm tools

N0xMare:


tools/native_transfer_request.py
```diff
@@ -0,0 +1,124 @@
...
+
+"to": recipient_address, 
+"value": value, 
+"wei_value": wei_value
+
+Do not respond with anything else other than the transaction object you constructed with the correct known variables the agent had before the request and the correct unknown values found in the user request prompt as input to the web3.py signing method.
+"""
+
+
+def native_transfer(**kwargs) -> str:
```
DavidMinarsch, line 38: For this to work with `mech` it needs to be called `run`

tools/native_transfer_request.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Execute a native token transfer on Ethereum. This involves:
+    1. using the user prompt to create the formatted tool prompt for the LLM (Done)
+    2. using the tool prompt to create the transaction object in string form (Done)
+    3. parsing the transaction object string to get the transaction object (WIP)
+    4. signing the transaction object with the private key of the agent (WIP)
+    5. sending the signed transaction object + other data if desired to the agent process (WIP)
```
DavidMinarsch, line 45: This happens in the mech

tools/native_transfer_request.py
```diff
@@ -0,0 +1,124 @@
...
+        "gasPrice": gas_price,
+        "nonce": current_nonce,
+    }
+
+    # sign the transaction
+    signed_txn = w3.eth.account.sign_transaction(transaction, private_key=AGENT_PRIVATE_KEY)
+
+    # send the transaction
+    tx_hash = w3.eth.send_raw_transaction(signed_txn.rawTransaction)
+
```
DavidMinarsch, line 95: Not happening here

tools/native_transfer_request.py
```diff
@@ -0,0 +1,124 @@
...
+    return response, tx_hash
```
DavidMinarsch, line 96: This should return something that can be passed to multisend

tools/openai_request.py
```diff
@@ -28,7 +28,10 @@
     "chat": ["gpt-3.5-turbo", "gpt-4"],
     "completion": ["text-davinci-002", "text-davinci-003"]
 }
-ALLOWED_TOOLS = [PREFIX + value for value in values for values in ENGINES.values()]
+#ALLOWED_TOOLS = [PREFIX + value for value in values for values in ENGINES.values()]
```
DavidMinarsch, line 31: Please revert all changes here. This has been fixed on main

tools/native_transfer_request.py
N0xMare: how do you avoid the namespace collision that creates an infinite loop because of the openai_request tool also having the name run?

tools/openai_request.py
N0xMare: Im curious, this works for with all the engines?

ALLOWED_TOOLS = [PREFIX + value for value in values for values in ENGINES.values()]

str.strip() in the case of gtp-3.5-turbo will not work as strip() removes prefix and postfix "o" from turbo

hence why I did tool.removeprefix(PREFIX) instead

tools/native_transfer_request.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # parse the response to get the transaction object string itself
+    parsed_txs = ast.literal_eval(response)
+    
+    print("\033[95m\033[1m" + "\n PARSED TXS OBJECT:" + "\033[0m\033[0m")
+    print(parsed_txs)
+
+    # build the transaction object, unknowns are referenced from parsed_txs
+    transaction = {
+        "from": agent_address,
```
0xArdi, line 82: We will only need the following fields:
- to
- value

0xArdi: You can return a dictionary

0xArdi: In case of multiple transactions, return a list of dictionaries

N0xMare:
not needed, merging into feat/multisend

PR #50, opened 2023-05-25
feat: add polling

0xArdi:


packages/valory/skills/transaction_preparation_abci/models.py
```diff
@@ -43,8 +43,9 @@ class Params(BaseParams):
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         """Initialize the parameters object."""
-
-        self.agent_mech_contract_address = self._ensure("agent_mech_contract_address", kwargs, str)
+        self.agent_mech_contract_address = kwargs.get("agent_mech_contract_address", None)
```
DavidMinarsch, line 46: Why the change?

packages/valory/skills/task_execution_abci/behaviours.py
```diff
@@ -183,6 +180,16 @@ def async_act(self) -> Generator:  # pylint: disable=R0914,R0915
...
+    def get_ipfs_file_hash(self, data: bytes) -> str:
+        """Get hash from bytes"""
+        try:
+            return str(CID.from_string(data.decode()))
+        except Exception:
+            # if something goes wrong, fallback to sha256
+            file_hash = data.hex()
+            file_hash = CID_PREFIX + file_hash
+            file_hash = str(CID.from_string(file_hash))
+            return file_hash
```
DavidMinarsch, line 192: why this change?

packages/valory/skills/multiplexer_abci/models.py
```diff
@@ -52,7 +53,11 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         """Initialize the parameters object."""
 
         self.reset_period_count = self._ensure("reset_period_count", kwargs, int)
-
+        self.use_polling = self._ensure("use_polling", kwargs, bool)
+        self.polling_interval = self._ensure("polling_interval", kwargs, int)
+        self.agent_mech_contract_address = kwargs.get("agent_mech_contract_address", None)
+        if self.agent_mech_contract_address is None:
```
DavidMinarsch, line 59: Same question as below

0xArdi: both params require this value, if use ensure one of them can have it only since ensure pops it

packages/valory/skills/task_execution_abci/behaviours.py
0xArdi: response format was diff, the rpc we will have to use was returning in a different format. I didnt want to waste time to figure out why

packages/valory/skills/transaction_preparation_abci/models.py
0xArdi: https://github.com/valory-xyz/mech/pull/50/files#r1205344050

Adamantios (approved):

PR #51, opened 2023-05-25
websocket client connection on mtop of aiohttp

solarw:


DavidMinarsch:
Closed in favour of #52

PR #52, opened 2023-05-25
Fix websocket connection implementation

angrybayblade:


packages/valory/connections/websocket_client/connection.py
```diff
@@ -140,18 +159,23 @@ async def receive(self, *args: Any, **kwargs: Any):  # noqa: V107
...
+            try:
+                msg = await self.loop.run_in_executor(
+                    executor=self._executor, func=self._wss.recv
+                )
+                return self._from_wss_msg_to_envelope(msg)
+            except websocket.WebSocketConnectionClosedException:
+                self.logger.error("Websocket connection closed.")
+                self.state = ConnectionStates.disconnecting
+                await self.disconnect()
+                await self._reconnect()
```
angrybayblade, line 178: This is something that should be handled by the skill, we can accomplish this once we introduce the protocol

Adamantios (approved):

0xArdi (approved):

PR #60, opened 2023-05-26
Introduce websocket protocol

angrybayblade:


packages/valory/protocols/__init__.py
Adamantios, line 1: The header is missing.

packages/valory/protocols/websocket_client/README.md
Adamantios, line 1: The `recv` part confused me. Shouldn't there be a way to check for data only with the connection id?

packages/valory/connections/websocket_client/connection.py
```diff
@@ -21,138 +21,463 @@
...
+                Envelope(
+                    to=self._to,  # self._envelope.sender,
+                    sender=self._sender,  # self._envelope.to,
+                    message=message,
+                )
+            )
+
+    async def subscribe(self, url: str) -> "WebsocketSubcription":
+        """Subscribe to websocket."""
+        # TODO: make these configurable
```
0xArdi, line 167: you are addressing this in a future version I assume?

0xArdi (approved):

Adamantios (approved):

PR #75, opened 2023-06-12
Fix tool to return proper data types

angrybayblade:


tools/stabilityai_request.py
```diff
@@ -110,11 +111,10 @@ def run(**kwargs: Any) -> Union[Dict, List]:
         raise ValueError(f"Non-200 response ({response.status_code}): {response.text}")
 
     data = response.json()
-
     finish_reason = data.get("finishReason", FinishReason.SUCCESS.name)
     if finish_reason == FinishReason.SUCCESS.name:
         # return image data
-        return data
+        return data, None
```
0xArdi, line 117: Afaik the multisend logic expects an empty list, not None when there are no native txs. Also, does this mean that now you want a single tx, instead of a list of them?

angrybayblade: Yes, if we fail to create a tx object we just send one transaction

0xArdi: Okay, and you dont want multiple txs right? The tool can output a single tx as-is.

angrybayblade: yes

0xArdi (approved):

PR #76, opened 2023-06-12
Add support for requesting data using ACN

angrybayblade:
- Implements ACN data requests protocol
- Adds support for processing data requests made using the ACN data request protocol


.github/workflows/common_checks.yaml
```diff
@@ -109,8 +109,9 @@ jobs:
         # Install `mdspell` for spelling checks
         sudo npm install -g markdown-spellcheck
 
-    - name: Check generate protocols
-      run: tox -e check-generate-all-protocols
+    # Temporary skip
+    # - name: Check generate protocols
+    #   run: tox -e check-generate-all-protocols
```
angrybayblade, line 114: I'll open a separate PR on top

DavidMinarsch: Before we merge - please fix this

DavidMinarsch: (I think you need to add a setup.cfg with the isort etc configs). We can remove this again if we update the tooling a bit in the future.

packages/valory/skills/task_execution_abci/behaviours.py
```diff
@@ -224,9 +231,17 @@ def get_task_result(  # pylint: disable=R0914,R1710
 
         # Handle invalid requests
         if self._invalid_request:
-            deliver_msg = "no_op"
             # respond with no_op and no multisend transactions
+            deliver_msg = "no_op"
             request_id = cast(str, self.request_id)
+            self.data_requests.set_data(
+                request_id=str(self.request_id), data=deliver_msg
+            )
```
DavidMinarsch, line 239: why only here and not below?

packages/valory/skills/task_execution_abci/behaviours.py
```diff
@@ -279,6 +299,27 @@ def get_task_result(  # pylint: disable=R0914,R1710
...
         return request_id, hex_multihash, transaction
 
+    def send_data_to_acn(
+        self,
+        sender_address: str,
+        request_id: str,
+        data: Any,
+    ) -> None:
+        """Handle callbacks."""
+        self.context.logger.info(f"Sending data to ACN for request ID {request_id}")
```
DavidMinarsch, line 309:
```suggestion
        self.context.logger.info(f"Sending data to {recipient} via ACN for request ID {request_id}")
```

packages/valory/skills/task_execution_abci/behaviours.py
angrybayblade: This is not required anymore since we have removed the pub-sub model

setup.cfg
```diff
@@ -0,0 +1,59 @@
+[flake8]
```
DavidMinarsch, line 1: Please include only what's needed for the `all-protocols generate` command. ALso add a comment so we can remove it later

packages/valory/skills/task_execution_abci/behaviours.py
```diff
@@ -279,6 +296,29 @@ def get_task_result(  # pylint: disable=R0914,R1710
         request_id = cast(str, self.request_id)
         return request_id, hex_multihash, transaction
 
+    def send_data_to_acn(
```
DavidMinarsch, line 299:
```suggestion
    def send_data_via_acn(
```

PR #84, opened 2023-07-05
chore: prep for minting components

DavidMinarsch:
Missing:
- [x] dependencies
- [x] prep metadata
- [x] mint components


mints/mapping.txt
```diff
@@ -0,0 +1,15 @@
+token_id-"component_id"
+X-"protocol/valory/default/1.0.0"
+X-"protocol/valory/acn_data_share/0.1.0"
+X-"connection/valory/websocket_client/0.1.0" -> "protocol/valory/default/1.0.0"
+WARNING: might we want to call the associated protocol websocket? For forward compatibility to when we change it
```
DavidMinarsch, line 5: @0xArdi I know this would create extra work. Thoughts?

mints/mapping.txt
```diff
@@ -0,0 +1,15 @@
...
+X-"contract/valory/agent_mech/0.1.0"
+X-"skill/valory/contract_subscription/0.1.0" -> "protocol/valory/default/1.0.0", "connection/valory/websocket_client/0.1.0"
+X-"skill/valory/multiplexer_abci/0.1.0" -> 3, 14, "contract/valory/agent_mech/0.1.0"
+X-"skill/valory/task_execution_abci/0.1.0" -> 3, 11, 19, 20, "contract/valory/agent_mech/0.1.0", "protocol/valory/acn_data_share/0.1.0", 14, 22
+X-
+What happened to transaction preparation?
```
DavidMinarsch, line 11: @Adamantios & @0xArdi

Adamantios: Not sure, it was removed on b6d31612c003e2c5a23cfa1a76f0e6f85246041e.

0xArdi: It only complicated things, especially when we added the multisend.

mints/mapping.txt
0xArdi: In reality it is the default protocol, is there any benefit to renaming it?

DavidMinarsch: It's fine. If we knew we'd immediately replace with websocket protocol then yes there'd be a benefit

PR #86, opened 2023-07-12
Fix additional information collection on prediction tool

angrybayblade:


Adamantios (approved):
LGTM, did you run the tool?

Also, do not forget to push it to IPFS and add the hash in the example env file.

packages/valory/skills/task_execution_abci/rounds.py
```diff
@@ -144,7 +145,7 @@ class TaskExecutionAbciApp(AbciApp[Event]):
         FinishedTaskExecutionWithErrorRound,
     }
     event_to_timeout: EventToTimeout = {
-        Event.ROUND_TIMEOUT: 30.0,
+        Event.TASK_EXECUTION_ROUND_TIMEOUT: 60.0,
```
Adamantios, line 148: Ideally, we should have an override for this in the config.

Adamantios (approved):

PR #88, opened 2023-07-12
fix: timeout mech

0xArdi:
Set `round_timeout_seconds` for `TaskExecutionEvent.TASK_EXECUTION_ROUND_TIMEOUT`.

### DIFF
--- a/packages/valory/skills/mech_abci/models.py
+++ b/packages/valory/skills/mech_abci/models.py
@@ -81,6 +81,10 @@ class SharedState(TaskExecSharedState):
         ] = self.context.params.round_timeout_seconds
 
         MechAbciApp.event_to_timeout[
+            TaskExecutionEvent.TASK_EXECUTION_ROUND_TIMEOUT
+        ] = self.context.params.round_timeout_seconds
+
+        MechAbciApp.event_to_timeout[
             ResetPauseEvent.ROUND_TIMEOUT
         ] = self.context.params.round_timeout_seconds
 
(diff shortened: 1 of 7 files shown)


angrybayblade (approved):

packages/valory/skills/mech_abci/models.py
```diff
@@ -80,6 +80,10 @@ def setup(self) -> None:
             TaskExecutionEvent.ROUND_TIMEOUT
         ] = self.context.params.round_timeout_seconds
 
+        MechAbciApp.event_to_timeout[
+            TaskExecutionEvent.TASK_EXECUTION_ROUND_TIMEOUT
+        ] = self.context.params.round_timeout_seconds
```
Adamantios, line 85: Then what is the meaning of `TASK_EXECUTION_ROUND_TIMEOUT` event? 
Shouldn't we have a different config for this timeout?

0xArdi: In reality it's just a normal round timeout. I don't see any reason why we should have a diff param atm.

Adamantios: This was introduced by @angrybayblade on #86. The only reason to do that is to have a different timeout for this round which takes longer. However, it was never assigned a value, but you did this here. So, this should not be assigned the round timeout's value, otherwise, the change was not necessary to begin with on #86.

Adamantios (approved):

PR #89, opened 2023-07-17
Fix/tool

dvilelaf:
* Fixes tool dependencies


packages/valory/agents/mech/aea-config.yaml
```diff
@@ -72,6 +72,8 @@ logging_config:
       propagate: true
 dependencies:
   open-aea-ledger-ethereum: {}
+  googlesearch-python:
+    version: ==1.2.3
```
0xArdi, line 76: I think you had beautifulSoup as well here

dvilelaf: Right!

0xArdi: i see its already in the skill

0xArdi: i think so

dvilelaf: Sorted

0xArdi (approved):

Adamantios (approved):

PR #97, opened 2023-08-02
Add request timeout on `openai` api calls

angrybayblade:


jmoreira-valory (approved):

tools/prediction_request.py
```diff
@@ -205,7 +205,8 @@ def fetch_additional_information(
         temperature=temperature,
         max_tokens=max_tokens,
         n=1,
-        timeout=120,
+        timeout=30,
+        request_timeout=90,
```
Adamantios, line 209: Why larger `request_timeout` than `timeout`?

angrybayblade: `request_timeout` refers to the overall timeout of the request where as the `timeout` takes effect when retrying. I've made them similar in [1499a2e](https://github.com/valory-xyz/mech/pull/97/commits/1499a2ecea652a0e17539ec4b3d5a624f7528666)

Adamantios: Yeap, it only makes sense if it is `>=request_timeout`.

Adamantios (approved):

PR #104, opened 2023-08-26
chore: fix website content retrieval

DavidMinarsch:


tools/prediction_request.py
```diff
@@ -157,29 +157,39 @@ def extract_text(
...
-    urls: List[str], window: int = 5
+    urls: List[str], window: int = 5, timeout: int = 10
 ) -> Generator[None, None, List[requests.Response]]:
     """Iter URLs in batches."""
     with ThreadPoolExecutor() as executor:
         for i in range(0, len(urls), window):
             batch = urls[i : i + window]
-            yield executor.map(requests.get, batch)
-
+            futures = [executor.submit(requests.get, url, timeout=timeout) for url in batch]
```
DavidMinarsch, line 166: We need a timeout. Otherwise this hangs indefinitely (exactly what we observe)!

tools/prediction_request.py
```diff
@@ -157,29 +157,39 @@ def extract_text(
...
+            yield futures
 
 def extract_texts(urls: List[str], num_words: int = 300) -> List[str]:
     """Extract texts from URLs"""
     max_allowed = 5
     extracted_texts = []
     count = 0
+    stop = False
```
DavidMinarsch, line 174: The logic here was wrong. The break only exists the inner loop, not the outer one. The return statement on the inner loop makes no sense

0xArdi (approved):

PR #117, opened 2023-10-04
fix: add deps for `optimization_by_prompting` tool

0xArdi:
This PR updates the deps to accommodate the  `optimization_by_prompting` tool.


packages/valory/connections/http_client/connection.yaml
```diff
@@ -0,0 +1,31 @@
...
+class_name: HTTPClientConnection
+config:
+  host: 127.0.0.1
+  port: 8000
+excluded_protocols: []
+restricted_to_protocols:
+- valory/http:1.0.0
+dependencies:
+  aiohttp:
+    version: <3.9,>=3.7.4
```
angrybayblade, line 28: Doesn't this need to be `>=3.8.3,<4.0.0`

0xArdi: This is the required change

angrybayblade (approved):

PR #118, opened 2023-10-10
chore: add bittensor prediction request tool - using bitapai.io

DavidMinarsch:
This adapts the prediction_request tool from openai to bittensor, using https://bitapai.io/docs/introduction/. 

To test:

```
from tools.prediction_request_bt import run
kwargs = {"tool": "prediction-offline", "prompt": "What is the weather in Phoenix, Arizona?", "api_keys": {"bitapai": "<YOUR_KEY_HERE>"}}
run(**kwargs)
```

### DIFF
--- a/tools/prediction_request_bt.py
+++ b/tools/prediction_request_bt.py
@@ -268,8 +268,6 @@ def run(**kwargs) -> Tuple[str, Optional[Dict[str, Any]]]:
           "content": prediction_prompt
         }
       ],
-      "pool_id": 4,
-      "count": 3,
       "return_all": True
     })
     headers = {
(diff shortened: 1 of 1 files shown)


tools/prediction_request_bt.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+NUM_URLS_EXTRACT = 5
+DEFAULT_OPENAI_SETTINGS = {
+    "max_tokens": 500,
+    "temperature": 0.7,
+}
+ALLOWED_TOOLS = [
+    "prediction-offline",
+    "prediction-online",
```
DavidMinarsch, line 40:
```suggestion
    "prediction-offline-bt",
    "prediction-online-bt",
```

DavidMinarsch:
Also looked into this, but again seems to be not working https://python.langchain.com/docs/integrations/llms/bittensor

tools/prediction_request_bt.py
```diff
@@ -0,0 +1,282 @@
...
+        {
+          "role": "system",
+          "content": "You are an AI assistant"
+        },
+        {
+          "role": "user",
+          "content": url_query_prompt
+        }
+      ],
+      "pool_id": 4,
```
salahawk, line 216: Better to remove `pool_id` here

tools/prediction_request_bt.py
```diff
@@ -0,0 +1,282 @@
...
+      "return_all": True
+    })
+    headers = {
+      'Content-Type': 'application/json',
+      'X-API-KEY': api_key
+    }
+    conn.request("POST", "/text", payload, headers)
+    res = conn.getresponse()
+    data = res.read()
+    json_data = data
```
salahawk, line 227: As you have `count` set being 3, the response has 3 choices.
You can either set `count` to be 1, or choose one of 3 choices in response.

mkutsovsky: @DavidMinarsch it works when you remove count and pool_id params (api will default to querying 20 miners, a good amount which gave reasonable responses when I just ran it). will require wrangling the outputs from the a bit after that point

DavidMinarsch: Thanks! Got it to work now.

PR #134, opened 2023-11-14
feat: add max timeout limit

### LINKED ISSUE #131: Handle task timeouts
Currently whenever a task times out during execution, we handle it by adding it to the end of the queue.
Although we haven't seen it yet, it might be possible that a task gets stuck forever in this loop. 
What would be better if we introduced a maximum retry per task instead of indefinitely re-adding to the queue.

0xArdi:
Fixes #131


packages/valory/skills/task_execution/behaviours.py
```diff
@@ -279,12 +292,30 @@ def _handle_timeout_task(self) -> None:
...
+            return None
+
+        self.context.logger.info(
+            f"Task {req_id} has reached the timeout limit of{self.params.timeout_limit}. "
+            f"It won't be added to the end of the queue again."
+        )
+        task_result = (
+            f"Task timed out {self.params.timeout_limit} times during execution. ",
+            "",
+            None,
```
Adamantios, line 316: Does this mean that the prompt will not be in the response in this case?

Adamantios (approved):

packages/valory/skills/task_execution/behaviours.py
0xArdi: Yes, we cannot know the prompt, as the task never returns in such instances.

angrybayblade (approved):

PR #136, opened 2023-11-24
Feat/add funds recovery

0xArdi:
This PR adds support for recovering funds from the mech.


packages/valory/agents/mech/aea-config.yaml
Adamantios, line 1: Do we need to override `profit_split_freq` here and in the service yaml?

0xArdi: yes

Adamantios: Ok, let's add this.

Adamantios (approved):

angrybayblade (approved):

PR #138, opened 2023-11-28
Fix/flaky tools

0xArdi:
This PR fixes some flaky behaviour in the `deepmind-optimization` and `prediction_request_claude` tools.


tools/optimization_by_prompting.py
```diff
@@ -187,10 +190,17 @@ def prompt_engineer(openai_api_key, init_instructions, instructions_format, iter
...
+        try:
+            prompt = PromptTemplate(
+                input_variables=["user_prompt", "additional_information"],
+                template=generated_template,
+            )
+        except Exception as e:
+            # it may happen that the generated prompt is not valid
+            # in that case, we just skip it
+            print(f"Failed to parse template {generated_template}: {e}")
+            continue
```
angrybayblade, line 203: Shouldn't we change something before retrying? because it looks like the value of generated_template stays same

0xArdi: I see your point, i guess it makes sense that in such instances we regenerate the template.

Adamantios (approved):

angrybayblade (approved):

Adamantios (approved):

PR #141, opened 2023-12-12
Feat/make tools as packages

0xArdi:
This PR adds support for having tools as custom packages.


angrybayblade (approved):

packages/valory/skills/task_execution/utils/task.py
```diff
@@ -28,8 +28,9 @@ class AnyToolAsTask:
     def execute(self, *args: Any, **kwargs: Any) -> Any:
         """Execute the task."""
         tool_py = kwargs.pop("tool_py")
-        if "run" in globals():
-            del globals()["run"]
+        callable_method = kwargs.pop("callable_method")
```
DavidMinarsch, line 31: Is this going to work with multiple packages having the same callable name?

DavidMinarsch (commented):
@0xArdi one comment - ow looks good 

please check for each package who the author was. In some cases it’s not Valory

packages/valory/skills/task_execution/utils/task.py
0xArdi: yes

0xArdi: all tools had `run` as their callable and it worked

jmoreira-valory (approved):

PR #148, opened 2024-01-02
feat: add a max block window size

0xArdi:
This PR introduces a max block window size when checking for mech reqs.

### DIFF
--- a/packages/valory/contracts/agent_mech/contract.py
+++ b/packages/valory/contracts/agent_mech/contract.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # ------------------------------------------------------------------------------
 #
-#   Copyright 2023 Valory AG
+#   Copyright 2023-2024 Valory AG
 #
 #   Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
--- a/packages/valory/contracts/agent_mech/contract.py
+++ b/packages/valory/contracts/agent_mech/contract.py
@@ -82,6 +82,9 @@ partial_abis = [
 ]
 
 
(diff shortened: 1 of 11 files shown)


jmoreira-valory (approved):

packages/valory/contracts/agent_mech/contract.py
```diff
@@ -82,6 +82,9 @@
 ]
 
 
+BLOCK_INDEXING_TOLERANCE = 50
+
+
```
Adamantios, line 87: Where is this used?

0xArdi: Nowhere, i took a different approach first. Ill remove.

packages/valory/contracts/agent_mech/contract.py
```diff
@@ -270,9 +273,13 @@ def get_multiple_undelivered_reqs(
         contract_address: str,
         contract_addresses: List[str],
         from_block: BlockIdentifier = "earliest",
+        max_block_window: int = 1000,
         **kwargs: Any,
     ) -> JSONLike:
         """Get the requests that are not delivered."""
+        current_block = ledger_api.api.eth.block_number
+        if from_block != "earliest" and current_block - from_block > max_block_window:
+            from_block = current_block - max_block_window
```
Adamantios, line 282: Shouldn't we change the `to_block` instead and split the search into batches instead of skipping blocks?

0xArdi: The way this works with the rest of the behaviour, it is guaranteed that we do not lose any blocks.

Adamantios (approved):

PR #151, opened 2024-01-15
feat: maintain agent balances

0xArdi:
This PR adds support for making sure agents have enough balance to operate.
Whenever the balance gets lower than allowed we fund them from the mech profits.


packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -527,6 +553,38 @@ def _accumulate_reqs_by_operator(
...
 
+    def _get_agent_balances(self) -> Generator[None, None, Optional[Dict[str, int]]]:
+        """Get the agent balances."""
+        balances = {}
+        for agent in self.synchronized_data.all_participants:
+            balance = yield from self._get_balance(agent)
+            if balance is None:
+                self.context.logger.warning(
+                    f"Could not get balance for agent {agent}. Skipping re-funding."
+                )
```
Adamantios, line 564: Maybe worth retrying?

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -527,6 +553,38 @@ def _accumulate_reqs_by_operator(
...
+    def _get_agent_funding_amounts(
+        self,
+    ) -> Generator[None, None, Optional[Dict[str, int]]]:
+        """Get the agent balances."""
+        agent_funding_amounts = {}
+        agent_balances = yield from self._get_agent_balances()
+        if agent_balances is None:
+            self.context.logger.warning(
+                "Could not get agent balances. Skipping re-funding."
+            )
```
Adamantios, line 579: Same as above, retrying here would be nice.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -527,6 +553,38 @@ def _accumulate_reqs_by_operator(
...
+                    f"Could not get balance for agent {agent}. Skipping re-funding."
+                )
+                return None
+            balances[agent] = balance
+
+        return balances
+
+    def _get_agent_funding_amounts(
+        self,
+    ) -> Generator[None, None, Optional[Dict[str, int]]]:
```
Adamantios, line 572: Could be returning a `set`/`tuple`/`list`. The funding amount is always the same.

Adamantios (approved):

jmoreira-valory (approved):

packages/valory/skills/task_submission_abci/behaviours.py
0xArdi: the way the logic is setup, it will be retired on the next fund splitting. The threshold should be good enough for the service to continue working until then. 
This is part of a multisend that goes along with the deliver. It's desirable that it is as quick as possible.

packages/valory/skills/task_submission_abci/behaviours.py
0xArdi: makes the logic where this is used simpler

PR #153, opened 2024-01-17
EvoResearcher Tool: prediction_with_research_report

namesty:
## Overview

This PR aims to add a new tool that uses: https://github.com/polywrap/evo.researcher in order to generate a research report which is then injected as `additional_information` to the already existing prediction prompt from the `prediction_sentence_embedding` tool.

This has shown to consistently impact the scores of predictions for questions such as: "Will Twitter implement a new misinformation policy before the 2024 elections?", "Will the goal of not surpassing 1.5 degrees Celsius set by the Paris Agreement be reached by 2030?", among others taken from [Omen's subgraph](https://api.thegraph.com/subgraphs/name/protofire/omen-xdai/graphql?query=query+MyQuery+%7B%0A++questions+%7B%0A++++title%0A++%7D%0A%7D). It makes prediction scores more nuanced and `information_utility` scores higher.

[Evo researcher's repo](https://github.com/polywrap/evo.researcher?tab=readme-ov-file#research-function) contains the implementation of the research function.

### Average times in seconds (s):

Using `gpt-3.5-turbo-1106`: **58**
Using `gpt-4-1106-preview`: **120**

### Average costs in USD ($):

Using `gpt-3.5-turbo-1106`: **0,0099**
Using `gpt-4-1106-preview`: **0,1012**

## Notes

- Sometimes behavior is not consistent when using `gpt-3.5-turbo-1106`
- Prediction prompt and generation are untouched - taken directly from `prediction_sentence_embedding`


tools/prediction_with_research_report/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains the bet amount per threshold strategy."""
```
Adamantios, line 20: Please update the docstring.

tools/prediction_with_research_report/prediction_with_research_report.py
```diff
@@ -0,0 +1,167 @@
...
+}
+
+ALLOWED_TOOLS = [
+    "prediction-sentence-embedding-conservative",
+    "prediction-sentence-embedding-bold",
+]
+TOOL_TO_ENGINE = {
+    "prediction-sentence-embedding-conservative": "gpt-3.5-turbo-1106",
+    "prediction-sentence-embedding-bold": "gpt-4-1106-preview",
+}
```
Adamantios, line 39: Let's prefix with `prediction-with-research` instead.

Adamantios (approved):

tools/prediction_with_research_report/component.yaml
```diff
@@ -0,0 +1,14 @@
+name: prediction_with_research_report
+author: evo.ninja
+version: 0.1.0
+type: custom
+description: A tool for making binary predictions on markets using more robust research from external sources, and OpenAI API.
+license: Apache-2.0
+aea_version: '>=1.0.0, <2.0.0'
+fingerprint:
+  __init__.py: bafybeid74x2il46x64tdpixbuw4l6iua2p2rxseewtxfnrjdp4dcer5yee
+  prediction_with_research_report.py: bafybeieqe6koonllyrx3oiocxm7gukzaoyxhkrrvpxwob4zsut74onkpcy
```
Adamantios, line 10: Looks like [bafybeieqe6koonllyrx3oiocxm7gukzaoyxhkrrvpxwob4zsut74onkpcy](https://gateway.autonolas.tech/ipfs/bafybeieqe6koonllyrx3oiocxm7gukzaoyxhkrrvpxwob4zsut74onkpcy) is not on IPFS.

pyproject.toml
```diff
@@ -61,4 +61,4 @@ pandas = "==2.1.1"
 tiktoken = "==0.5.1"
 python-dateutil = "==2.8.2"
 tomte = {version = "==0.2.15", extras = ["cli", "tests"]}
-
+evo_researcher = "==0.1.8"
```
0xArdi, line 64: The mech specifies `^3.10` for the allowed python versions. But this package (or one of its deps) requires sqlite3>3.35.0, which doesnt come with all ^3.10 versions.

tools/prediction_with_research_report/prediction_with_research_report.py
```diff
@@ -0,0 +1,167 @@
...
+        raise ValueError(f"TOOL {tool} is not supported.")
+    
+    engine = TOOL_TO_ENGINE[tool]
+    
+    (research_report, _) = research(
+        prompt,
+        openai_key=openai_api_key,
+        tavily_key=tavily_api_key,
+        model=engine,
+        kwargs={
```
kongzii, line 130: `research` function in [evo.research's code ](https://github.com/polywrap/evo.researcher/blob/main/evo_researcher/functions/research.py#L11) is defined as

```
def research(goal: str, openai_key: str, tavily_key: str, model: str = "gpt-4-1106-preview",  **kwargs) -> tuple[str, str]:
    initial_subqueries_limit = kwargs.get('initial_subqueries_limit', 20)
    ...
```

Shouldn't this be written as 

```
research(
  ...,
  initial_subqueries_limit= initial_subqueries_limit,
  subqueries_limit=subqueries_limit,
  ...
)
```

? Python will assign it to dict automatically, if there is `**`. 

Right now, the code should work, but it will always use default arguments, as for example `initial_subqueries_limit` won't be in the kwargs, it will be as `kwargs["kwargs"]["initial_subqueries_limit"]`.

namesty: Done, updated the function signature on `evo_researcher`

tools/prediction_with_research_report/prediction_with_research_report.py
```diff
@@ -0,0 +1,167 @@
...
+import openai
+
+DEFAULT_OPENAI_SETTINGS = {
+    "temperature": 0,
+    "max_compl_tokens": 500,
+}
+
+ALLOWED_TOOLS = [
+    "prediction-sentence-embedding-conservative",
+    "prediction-sentence-embedding-bold",
```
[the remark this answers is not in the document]
namesty: Yes, updated

tools/prediction_with_research_report/prediction_with_research_report.py
namesty: Done :)

tools/prediction_with_research_report/__init__.py
namesty: Should it be any different? I see it's the same in all tools

tools/prediction_with_research_report/component.yaml
namesty: Just updated the hashes and pinned the files. They should be:

`Dir`: https://ipfs.io/ipfs/bafybeifrvq6sgzu4iykr7jwr53lup7gpmk235d7cc4xh7zy4kc6ypww7ia/
`__init__.py`: https://ipfs.io/ipfs/bafybeid74x2il46x64tdpixbuw4l6iua2p2rxseewtxfnrjdp4dcer5yee
`prediction_with_research_report.py`: https://ipfs.io/ipfs/bafybeidypmypdaedqmpvqgtmfvi53usozq4nytffvchlwhepb2vyk6kyae
`component.yaml`: https://ipfs.io/ipfs/bafybeigekhcjjvzivjmakckvwf3ryqcxnoa6sotcov7tuicjrtoq4fjlwe

pyproject.toml
namesty: You're right. This is due to ChromaDB: https://docs.trychroma.com/troubleshooting#sqlite

I went with their second proposed approach (the first being just upgrading to a later Python 3.10 version): https://gist.github.com/defulmere/8b9695e415a44271061cc8e272f3c300

But do let me know if this should be handled differently

tools/prediction_with_research_report/__init__.py
Adamantios: I see :disappointed: 
Opened #157.

Ideally, it should include a description of the corresponding module. Something like:

```suggestion
"""This module contains the evo researcher tool."""
```

0xArdi:
@namesty we are aiming to include this next week on the mech. There are two things required to do that:
- Deps introduced here should be satisfied in the mech agent.
- Dynamic pricing of requests is introduced in the mech.

0xArdi:
Dynamic pricing was added on #166

0xArdi (approved):

namesty:
@0xArdi I've updated the deps, added token counting for dynamic pricing as per the other example tools, and re-pinned everything to IPFS

dOrgJelli:
I believe in this PR there should have been updates to the mech's `api_keys`, as it adds a new dependency on `tavily`:
https://github.com/valory-xyz/mech/blob/5cb03836c00b90580ab61c782cafa64a0f7925ca/packages/polywrap/customs/prediction_with_research_report/prediction_with_research_report.py#L519

Not sure what needs to be updated though, would appreciate some guidance @0xArdi @Adamantios @kongzii
https://github.com/search?q=repo%3Avalory-xyz%2Fmech%20api_keys&type=code

PR #159, opened 2024-02-06
Feat/bump openai

0xArdi:
This PR bumps the openai version to `v1.11.0`, due to [#820](https://github.com/openai/openai-python/issues/820).


tools/native_transfer_request/native_transfer_request.py
```diff
@@ -24,7 +24,26 @@
...
+    if client is None:
+        client = OpenAI(api_key=api_key)
+    return client
+
+def close_openai_client() -> None:
+    """Close the OpenAI client"""
+    global client
+    if client is not None:
+        client.close()
+        client = None
```
Adamantios, line 45: How about a context manager instead?

0xArdi: I have another PR overlapping this, I can address there.

0xArdi: @Adamantios merging this as-is. Will be addressed in an upcoming PR.

PR #155, opened 2024-01-25
proposal changes for benchmark runs

moarshy:
- a proposal to run benchmark

PR #160, opened 2024-02-06
Feat/mech support

0xArdi:
This PR adds support in the mech for the changes proposed in #155 .


packages/valory/skills/task_execution/utils/benchmarks.py
```diff
@@ -0,0 +1,105 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+"""Benchmarking for tools."""
+
+import logging
+from typing import Any, Dict, Union
+
+import anthropic
```
DavidMinarsch, line 24: Why? That seems strange to include here

packages/valory/skills/task_execution/utils/benchmarks.py
```diff
@@ -0,0 +1,105 @@
...
+
+def encoding_for_model(model: str) -> Encoding:
+    """Get the encoding for a model."""
+    return tiktoken.encoding_for_model(model)
+
+
+def count_tokens(text: str, model: str) -> int:
+    """Count the number of tokens in a text."""
+    if "claude" in model:
+        return anthropic.Anthropic().count_tokens(text)
```
DavidMinarsch, line 40: The benchmark utils needs to be agnostic to this

DavidMinarsch (changes requested):

packages/valory/skills/task_execution/utils/benchmarks.py
0xArdi: Its copied from here https://github.com/valory-xyz/olas-predict-benchmark/blob/main/benchmark/benchmark/utils.py

0xArdi:
Req. changes will be addressed in another PR.

PR #169, opened 2024-02-19
Fix/ tools

0xArdi:
This PR fixes misc issues with tools.

### DIFF
--- a/tools/prediction_request_claude/prediction_request_claude.py
+++ b/tools/prediction_request_claude/prediction_request_claude.py
@@ -119,7 +119,7 @@ OUTPUT_FORMAT
 * Never use Markdown syntax highlighting, such as ```json```. Only output the raw json string.
 * This is incorrect:"```json{{\n  \"queries\": [\"term1\", \"term2\"]}}```"
 * This is incorrect:```json"{{\n  \"queries\": [\"term1\", \"term2\"]}}"```
-* This is correct:"{{\n  \"quries\": [\"term1\", \"term2\"]}}"
+* This is correct:"{{\n  \"queries\": [\"term1\", \"term2\"]}}"
 """
 
 ASSISTANT_TEXT = "```json"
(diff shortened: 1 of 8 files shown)


tools/prediction_request_claude/prediction_request_claude.py
```diff
@@ -119,7 +119,7 @@
 * Never use Markdown syntax highlighting, such as ```json```. Only output the raw json string.
 * This is incorrect:"```json{{\n  \"queries\": [\"term1\", \"term2\"]}}```"
 * This is incorrect:```json"{{\n  \"queries\": [\"term1\", \"term2\"]}}"```
-* This is correct:"{{\n  \"quries\": [\"term1\", \"term2\"]}}"
+* This is correct:"{{\n  \"queries\": [\"term1\", \"term2\"]}}"
```
0xArdi, line 122: typo, which mightve played a role in invalid responses from claude

tools/prediction_with_research_report/prediction_with_research_report.py
```diff
@@ -158,8 +164,32 @@
...
+        texts = [t.replace("\n", " ") for t in texts]
+
+        # Call the OpenAI Embedding API
+        embeddings = self._client.create(input=texts, model=self._model_name).data
+
+        # Sort resulting embeddings by index
+        sorted_embeddings = sorted(embeddings, key=lambda e: e.index)  # type: ignore
+
+        # Return just the embeddings
+        return [result.embedding for result in sorted_embeddings]
```
0xArdi, line 188: Required as the default one is not compatible with openai@v1

tools/prediction_with_research_report/prediction_with_research_report.py
```diff
@@ -158,8 +164,32 @@
...
 # MODELS
 
+MAX_TEXT_LENGTH = 7500
```
0xArdi, line 191: @namesty we need to make sure the length of the text doesnt exceed the limitations of the model. Added this in here for now, please propose a more appropriate number / algorithm to avoid this if you think this is not ok.

angrybayblade (approved):

Adamantios (approved):

PR #184, opened 2024-03-02
fix prediction-request-rag

moarshy:
this fixes errors encountered in mech over the last week.

### DIFF
--- a/packages/napthaai/customs/prediction_request_rag/prediction_request_rag.py
+++ b/packages/napthaai/customs/prediction_request_rag/prediction_request_rag.py
@@ -413,7 +413,7 @@ def extract_texts(urls: List[str], num_words: Optional[int] = None) -> List[Docu
                         doc = extract_text_from_pdf(url, num_words=num_words)
                     else:
                         doc = extract_text(html=result.text, num_words=num_words)
-                    doc.url = url  # Ensure the URL is attached to the Document
+                    doc.url = url
                     extracted_texts.append(doc)
             except Exception as e:
                 print(f"Error processing {url}: {e}")
@@ -479,17 +479,11 @@ def fetch_additional_information(
     docs = [doc for doc in docs if doc]
 
     # remove empty documents ""
-    filtered_docs = []
-    for doc in docs:
-        try:
-            if doc.text != "":
-                filtered_docs.append(doc)
-        except Exception as e:
-            continue
-    
+    filtered_docs = [doc for doc in docs if hasattr(doc, 'text') and doc.text != ""]
+
     # Chunk the documents
     split_docs = []
-    for doc in docs:
+    for doc in filtered_docs:
         try:
             t = recursive_character_text_splitter(
                 doc.text, SPLITTER_CHUNK_SIZE, SPLITTER_OVERLAP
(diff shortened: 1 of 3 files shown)


packages/napthaai/customs/prediction_request_rag/prediction_request_rag.py
```diff
@@ -437,22 +478,45 @@ def fetch_additional_information(
...
     docs = [doc for doc in docs if doc]
 
+    # remove empty documents ""
+    filtered_docs = []
+    for doc in docs:
+        try:
+            if doc.text != "":
+                filtered_docs.append(doc)
+        except Exception as e:
+            continue
```
Adamantios, line 488: What exception could occur when trying to append here?

packages/napthaai/customs/prediction_request_rag/prediction_request_rag.py
```diff
@@ -437,22 +478,45 @@ def fetch_additional_information(
...
     split_docs = [doc for doc in split_docs if doc]
 
+    # remove empty documents ""
+    filtered_split_docs = []
+    for doc in split_docs:
+        try:
+            if doc.text != "":
+                filtered_split_docs.append(doc)
+        except Exception as e:
+            continue
```
Adamantios, line 515: Same question.

packages/napthaai/customs/prediction_request_rag/prediction_request_rag.py
moarshy: I was getting an error when accessing `doc.text` in some cases. wasn't able to identify why

packages/napthaai/customs/prediction_request_rag/prediction_request_rag.py
moarshy: will do without try/except

packages/napthaai/customs/prediction_request_rag/prediction_request_rag.py
Adamantios: So how about:

```python
for doc in docs:
    try:
        text = doc.text
    except {{Add the specific exception here}}:
        # also some message here maybe?
        continue
    if text:
        filtered_docs.append(doc)
```

packages/napthaai/customs/prediction_request_rag/prediction_request_rag.py
Adamantios: Well this is accessing `text` too (https://github.com/valory-xyz/mech/pull/184#discussion_r1510933808).

packages/napthaai/customs/prediction_request_rag/prediction_request_rag.py
moarshy: `filtered_docs = [doc for doc in docs if hasattr(doc, 'text') and doc.text != ""]`

maybe can do like this?

Adamantios: Yes, if the exception was about the attribute not existing that would be much better.

moarshy: changed accordingly

jmoreira-valory (approved):

PR #191, opened 2024-03-19
Improve interface doc & fixes

jmoreira-valory:
Improve interface doc & fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ You can create and mint your own AI Mech that handles requests for tasks that yo
         return result_str, prompt_used, generated_tx, counter_callback
     ```
 
-    - INPUT: Keyword arguments (`**kwargs`). The `kwargs` object is guaranteed to contain the following keys:
+    - **Input**: Keyword arguments (`**kwargs`). The `kwargs` object is guaranteed to contain the following keys:
         - `tool` (`kwargs["tool"]`): A string specifying the (sub-)tool to be used. The `run` command must parse this input and execute the task corresponding to the particular sub-tool referenced. These sub-tools will allow the user to fine-tune the use of your tool.
         - `prompt` (`kwargs["prompt"]`): A string containing the user prompt.
         - `api_keys` (`kwargs["api_keys"]`): A dictionary containing the API keys required by your tool:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -208,15 +208,25 @@ You can create and mint your own AI Mech that handles requests for tasks that yo
...
+            <api_key>=kwargs["api_keys"][<api_key_id>].
+            ```
+
+    - OUTPUT: It must **always** return a tuple (`Tuple[Optional[str], Any, Optional[Dict[str, Any]], Any]`):
+        - `result_str`: A string-serialized JSON object containing the result of the tool execution (custom format).
+        - `prompt_used`: A string representing the prompt used internally by the tool. This output is only used for analytics and it can be set to `None`.
+        - `generated_tx`: A dictionary containing the fields of a generated transaction to be submitted following the execution of the tool (e.g., a token transfer). It can be set to `None`.
+        - `counter_callback`: Object to be called for calculating the cost when making requests to this tool. It can be set to `None`.
+
+    - EXCEPTIONS: The `run` function must not raise any exception. If exceptions occur inside the function, they must be processed and returned appropriately, for example as an error code in `result_str`.
```
0xArdi, line 226: Depends on the case, but exceptions will be handled by the mech itself. In case a of an expcetion, `Invalid response` will be sent.

jmoreira-valory: Addressed.

Adamantios (approved):

dvilelaf (approved):

README.md
```diff
@@ -208,15 +208,35 @@ You can create and mint your own AI Mech that handles requests for tasks that yo
...
+    - **Input**: Keyword arguments (`**kwargs`). The `kwargs` object is guaranteed to contain the following keys:
+        - `tool` (`kwargs["tool"]`): A string specifying the (sub-)tool to be used. The `run` command must parse this input and execute the task corresponding to the particular sub-tool referenced. These sub-tools will allow the user to fine-tune the use of your tool.
+        - `prompt` (`kwargs["prompt"]`): A string containing the user prompt.
+        - `api_keys` (`kwargs["api_keys"]`): A dictionary containing the API keys required by your tool:
+
+            ```python
+            <api_key>=kwargs["api_keys"][<api_key_id>].
+            ```
+
+    - **Output**: It must **always** return a tuple (`Tuple[Optional[str], Any, Optional[Dict[str, Any]], Any]`):
```
0xArdi, line 220: @jmoreira-valory this tuple is wrong.
It should be `Tuple[str, Optional[str], Optional[Dict[str, Any]], Any]`.

PR #192, opened 2024-03-19
fix: transaction response on `resolve_market_reasoning.py`

0xArdi:
This PR fixes the response for the `transaction` on the response on the `resolve_market_reasoning` tool.

### DIFF
--- a/packages/napthaai/customs/resolve_market_reasoning/resolve_market_reasoning.py
+++ b/packages/napthaai/customs/resolve_market_reasoning/resolve_market_reasoning.py
@@ -782,7 +782,7 @@ def run(**kwargs) -> Tuple[Optional[str], Any, Optional[Dict[str, Any]], Any]:
         print(f"Determinable: {determinable_results}")
 
         if not determinable_results.is_determinable:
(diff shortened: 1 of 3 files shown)


packages/napthaai/customs/resolve_market_reasoning/resolve_market_reasoning.py
Adamantios, line 1: Can you explain the change?

packages/napthaai/customs/resolve_market_reasoning/resolve_market_reasoning.py
```diff
@@ -782,7 +782,7 @@ def run(**kwargs) -> Tuple[Optional[str], Any, Optional[Dict[str, Any]], Any]:
         print(f"Determinable: {determinable_results}")
 
         if not determinable_results.is_determinable:
-            return determinable_results.json(), reasoning, additional_information, None
+            return determinable_results.json(), reasoning, None, None
```
0xArdi, line 785: The additional_information is in the wrong place, that is reserved for the optional transcation, which is a Dict.

jmoreira-valory (approved):

PR #201, opened 2024-03-28
Implement a dummy transaction preparation tool

Adamantios:


angrybayblade (approved):

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.4.0 and should not be changed by hand.
```
0xArdi, line 1: Both on my local and CI we use poetry 1.3.2, which may explain why this is happening.

Adamantios: Yeah, it's not causing any issues though.

0xArdi (approved):

PR #203, opened 2024-04-02
Feat/transfer

dvilelaf:
* Ports native transfer tool into prepare_tx tool

### DIFF
--- a/packages/valory/customs/prepare_tx/prepare_tx.py
+++ b/packages/valory/customs/prepare_tx/prepare_tx.py
@@ -22,30 +22,107 @@ This module implements a tool which prepares a transaction for the transaction s
 Please note that the gnosis safe parameters are missing from the payload, e.g., `safe_tx_hash`, `safe_tx_gas`, etc.
 """
 
-import json
+from openai import OpenAI
 from typing import Any, Dict, Optional, Tuple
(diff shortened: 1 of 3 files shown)


packages/valory/customs/prepare_tx/prepare_tx.py
```diff
@@ -22,30 +22,107 @@
...
+    transaction = {
+        "to": str(parsed_txs["to"]),
+        "value": int(parsed_txs["wei_value"]),
+    }
+    return response, prompt, transaction, None
 
 
-AVAILABLE_TOOLS = {"prepare_tx": prepare_tx}
+AVAILABLE_TOOLS = {
+    "native_transfer": native_transfer,
```
dvilelaf, line 124: Does this need to match the file name? I'm a bit confused about tool name vs available tools names

0xArdi: No it does not

0xArdi (approved):

Adamantios (approved):
Isn't that the same as the native transfer tool?

dvilelaf:
> Isn't that the same as the native transfer tool?

One is "native", this one is "native_request"

PR #213, opened 2024-04-16
fix: funds splitting logic

0xArdi:
This PR fixes the funds splitting logic.


packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -529,7 +541,7 @@ def _accumulate_reqs_by_operator(
         agent_instances = list(reqs_by_agent.keys())
         contract_api_msg = yield from self.get_contract_api_response(
             performative=ContractApiMessage.Performative.GET_STATE,  # type: ignore
-            contract_address=self.params.agent_registry_address,
+            contract_address=self.params.service_registry_address,
```
0xArdi, line 544: not sure how this was ever working with this issue

dvilelaf (approved):

packages/valory/skills/task_submission_abci/behaviours.py
Adamantios: Remember to set the env var when deploying, because the default is zero address:

https://github.com/valory-xyz/mech/blob/cd20e6bb8cf86fbdfa15ee577d6984e60030502b/packages/valory/services/mech/service.yaml#L39

Adamantios (approved):

PR #216, opened 2024-04-19
Feat/pessimistic delivery

0xArdi:
This PR adds support for pessimistic delivery of mech requests.


packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -93,6 +93,18 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
...
+            # Set to store unique request_ids
+            unique_ids = set()
+            unique_objects = []
+
+            # filter out the tasks that have duplicate ids
+            for obj in all_done_tasks:
+                request_id = obj.get("request_id")
+                if request_id not in unique_ids:
+                    unique_ids.add(request_id)
+                    unique_objects.append(obj)
```
0xArdi, line 106: This is causing the recent issues, duplicate req ids.

packages/valory/contracts/agent_mech/contract.py
```diff
@@ -441,3 +454,27 @@ def get_set_subscription_tx_data(
...
+        try:
+            ledger_api.api.eth.call(
+                {
+                    "from": ledger_api.api.to_checksum_address(sender_address),
+                    "to": ledger_api.api.to_checksum_address(contract_address),
+                    "data": data,
+                }
+            )
+            simulation_ok = True
+        except Exception as e:
```
0xArdi, line 476: we need a broad exception because on some chains web3py cannot handle the reverts properly

packages/valory/contracts/agent_mech/contract.py
```diff
@@ -441,3 +454,27 @@ def get_set_subscription_tx_data(
...
+            _logger.info(f"Simulation failed: {str(e)}")
```
Adamantios, line 477: This does not follow the configurations of the agent's logger. It would be preferable to return a message and log from the behaviour side.

Adamantios (approved):

packages/valory/contracts/agent_mech/contract.py
0xArdi: @Adamantios why does it not? Configurations are done based on the logger path, and this logger is initialized with the correct agent path. Checking the logs, it looks correct, example:
```
2024-04-19T15:31:05.579933068Z stderr F [2024-04-19 15:31:05,579][INFO] Simulation failed: non-hexadecimal number found in fromhex() arg at position 0
```

dvilelaf (approved):

PR #219, opened 2024-04-24
auto-infer llm provider

richardblythman:


tests/test_tools.py
```diff
@@ -82,19 +82,12 @@ def test_run(self) -> None:
...
-                    elif "claude" in model:
-                        llm_provider = "anthropic"
-                    else:
-                        llm_provider = "openrouter"
                     kwargs = dict(
                         prompt=prompt,
                         tool=tool,
                         api_keys=self.keys,
                         counter_callback=TokenCounterCallback(),
                         model=model,
```
0xArdi, line 90: Thats good. We shouldn't update this function call here, otherwise we run into mismatches with the mech.

0xArdi: This is what lead to this issue not being detected in the first place.

packages/valory/customs/prediction_request/prediction_request.py
```diff
@@ -44,9 +44,14 @@
...
-    def __init__(self, api_keys: List, llm_provider: str = None):
+    def __init__(self, api_keys: List, model: str = None):
         self.api_keys = api_keys
-        self.llm_provider = llm_provider
+        if "gpt" in model:
+            self.llm_provider = "openai"
+        elif "claude" in model:
+            self.llm_provider = "anthropic"
+        else:
+            self.llm_provider = "openrouter"
```
0xArdi, line 54: how can these two be practically ever called?

0xArdi: the traders don't send a model paramater, so effectively only the default model will always be used. This kinda makes the tools "claude" and "openrouter" tools unselectable from the perspective of the trader

tests/test_tools.py
richardblythman: run() no longer takes llm_provider as a kwarg though. 
Why would there be a mismatch with the mech?

packages/valory/customs/prediction_request/prediction_request.py
0xArdi: I understand, but the traders only provide a tool, not a model. So this would always result in `openai` being used. Effectively making the claude tools unusable.

richardblythman: Isn't model passed as a kwarg to the tool?

0xArdi: The trader don't send a model, so the mech will always pass the default one. [Example request](https://gateway.autonolas.tech/ipfs/f0170122054ADDF9AE79D2FA3C708D81FCCF561EB211ACA849852E79C9350F247FADCABF6/metadata.json) from the trader.

tests/test_tools.py
0xArdi: It's fine the way that it is now, Im just pointing out why the test didn't catch this problem.

packages/valory/customs/prediction_request/prediction_request.py
richardblythman: What about [this](https://github.com/valory-xyz/mech/blob/6dcf1244f276859d923443d5019f4dc8dd3b93de/packages/valory/skills/task_execution/behaviours.py#L402)?

tests/test_tools.py
richardblythman: Oh sorry I misunderstood. I thought you meant not to remove these lines as I did.
Agreed - we shouldn't change any kwargs in tests.

packages/valory/customs/prediction_request/prediction_request.py
0xArdi: As you can see the default model gets passed if a model is not present. The traders DO NOT send a model parameter, hence the mech will always pass the default model.

richardblythman: The default model of some tools is claude, some openai. Could set it to whatever. 
And we use a model kwarg when using mechx.

richardblythman: Also, since prediction-request and prediction-request-claude have been unified now, tool selection of the trader should probably use the model kwarg.

0xArdi: I understand that, but the trader currently does not use a model. In the current setup, the claude tool cannot be used in prediction-online or offline by the trader. I think this is an issue.

richardblythman: But if we pass the wrong model kwarg with mechx, we would get an error without this code.

0xArdi: My only concern is that the traders will not be able to use claude for prediction thats it

richardblythman: The trader could use Claude for prediction if they pass it as a model kwarg right?
Maybe I'm misunderstanding something here.

0xArdi: The trader DOES NOT pass a model parameter. I shared how requests from the trader look like.

richardblythman: Then change the trader to pass a model parameter. 

Why is trader is relevant here? This is a PR to mech. Mechx already passes a model parameter. 

I don't understand what the issue is. But let me know what you want me to remove and I can do it.

cyberosa: let me check on the main repo of the trader agent which are the available tools of the available_tools_store.json to see which is the right answer here

poetry.lock
```diff
@@ -1,88 +1,88 @@
-# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
```
0xArdi, line 1: likely related to the diff version change, however the lock seems fine

0xArdi (approved):

PR #295, opened 2025-06-03
Dynamic pricing for nvm mechs

KahanMajmudar:
## Proposed changes

Adds dynamic pricing for nvm mechs

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [x] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...


Adamantios (approved):

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -401,6 +409,9 @@ def _execute_task(self) -> None:
             request_id = task_data["requestId"]
             task_data["requestId"] = int.from_bytes(request_id, byteorder="big")
 
+        request_id = task_data["requestId"]
+        delivery_rate = task_data["request_delivery_rate"]
```
OjusWiZard, line 413: **Unconditional key access — KeyError risk.** `_execute_task` runs for *all* pending tasks, but `request_delivery_rate` is only populated by two producers (the marketplace path in `get_marketplace_undelivered_reqs` and the offchain path in `_handle_signed_requests`). Any other producer that appends to `pending_tasks` without this key will crash task execution here. Please verify every `pending_tasks.append(...)` site sets it, or make this defensive: `task_data.get("request_delivery_rate", 0)`.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -152,6 +154,11 @@ def ipfs_tasks(self) -> List[Dict[str, Any]]:
         """Get ipfs_tasks."""
         return self.context.shared_state[IPFS_TASKS]
 
+    @property
+    def request_id_to_delivery_rate_info(self) -> List[Dict[str, int]]:
```
OjusWiZard, line 158: **Wrong type hint.** This is initialized as `{}` in `handlers.setup()` and used as a map (`self.request_id_to_delivery_rate_info[request_id] = delivery_rate`). It's a `Dict[int, int]`, not a `List[Dict[str, int]]`.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -672,15 +697,19 @@ def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> No
...
-            ).hex()
+        tool = str(done_task.get("tool"))
+        dynamic_tool_cost = self._tools_to_pricing.get(tool)
+        if dynamic_tool_cost is not None:
+            self.context.logger.info(
+                f"Tools to pricing found for tool {tool}. Adding dynamic pricing of {dynamic_tool_cost} for request id {req_id}"
+            )
+            done_task["dynamic_tool_cost"] = dynamic_tool_cost
+        else:
+            dynamic_tool_cost = 0
```
OjusWiZard, line 708: **Dead assignment.** With the old `mech_config.use_dynamic_pricing` encoding block removed, this local `dynamic_tool_cost = 0` is never read again (only `done_task["dynamic_tool_cost"]`, set in the `if` branch, is consumed downstream). The `cost = get_cost_for_done_task(...)` below is also computed solely to log it. Consider dropping the dead local, and confirm the `cost` log is still intended.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -573,6 +597,7 @@ def _submit_task(self, fn: Any, *args: Any, **kwargs: Any) -> Future:
 
     def _prepare_task(self, task_data: Dict[str, Any]) -> None:
         """Prepare the task."""
+        self.context.logger.info(f"Preparing tool task with data: {task_data}")
```
OjusWiZard, line 600: **Verbose INFO log.** `task_data` includes the full request/prompt payload and can be large. Consider DEBUG level, or log just the keys / request id.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -286,10 +288,12 @@ def _handle_signed_requests(
             ipfs_hash = data["ipfs_hash"]
             request_id = data["request_id"]
             ipfs_data = data["ipfs_data"]
+            request_delivery_rate = data["delivery_rate"]
```
OjusWiZard, line 291: Same KeyError exposure as the `_execute_task` access: `data["delivery_rate"]` is required here. If any signed-request producer omits it, this raises. Worth a defensive default unless the upstream contract guarantees the field.

packages/valory/contracts/agent_mech/contract.py
```diff
@@ -667,6 +673,13 @@ def get_marketplace_undelivered_reqs(
             if request["requestId"] not in [
                 deliver["requestId"] for deliver in delivers
             ]:
+                # fetch and store max delivery rate for each request id
+                request_id_info = MechMarketplaceContract.get_request_id_info(
```
OjusWiZard, line 677: **N+1 contract calls + no error guard.** This issues one live `get_request_id_info(...).call()` per undelivered request inside the loop. With a large backlog this multiplies RPC load per poll, and an unguarded failure on any single call aborts the whole undelivered-reqs fetch. Consider batching, or wrapping the per-request call so one bad request id doesn't drop the entire batch.

OjusWiZard (commented):
Retrospective review (post-merge). Core mechanism — per-tool pricing with the charged rate capped at the on-chain delivery rate — is sound. A few correctness/quality notes inline.

**Not line-specific:** no tests were added (PR checklist confirms). The new payment-affecting surface is untested: `min()` rate-capping in `get_encoded_data_for_request`, the `request_id_delivery_rate < tool_pricing` rejection in `_handle_get_task`, and `_ensure_same_keys` config validation. For a breaking feature that changes what requesters are charged, worth covering in a follow-up.

packages/valory/contracts/agent_mech/contract.py
```diff
@@ -624,6 +629,7 @@ def get_marketplace_undelivered_reqs(
         cls,
         ledger_api: LedgerApi,
         contract_address: str,
+        marketplace_address: str,
```
OjusWiZard, line 632: New **required** parameter inserted mid-signature. Low risk here since the contract API dispatches by keyword (the one caller passes `marketplace_address=...`), but any positional caller would break. Flagging for awareness.

packages/valory/contracts/mech_marketplace/contract.py
```diff
@@ -369,20 +369,36 @@ def get_encoded_data_for_request(
...
         request_id: bytes,
         data: str,
+        delivery_rate: int,
     ) -> JSONLike:
         """Fetch info for a given request id."""
-        contract_instance = cls.get_instance(ledger_api, contract_address)
+        request_id_info = cls.get_request_id_info(
+            ledger_api, contract_address, request_id
+        )
+        final_delivery_rate = min(
```
OjusWiZard, line 378: Capping the charged rate at the on-chain max (`min(on_chain_rate, delivery_rate)`) is the right call — prevents charging above what the requester authorized.

PR #345, opened 2025-08-26
Check timedout list status

Totoual:
## Proposed changes

Extend the step in functionality to further check the status code 3 for prio mech, and refresh the timedout list based on status changes..

## Fixes

If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

### DIFF
--- a/packages/valory/skills/task_execution/handlers.py
+++ b/packages/valory/skills/task_execution/handlers.py
@@ -51,6 +51,7 @@ WAIT_FOR_TIMEOUT = "wait_for_timeout"
 REQUEST_ID_TO_DELIVERY_RATE_INFO = "request_id_to_delivery_rate_info"
 TIMED_OUT_STATUS = 2
 WAIT_FOR_TIMEOUT_STATUS = 1
+DELIVERED_STATUS = 3
 
 LEDGER_API_ADDRESS = str(LEDGER_CONNECTION_PUBLIC_ID)
 
@@ -184,6 +185,12 @@ class ContractHandler(BaseHandler):
         """Get timed_out_tasks for other mechs"""
         return self.context.shared_state[TIMED_OUT_TASKS]
 
+    @timed_out_tasks.setter
+    def timed_out_tasks(self, value) -> None:
+        """Get timed_out_tasks for other mechs"""
+        self.context.shared_state[TIMED_OUT_TASKS] = value
+        return
+
     def handle(self, message: Message) -> None:
         """
         Implement the reaction to a contract message.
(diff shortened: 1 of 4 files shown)


packages/valory/skills/task_execution/handlers.py
```diff
@@ -293,7 +306,10 @@ def _validate_and_flatten(self, body: Dict[str, Any]) -> List[Dict[str, Any]]:
     def filter_requests(self, reqs: List[Dict[str, Any]]) -> None:
         """Filtering requests based on priority mech and status."""
         for req in reqs:
-            if req["priorityMech"].lower() == self.mech_address.lower():
+            if (
+                req["priorityMech"].lower() == self.mech_address.lower()
+                and req["status"] != DELIVERED_STATUS
```
mariapiamo, line 311: can we have a comment here saying that, since reqs is obtained by looking at contracts events, we will not expect req["status"] =  non existed status?

mariapiamo (commented):
ci to be fixed and a comment

packages/valory/skills/task_execution/handlers.py
Totoual: I mean, if we don't expect to see status 0 ever inside the mech, I don't see a reason to have a comment that is not connected with a part of the code. If you insist though can add it :D

packages/valory/skills/task_execution/handlers.py
```diff
@@ -293,7 +305,10 @@ def _validate_and_flatten(self, body: Dict[str, Any]) -> List[Dict[str, Any]]:
     def filter_requests(self, reqs: List[Dict[str, Any]]) -> None:
```
Adamantios, line 305: Unrelated to the PR. This method could use some early returns. Also, no need for this nested condition:

```python
elif req["status"] == WAIT_FOR_TIMEOUT_STATUS:
    if req["request_delivery_rate"] >= self.mech_to_max_delivery_rate:
```

PR #429, opened 2026-04-15
Fix/attack surface hardening

DIvyaNautiyal07:
# Add input validation and request-deadline handling in task_execution

## Summary

Addresses the [mech audit](https://github.com/valory-xyz/launch-predict/blob/main/mech-audit/AUDIT.md). Tightens input validation on the two places the agent ingests external data (IPFS-delivered task payloads and the off-chain HTTP handler) and adds a deadline to the payment-model request so a missing response can no longer block task execution indefinitely.

## Changes

Split across six commits for easier review:

1. Cap IPFS task payload size. Reject combined file content over 1 MB in `_handle_get_task` before JSON parsing runs.
2. Validate task_data prompt type and length. Require `prompt` and `tool` to be strings and cap `prompt` at 100 KB.
3. Add deadline to payment-model request. If the contract response for the payment model does not arrive within 60s, clear `in_flight_req` so the request can be retried instead of wedging `_ensure_payment_model`.
4. Validate offchain HTTP request body and ipfs_hash format. Cap inbound bodies at 1 MB in `_parse_http_body`, validate `ipfs_hash` matches `^0x[0-9a-fA-F]+$` with length 66 or 70 before parsing, and stop logging raw request bodies on failure.
5. Bound delivery_rate range. Reject negative and above-uint256 values on the off-chain HTTP path. `delivery_rate=0` remains accepted for free tasks.
6. Update package hashes. Mechanical lock bump.

## Test plan

- [ ] `make code-checks` (black, isort, flake8, mypy, pylint, darglint)
- [ ] `pytest packages/valory/skills/mech_abci/tests packages/valory/skills/task_submission_abci/tests packages/valory/skills/task_execution/tests` (532 pass, 17 new)
- [ ] `tox -e check-hash` and `tox -e check-packages`
- [ ] `tox -e check-abci-docstrings`, `tox -e check-abciapp-specs`, `tox -e check-handlers`
- [ ] `autonomy push-all --remote` before pushing the branch

## New test coverage

Behaviour-level:

- Oversized IPFS payload rejection, spying on `json.loads` to confirm the cap gates before parse
- Non-string `prompt` rejection
- Over-cap `prompt` length rejection. Both prompt tests spy on the designated-mech lookup to confirm the check gates the happy path
- Payment-model `in_flight_req` reset after timeout
- Payment-model `in_flight_req` preserved within timeout window
- Payment-model `in_flight_req` untouched when no payment-model timestamp exists

Handler-level:

- Oversized HTTP body rejection, spying on `urllib.parse.parse_qs` to confirm the cap trips before decode
- Invalid `ipfs_hash` rejection across 5 malformed shapes (too short, wrong length, too long, missing prefix, non-hex)
- Out-of-range `delivery_rate` rejection (negative, above uint256)
- `delivery_rate=0` accepted for free tasks

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -8,12 +8,12 @@
         "connection/valory/websocket_client/0.1.0": "bafybeid4cnbjpmzcsw3usg7umufgqyod2u3xfw5uso3vukt3n5hfuxvdea",
         "skill/valory/contract_subscription/0.1.0": "bafybeibtcbtgm5ptzn6srfts3lxts5722sygmkypcknv5ocmydqwcvgeem",
         "skill/valory/delivery_rate_abci/0.1.0": "bafybeifpoqkcdwpyq4lijumyt3gzj3vs3pfxxm4xefqxagd4b5pg6x3vty",
-        "skill/valory/mech_abci/0.1.0": "bafybeieqqsopesprm2xvyhyawwnz2qwplwqyuaqu6nzwmaps524ieh5mja",
-        "skill/valory/task_execution/0.1.0": "bafybeiecggxojaqpseamguzivxeyq52okl2onbeddgp4x6t2qz6kl2nyvy",
-        "skill/valory/task_submission_abci/0.1.0": "bafybeihxp4jhz4r2ant3d6hj5izcs7h3zkbkfwgajxfzvaubzv6vntva3e",
+        "skill/valory/mech_abci/0.1.0": "bafybeih54mcn5pp2uanbkurgd5vnk2paky6z7plfe6ygl24dyvtc4ipryy",
+        "skill/valory/task_execution/0.1.0": "bafybeig7pzey3umn2uub6gkafnam5znwic4skziyp2l5wcqcp2xmljcwde",
+        "skill/valory/task_submission_abci/0.1.0": "bafybeig27jre4yhwyugi6mynm3bv5b5t3nba34amyebeatyd44ozdxesbq",
         "skill/valory/websocket_client/0.1.0": "bafybeie63j65ffblz6g6oiaynjpz4ae5mgyjz2w6o5ovqitzdbnvqzlhzu",
-        "agent/valory/mech/0.1.0": "bafybeidrfivkhnip3kfd2i6dxfgmidqyriualayarnjrdv7x3v3u26lcxu",
-        "service/valory/mech/0.1.0": "bafybeief4vqbogw6yvud2emtrzkpp3exslnzqjdqsb5vu5d4ygbo2axsse"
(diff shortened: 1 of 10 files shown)


Adamantios (approved):

LOCKhart07 (approved):
Approving — clean security hardening, caps placed before parse/decode, tests exercise real logic. A few non-blocking nits for follow-up:

1. `behaviours.py` `_handle_get_task`: `executing_task = cast(...)` + `req_id = ...` is now computed three times (top, inside the `except`, and after `is_data_valid`). The new top-of-function version makes the later recomputes dead — consolidate to one.

2. `behaviours.py` byte-count helpers: `.encode("utf-8", errors="replace")` on a `str` is unreachable in Python 3 (str is always valid Unicode). Drop the `errors=` kwarg on both the `total_bytes` sum and the `prompt_bytes` calc.

3. `handlers.py`: `IPFS_HASH_RE = r"^0x[0-9a-fA-F]+$"` accepts any length; the real constraint is `VALID_IPFS_HASH_LENGTHS`. Could fold into one regex `^0x([0-9a-fA-F]{64}|[0-9a-fA-F]{68})$` and drop the length tuple.

4. `_payment_model_request_sent_at` is only cleared the next time `_ensure_payment_model` runs. Harmless (nothing depends on it once `payment_model` is set) but tidier to clear it wherever the contract response sets `payment_model`.

5. `test_ensure_payment_model_does_not_reset_fresh_inflight_request` asserts `_payment_model_request_sent_at is not None` but not that it's unchanged — comparing against the pre-call value would catch accidental resets that happen to leave a non-None value.

Worth calling out as a real fix (not just hardening): reordering `_ensure_payment_model` to check `payment_model` before `in_flight_req` closes a deadlock where both could be set simultaneously. Good catch.

DIvyaNautiyal07:
Thanks for the review. Addressed 1, 2, 3, 5 in the latest commit; skipping 4 per your "harmless" note.

1. Consolidated `executing_task` / `req_id` to a single computation at the top of `_handle_get_task`. Removed the duplicates inside the JSON-parse `except` and after the `is_data_valid` block.
2. Dropped `errors="replace"` on both `total_bytes` and `prompt_bytes` `.encode("utf-8")` calls.
3. Folded length and format into one regex `^0x([0-9a-fA-F]{64}|[0-9a-fA-F]{68})$` and removed `VALID_IPFS_HASH_LENGTHS`.
4. Skipping. The lazy clear in `_ensure_payment_model` is fine and keeps the cross-module coupling out of `ContractHandler`.
5. `test_ensure_payment_model_does_not_reset_fresh_inflight_request` now captures the timestamp before the call and asserts equality after, so a mutation that resets to a different non-None value would fail.

And yes, good catch on the reorder reading as a real fix rather than just hardening. The previous order (`in_flight_req` before `payment_model`) returned False even when the payment model was already cached, which was a latent issue.

PR #435, opened 2026-04-22
Fix/multisend str vs bytes

DIvyaNautiyal07:
## Summary

- Two mech contract helpers (`hash_checkpoint.get_checkpoint_data`, `complementary_service_metadata.get_update_hash_tx_data`) were returning `encode_abi(...)` as a hex string instead of bytes. Open-autonomy 0.21.18 removed the `HexBytes(data)` wrapper inside `multisend.encode_data` that had been silently coercing those strings to bytes. After the 0.21.19 bump, every delivery cycle hit `TypeError: can't concat str to bytes` inside `encode_data`, the dispatcher swallowed it into an ERROR performative, and the skill logged only the generic `Couldn't compile the multisend tx` line.
- Fix each helper to return `bytes.fromhex(data[2:])`, matching the pattern already used in `olas_mech` helpers.

Key findings:
- The bug is on our side. The two helpers have been returning `str` since they were written. The old multisend contract was quietly fixing it for us via `HexBytes(data)` coercion. The upstream refactor was a dependency cleanup ("drop type-only web3/eth_typing imports across 16 contracts") that did not flag the removed coercion as a behaviour change.
- The sibling helper `olas_mech.get_exec_tx_data` already does `bytes.fromhex(data[2:])`, so the correct pattern existed in the same codebase.
- The log line at `task_submission_abci/behaviours.py:1407` did not include `response.message`. Once the error became active, the logs alone were insufficient to diagnose the failure.

### `packages/valory/contracts/hash_checkpoint/contract.py`
`get_checkpoint_data` returns `{"data": bytes.fromhex(data[2:])}` instead of `{"data": data}`. Used on every delivery cycle via `get_update_usage_tx`.

### `packages/valory/contracts/complementary_service_metadata/contract.py`
`get_update_hash_tx_data` returns `dict(data=bytes.fromhex(data[2:]))` instead of `dict(data=data)`. Used intermittently via `get_mech_update_hash_tx` when the on-chain metadata hash drifts.

### `packages/valory/skills/task_submission_abci/behaviours.py`
`_to_multisend` now logs the full response on non-`RAW_TRANSACTION` performatives:

```python
self.context.logger.error(f"Couldn't compile the multisend tx: {response}")
```

`ContractApiMessage.__str__` includes the dispatcher's `code` and `message`, so the underlying exception is now visible. Matches the existing pattern at [behaviours.py:1486](packages/valory/skills/task_submission_abci/behaviours.py#L1486) (`f"get_deliver_data unsuccessful!: {contract_api_msg}"`).


Adamantios (dismissed):

OjusWiZard (dismissed):
was that not already covered by the unit tests?

packages/valory/contracts/hash_checkpoint/contract.py
```diff
@@ -104,7 +104,7 @@ def get_checkpoint_data(
 
         contract_instance = cls.get_instance(ledger_api, contract_address)
         data = contract_instance.encode_abi(abi_element_identifier="checkpoint", args=[data])
-        return {"data": data}
+        return {"data": bytes.fromhex(data[2:])}  # type: ignore
```
bennyjo, line 107: Was it deliberate to drop `test_contract_tx_data_bytes.py` in `3dd4c44`? The failure mode here is specifically "helper returns `str`, upstream silently coerces, regression goes undetected for months" — which is the exact class of thing a one-line `assert isinstance(result["data"], bytes)` gate protects against. Live delivery on service 2692 confirms current state but doesn't guard future refactors.

The commit message points at behaviour-level coverage as redundant, but if the task-submission behaviour tests mock the contract dispatcher directly, they'd inject a bytes value at the boundary and never actually exercise this helper's return type. Happy to leave as-is if you've confirmed the behaviour tests exercise the real helper path (or if the invariant is considered out of scope for this PR) — flagging because the same thing applies to `get_update_hash_tx_data` at `complementary_service_metadata/contract.py:140`, and both helpers are back to being un-pinned against a str/bytes regression.

bennyjo (commented):
Flagging one question about the regression-test deletion — see inline.

LOCKhart07 (approved):

packages/valory/contracts/hash_checkpoint/contract.py
DIvyaNautiyal07: Valid catch, you're right. Confirmed the skill tests all mock `get_contract_api_response` at the skill boundary, so neither helper's real return path runs in CI. I dropped the original file in 3dd4c44 for a structural reason (contract-level test under a skill tests dir), but the invariant should still be pinned.

Added a `TestContractHelpersReturnBytes` class in `test_behaviours.py` with one `isinstance(result["data"], bytes)` assertion per helper. Mutation-verified by reverting each `bytes.fromhex` in turn and watching the matching test fail.

Kept it in the skill tests dir because first-party contracts don't have `tests/` infrastructure today, and adding it for only these two would establish a new convention mid-repo. Docstring on the class explains the rationale. Happy to move to `packages/valory/contracts/<name>/tests/` (with a `CONTRACTS_PATHS` wire-in to `tox.ini`) if you'd rather we set that precedent here.

OjusWiZard (approved):

Adamantios (approved):

PR #436, opened 2026-04-23
chore: minimise runtime dependency footprint

DhairyaPatel7:
## Summary

Wave 1 of the fleet-wide dependency-minimisation pass (tracking
[open-autonomy#2477](https://github.com/valory-xyz/open-autonomy/pull/2477))
applied to `mech`.

- **Contracts**: custom contracts (`olas_mech`, `hash_checkpoint`,
  `balance_tracker`) reach through `ledger_api.api.*` instead of direct
  `web3` / `eth_abi` / `eth_utils` / `hexbytes` / `eth_typing` imports.
  All of those imports are gone from `olas_mech/contract.py`; the dead
  `web3.types` import is removed from `hash_checkpoint/contract.py`;
  the unused `web3` YAML dep is pruned from `balance_tracker`.
- **Skills**: dead `web3` / `eth-abi` imports and YAML declarations
  removed from `websocket_client`, `task_execution`, `task_submission_abci`.
  `contract_subscription` is intentionally left alone — it uses
  `Web3.HTTPProvider` at runtime.
- **Agent config**: `agents/mech/aea-config.yaml` dep block trimmed of
  14 deps that no longer have callers. `protobuf` widened to `<7,>=5`
  and `py-ecc` to `<10,>=8` to match the post-2477 vendored packages.
- **pyproject.toml / tox.ini**: 15 stale direct deps removed;
  4 dead `[mypy-*]` ignore sections dropped.
- **CI parity**: new `tox -e check-third-party-hashes` env mirrors the
  agent-repo pattern, pinned to `open-autonomy@0.21.19` +
  `open-aea@2.2.1`. Wired into `common_checks.yaml` so the repo now
  verifies its `third_party` hashes against upstream on every PR.

### HexBytes regression check (from mech PR #435)

OA PR 2477 removed the `HexBytes(data)` coercion from
`multisend.encode_data`. This surfaced in mech as
`TypeError: can't concat str to bytes` on every delivery cycle.
PR [#435](https://github.com/valory-xyz/mech/pull/435) fixed the
two helpers that returned hex strings (`hash_checkpoint.get_checkpoint_data`,
`complementary_service_metadata.get_update_hash_tx_data`). This PR
builds on that fix — the `bytes.fromhex(...)` idiom is now the norm
across every affected helper in `olas_mech/contract.py` too.

## Test plan

- [x] `tox -e check-hash` — verified locally
- [x] `tox -e check-packages` — verified locally
- [x] `tox -e check-dependencies` — verified locally (no issues)
- [x] `tox -e check-third-party-hashes` — verified locally (27/27 consistent)
- [x] `tox -e black-check`
- [x] `tox -e isort-check`
- [x] `tox -e flake8`
- [x] `tox -e mypy` — 64 files, no issues
- [x] `tox -e pylint`
- [x] `tox -e darglint`
- [ ] Live delivery smoke test on Gnosis (reviewer to confirm — the
  regression fixed by PR #435 is retained, no new bytes/str footguns
  introduced)

## Related

- open-autonomy#2477 — upstream fleet-wide dependency cleanup
- mech#435 — the narrow fix for the 0.21.19 multisend regression


packages/valory/agents/mech/aea-config.yaml
```diff
@@ -117,42 +105,30 @@ dependencies:
...
-    version: <6,>=5
+    version: <7,>=5
   py-ecc:
-    version: ==8.0.0
-  py-eth-sig-utils:
-    version: ==0.4.0
+    version: <10,>=8
   py-multibase:
     version: ==1.0.3
   py-multicodec:
```
DavidMinarsch: what still uses these?

packages/valory/contracts/hash_checkpoint/contract.py
```diff
@@ -18,15 +18,13 @@
...
 
 from aea.common import JSONLike
 from aea.configurations.base import PublicId
 from aea.contracts.base import Contract
 from aea.crypto.base import LedgerApi
 from aea_ledger_ethereum import EthereumApi
-from web3.types import BlockIdentifier, TxReceipt
 
 
 class HashCheckpointContract(Contract):
```
DavidMinarsch, line 31: There's no changes below?

packages/valory/contracts/olas_mech/contract.py
```diff
@@ -302,18 +296,18 @@ def get_request_events(
...
         for abi in partial_abis:
             contract_instance = ledger_api.api.eth.contract(contract_address, abi=abi)
-            event_abi = contract_instance.events.Request().abi
+            event = contract_instance.events.Request()
             entries = cls.get_event_entries(
                 ledger_api=ledger_api,
-                event_abi=event_abi,
+                event=event,
                 address=contract_instance.address,
                 from_block=from_block,
```
DavidMinarsch, line 318: wasn't this the issue it's meant to fix: fromBlock

packages/valory/skills/task_execution/skill.yaml
```diff
@@ -115,8 +115,6 @@ dependencies:
     version: ==1.0.3
   py-multicodec:
     version: ==0.2.1
```
DavidMinarsch: The skill should replace these as well - so needs to be extended with what we did on open-aea PR as well

packages/valory/skills/task_submission_abci/skill.yaml
```diff
@@ -181,8 +181,6 @@ models:
     args: {}
     class_name: TendermintDialogues
 dependencies:
-  web3:
-    version: <8,>=7.0.0
   py-multibase:
     version: ==1.0.3
   py-multicodec:
```
DavidMinarsch: see above

packages/valory/skills/websocket_client/handlers.py
```diff
@@ -25,8 +25,6 @@
 
 from aea.protocols.base import Message
 from aea.skills.base import Handler
-from web3 import Web3
-from web3.types import TxReceipt
 from typing import Callable, cast
 from packages.valory.protocols.websocket_client.message import WebsocketClientMessage
 from enum import Enum
```
DavidMinarsch: something wrong here with litners, typing and enum are stdlibs

tox.ini
```diff
@@ -280,6 +280,17 @@ commands =
...
     aea-helpers check-dependencies --check --exclude requests --exclude pyyaml --exclude PyYAML
 
+[testenv:check-third-party-hashes]
+skipsdist = True
+skip_install = True
+deps =
+    open-aea==2.2.1
+    open-aea-ci-helpers==2.2.1
+commands =
+    aea-ci check-third-party-hashes \
```
DavidMinarsch, line 276: nice!

packages/valory/agents/mech/aea-config.yaml
DhairyaPatel7: All four surviving deps in this block have live callers in the agent's transitive package set:

- `pebble` → `skills/task_execution/behaviours.py:36` (`from pebble import ProcessPool`)
- `prometheus_client` → `skills/task_execution/handlers.py:33`, `.../behaviours.py:37`, `skills/task_submission_abci/behaviours.py:32`, `.../models.py:26`
- `protobuf` → transitive through `connections/abci/connection.py:46` and `.../tendermint_decoder.py:26` (`google.protobuf.*`)
- `py-ecc` → `skills/abstract_round_abci/utils.py:50` (`from py_ecc.bls import G2Basic as bls`)

`py-multibase` and `py-multicodec` are gone after 4f1e3e33 — see the reply on the `task_execution/skill.yaml` thread for the migration to `aea.helpers.multiformat`.

packages/valory/contracts/hash_checkpoint/contract.py
DhairyaPatel7: Correct — those two imports were genuinely dead. `grep -n 'BlockIdentifier\|TxReceipt' contract.py` returns nothing; the file's public surface only uses `LedgerApi` / `JSONLike` / `bytes` / `Any`. No signatures referenced the removed web3 types, so no downstream retyping was needed.

packages/valory/contracts/olas_mech/contract.py
DhairyaPatel7: This PR isn't a fromBlock fix — that landed earlier and is preserved here:

- `364d653a` (2025-05-28, #293 "Remove filter and use filterParams") replaced the stateful `event.create_filter(fromBlock=...)` with stateless `eth.get_logs({"fromBlock": ...})`.
- `33b907cd` (2025-08-19) fixed a latent bug where `get_event_entries` hardcoded `SafeReceived` regardless of which event the caller wanted.

The filter dict still uses `"fromBlock"` (camelCase) at `contract.py:762` — unchanged. What this PR touches on this function is only two type-import cleanups:

1. `BlockIdentifier` → `Union[int, str]` (signature-level), so we can drop the `web3.types` import.
2. `event_abi=…` → `event=…` (helper param), so we can use `event.topic` + `event.process_log()` instead of `eth_utils.event_abi_to_log_topic` + `web3._utils.events.get_event_data`.

Both replacements were verified on web3 7.14.1 (the version resolved in `uv.lock`) and produce byte-identical output to the old pair.

Follow-up 4f1e3e33 reintroduces `BlockIdentifier = Union[int, str]` as a local module alias at the top of the file so the public signatures keep the intent-documenting name without reintroducing the `web3.types` import. Audit of all 11 `from_block=` call sites in mech confirmed every caller passes either an `int` block number or one of the literals `"earliest"` / `"latest"` — `Union[int, str]` is strictly sufficient coverage.

packages/valory/skills/task_execution/skill.yaml
DhairyaPatel7: Done in 4f1e3e33. `skills/task_execution/utils/ipfs.py` now uses `aea.helpers.multiformat.multibase_decode` and `aea.helpers.multiformat.multicodec_remove_prefix` (introduced in valory-xyz/open-aea#861, vendored in `open-aea==2.2.1`). New API takes bytes input, so the call-site ASCII-encodes the CID string. Cross-validated byte-for-byte against the legacy libraries on a real mech CID. Tests updated to patch the new symbol targets.

Three skill yamls (`task_execution`, `task_submission_abci`, `delivery_rate_abci`), the agent `aea-config.yaml`, `pyproject.toml`, and `tox.ini [deps-packages]` no longer declare `py-multibase` / `py-multicodec`, and the two `[mypy-*]` ignore sections are gone. `uv.lock` drops 5 packages total (2 direct + 3 transitives: `morphys`, `python-baseconv`, `varint`).

packages/valory/skills/task_submission_abci/skill.yaml
DhairyaPatel7: Addressed together with the task_execution thread — covered in 4f1e3e33. This skill had no direct `multibase` / `multicodec` imports of its own; it just carried the YAML declarations. Both are dropped.

packages/valory/skills/websocket_client/handlers.py
DhairyaPatel7: Good catch — these were already mis-ordered before this PR; removing the `web3.*` imports between the two stdlib groups exposed the existing drift. Fixed in 4f1e3e33: stdlib (`json`, `time`, `enum`, `typing`) first, third-party (`aea.*`) second, first-party (`packages.*`) third, per isort `profile=black`.

For future-proofing: this file sits under `packages/valory/skills/websocket_client`, which is **not** in the repo's `SERVICE_SPECIFIC_PACKAGES` lint scope (`tox.ini:78` — scope includes `packages/valory/connections/websocket_client` only). That's why neither local `tox -e isort-check` nor CI's `linter_checks` job caught it. Happy to broaden the scope in a follow-up PR — flagging separately rather than quietly expanding here since it would likely surface pre-existing violations in other files under that dir.

DavidMinarsch: all packages that are not third-party should be included into the scope per this pr

packages/valory/agents/mech/aea-config.yaml
```diff
@@ -83,26 +83,14 @@ dependencies:
...
-    version: ==5.2.1
   grpcio:
     version: ==1.78.0
-  hexbytes:
-    version: ==1.3.1
   hypothesis:
     version: ==6.151.9
-  ipfshttpclient:
-    version: ==0.8.0a2
   multidict:
```
DavidMinarsch: is multidict needed?

pyproject.toml
```diff
@@ -13,45 +13,28 @@ dependencies = [
     "open-autonomy==0.21.19",
     "openai==1.30.2",
     "requests==2.32.5",
```
DavidMinarsch: All dependencies here that aren't oa/oaea/plugins should have ranges (consider them "frameworks/libs), rather than exact pins. The ranges should match what we have upstream.

Only agent repos (consider them app) should have exact pins throughout!

packages/valory/skills/websocket_client/handlers.py
DhairyaPatel7: Done in 97d4f2f6. `SERVICE_SPECIFIC_PACKAGES` in `tox.ini` now covers every dev-section package with Python content — added the 5 that were missing:

- `contracts/balance_tracker`
- `contracts/complementary_service_metadata`
- `contracts/hash_checkpoint`
- `skills/delivery_rate_abci`
- `skills/websocket_client`

11 pre-existing violations surfaced by the expanded scope, fixed in the same commit:

- **flake8**: dropped 14 unused imports across `websocket_client/{behaviours,handlers}.py`, `balance_tracker/contract.py`, `delivery_rate_abci/models.py`; renamed `i`/`j` → `_i`/`_j` for unused loop vars.
- **docstrings**: added 4 missing method docstrings on `websocket_client/behaviours.py` subscription-status properties; collapsed a multiline module docstring in `dialogues.py`.
- **mypy**: `_last_subscription_check` annotated `Optional[float]` (was being reassigned from `None` to `float`); three helper methods in `websocket_client/behaviours.py` retyped from `Generator[None, None, X]` to `None` (they had no `yield` statements).
- **darglint**: added `:yield:` lines to two `_get_safe_tx_hash` docstrings (DAR301); documented `**overrides` + 4 kwargs in `task_submission_abci/tests/conftest.py::_make_full_ctx`.

Follow-up 78ad10df re-ran `autonomy packages lock` after a copyright-header sed in 836990a7 which dirtied fingerprints on the 5 `delivery_rate_abci` files (my miss).

packages/valory/agents/mech/aea-config.yaml
DhairyaPatel7: Yes — `multidict` is imported at `packages/valory/connections/http_client/connection.py:37`. That connection is third-party (synced from open-autonomy), so we can't drop the import. Per AEA's agent-as-union-of-component-deps invariant, the agent `aea-config.yaml` must declare every dep that any of its components (connections, skills, contracts, protocols) declares — `connections/http_client/connection.yaml:31` lists `multidict: {}`, which is why the agent yaml restates it.

`tox -e check-dependencies` enforces this union at CI time; dropping it locally would fail that check.

Side effect of 46b4ba7e: the agent yaml entry is now `multidict: {}` (unpinned) instead of `multidict: ==6.7.1` (exact pin), matching OA upstream — consistent with the framework/lib range rule.

Adamantios (dismissed):

tox.ini
```diff
@@ -22,29 +22,27 @@ deps =
...
     open-aea-test-autonomy==0.21.19
+    open-aea-cli-ipfs==2.2.1
+    open-aea-helpers==0.21.19
+    requests>=2.32.5
+    grpcio==1.78.0
+    asn1crypto<1.5.0,>=1.4.0
+    protobuf>=5,<6
+    hypothesis>=6
     web3<8,>=7.0.0
     ipfshttpclient==0.8.0a2
```
DIvyaNautiyal07: `tox.ini` `[deps-packages]` still pins 14 deps that `pyproject.toml` dropped as transitive (`ipfshttpclient`, `eth_typing` L42, `hexbytes` L43, `packaging`, `pytz`, `typing_extensions`, `toml`, `eth-abi`, `eth-utils`, `pycryptodome`, `multiaddr`, `werkzeug`, `attrs`, `py-eth-sig-utils`). If they're transitive-only in `pyproject.toml`, tox shouldn't re-pin them either — the CI test-env ends up with the full old set, which contradicts the PR's "minimise runtime dependency footprint" goal. Suggest trimming `[deps-packages]` to match `pyproject.toml` (keep only what isn't transitive via `open-autonomy[all]==0.21.19`).

packages/valory/agents/mech/aea-config.yaml
```diff
@@ -114,45 +100,29 @@ dependencies:
...
-  packaging:
-    version: ==26
   pebble:
-    version: ==5.1.3
+    version: <6,>=5.1
+  prometheus_client:
+    version: <0.24,>=0.23.1
   protobuf:
-    version: <6,>=5
+    version: <7,>=5
```
DIvyaNautiyal07, line 111: `protobuf` range is inconsistent across scopes: here it's `<7,>=5` (widened per commit message to match the abci connection), but `pyproject.toml` dropped the explicit pin entirely and `tox.ini` still has `protobuf>=5,<6` (L33). Not broken today since the transitive resolves inside `>=5,<6`, but the three scopes should agree — either widen tox to `<7,>=5` or revert this one to `<6,>=5`.

tox.ini
DhairyaPatel7: Addressed in 1851f01c: trimmed [deps-packages] to drop the 12 transitives (ipfshttpclient, eth_typing, hexbytes, packaging, pytz, typing_extensions, toml, eth-abi, eth-utils, pycryptodome, multiaddr, py-eth-sig-utils). Kept werkzeug and attrs because they're declared by the third-party http_server/connection.yaml — aea-helpers check-dependencies requires them. Grouped under an explanatory comment so future readers know why they stay.

packages/valory/agents/mech/aea-config.yaml
DhairyaPatel7: Addressed in 1851f01c: widened tox.ini both sites (L33 [deps-packages] and L245 [testenv:check-hash]) from >=5,<6 to <7,>=5 to match aea-config.yaml. The <7,>=5 range is the correct source of truth — upstream connection/valory/abci declares it too (pulled from OA 0.21.19), which is what the agent actually ships with. pyproject.toml doesn't pin protobuf explicitly (transitive via open-autonomy[all]) so no change needed there. Zero package-hash impact; v0.31.2 tag retagged at new HEAD.

DIvyaNautiyal07 (approved):

Adamantios (approved):

PR #454, opened 2026-06-12
feat(task_execution): skip IPFS upload on offchain, structured 402, Payment-Receipt header

DIvyaNautiyal07:
## Summary

PR 1 of 3 for the mech prepay migration (see `docs/mech_prepay_spec.md`). Two additive changes to the off-chain HTTP path. Contracts unchanged. On-chain (autonomous polling) behaviour is bit-for-bit identical (regression-tested below).

## PR series (how the three connect)

- **PR 1/3 — this PR:** mech server off-chain path. In off-chain mode, skip the IPFS upload of the response (keep it private) and commit a locally-computed **bare-file** CIDv1 instead, plus the structured 402 challenge + `Payment-Receipt` header. The off-chain return channel persists the result under `offchain_request_responses` so `/fetch_offchain_info` can serve it (the response never hits public IPFS). Ships dark behind `use_offchain=false`.
- **PR 2/3 — `valory-xyz/kv-store#16`:** adds `DELETE_REQUEST` / `LIST_REQUEST` performatives to the `valory/kv_store` protocol. Pure enabler — no mech wiring.
- **PR 3/3 — `#455`:** operator-side **preimage retention buffer** — mirrors each off-chain `(request, response)` into the `kv_store` connection keyed by `request_id`, with a background sweeper that LISTs by prefix + DELETEs entries past a retention window (uses PR 2's performatives). Ships dark behind `preimage_retention_enabled=false`. Draft, blocked on PR 2 publishing.

So: PR 1 makes off-chain private, PR 2 gives `kv_store` the delete/list verbs, PR 3 uses them for durable retention.

### Piece 1: skip IPFS upload on the off-chain path, compute CID locally

- New `packages/valory/skills/task_execution/utils/local_cid.py` with `compute_cidv1(content)` that produces the exact CID a real `ipfs add --cid-version=1 --raw-leaves=false` would for the same bytes. Pure-Python UnixFS + DAG-PB encoder, sha256, multibase base32-lower. No new top-level dependencies. Single-block only (raises if content exceeds 1 MiB).
- `_enqueue_offchain_request` no longer pushes to `ipfs_tasks`; it buffers the request JSON in a new `shared_state[in_memory_requests]` dict keyed by `request_id`.
- `_handle_done_task` gates the IPFS store upload on `is_offchain`. When `True` it computes the CID locally and finalizes the done_task synchronously via the new extracted `_finalize_done_task` helper. When `False` (on-chain autonomous-polling tasks) the existing IPFS flow runs unchanged.
- `task_result` (multihash hex passed as on-chain `delivery_data`) is unchanged across both paths.

### Piece 2: structured 402 + Payment-Receipt header

- 402 body now includes `scheme, payTo, asset, chainId, currentBalance, required, depositInstructions, error` alongside the existing `{request_id, status, reason}` keys. Legacy clients reading only `reason` keep working.
- 402 emits `WWW-Authenticate: Payment scheme="olas-prepay" realm="<mech_address>"`.
- 200 emits `Payment-Receipt: <base64 JSON>` carrying `{request_id, accepted_at, accepted_amount, settlement_status: "pending"}`. Header is intentionally a snapshot at HTTP-accept time, NOT a settlement confirmation.
- Adds `payment_type_to_asset_address: Dict[str, str]` skill param. Empty by default; payment types not in the mapping surface the zero address as `asset` so clients can detect native-asset deposits.

## Review updates

- **Off-chain return channel.** The success path now persists the result under `offchain_request_responses[request_id]` so `/fetch_offchain_info` serves the actual `result`/`prompt`/`cost_dict`; previously it fell back to the `done_task`, which carries the CID but not the content (which is no longer on public IPFS).
- **Verifiable commitment.** The committed object is served verbatim under a dedicated `response` field, with the envelope (`request_id`/`status`/`content_cid`) kept outside it, so a consumer re-derives `compute_cidv1(json.dumps(stored["response"])) == content_cid` byte-for-byte (mirrors the on-chain path, which content-addresses `json.dumps(response)`).
- **Fail-honest status.** A task that ran but failed (`_invalid_request`) is now routed through the terminal rejection channel (`status: "rejected"` + reason), not served as `status: "ok"` with an error string — a prepaid requester keys refund/retry/dispute on `status`.

## Tests

- `tests/utils/test_local_cid.py`: 4 fixtures compare the local CID against real `ipfs add` output for known content (empty, small text, small JSON, 100 KiB repeated). Plus oversize rejection + determinism.
- `tests/test_handlers.py`: 5 new tests covering 402 body schema, `WWW-Authenticate` header, `Payment-Receipt` header valid base64-JSON, 200 body shape unchanged for legacy clients, off-chain buffer goes to memory and not to `ipfs_tasks`. Existing balance-scenarios and rollback tests updated to reflect the new flow.
- `tests/test_behaviours.py`: off-chain `_handle_done_task` skips the IPFS store call and finalizes with a local CID (asserting the `content_cid` round-trip); a failed off-chain task is recorded as a rejection, never `status: "ok"`; on-chain path still fires the IPFS upload (regression for the gate).

Local results: full `common_checks.yaml` suite green (lock_check, copyright/deps, third-party + doc hashes, linters, abci specs, handlers, bandit, liccheck, gitleaks); 569 skill tests pass. `safety` reports two pre-existing transitive CVE warnings that also fail on `main`.

## Out of scope

- kv_store-backed preimage buffer + background sweeper. Lands in PR 3 after `kv_store` gains `DELETE_REQUEST` and `LIST_REQUEST` performatives (PR 2 against `valory-xyz/kv-store`).
- mech-side write to wildcard, mech-client, mech-interact, 503 backpressure: separate workstreams.

## Test plan

- [ ] CI green on all `tomte tox` envs that run in `common_checks.yaml`
- [ ] Manual: drive a signed off-chain request against a running mech, confirm the response carries the new `Payment-Receipt` header
- [ ] Manual: drive an insufficient-balance request, confirm 402 body has the structured fields and the `WWW-Authenticate` header is set
- [ ] Manual: confirm the on-chain `delivery_data` written at settlement equals `to_multihash(compute_cidv1(response_json))` for the off-chain path
- [ ] Manual: confirm an on-chain (autonomous-polling) request still uploads to IPFS and behaves exactly as on `main`
- [ ] Manual: poll `/fetch_offchain_info` after an off-chain delivery, confirm `compute_cidv1(json.dumps(body["response"]))` equals the served `content_cid`

### DIFF
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -910,9 +910,31 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             # path's directory-wrapped CID (see utils/local_cid.py). The on-chain
             # commitment derivation (to_multihash, inside _finalize_done_task) is
             # otherwise identical.
+            if self._invalid_request:
+                # The task ran but produced no valid result, so `response`
+                # carries an error string rather than an answer. Route it through
+                # the same terminal-failure channel as the CID / done-task
+                # failures below instead of serving it as a success: a paying
+                # requester keys refund / retry / dispute on `status`, so a "ran
+                # but failed" delivery must be distinguishable from a successful
+                # one — otherwise the client accepts and pays for a failure.
+                self._record_offchain_failure(
+                    cast(str, req_id),
+                    cast(str, response.get("result") or "task execution failed"),
+                )
+                return
             try:
-                response_bytes = json.dumps(response).encode("utf-8")
-                local_cid = compute_cidv1(response_bytes)
+                # content_bytes is exactly what the CID commits to. The serving
+                # payload below carries this same object verbatim under
+                # "response" (envelope fields hang outside it), so a client that
+                # re-derives compute_cidv1(json.dumps(stored["response"]))
+                # reproduces content_cid byte-for-byte and can verify the
+                # delivery against the on-chain commitment. json.dumps round-trips
+                # stably here (default separators, dict insertion order is
+                # preserved through JSON), matching the on-chain path which also
+                # content-addresses json.dumps(response).
+                content_bytes = json.dumps(response).encode("utf-8")
+                local_cid = compute_cidv1(content_bytes)
             except (ValueError, TypeError) as exc:
                 # compute_cidv1 raises ValueError above the single-block bound;
                 # json.dumps raises TypeError (non-serializable) / ValueError
@@ -934,16 +956,18 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             )
             # Off-chain return channel: the response was not uploaded to IPFS, so
             # persist it under offchain_request_responses for /fetch_offchain_info
-            # to serve. Without this the poll falls back to the done_task, which
-            # carries the CID/multihash but not the result/prompt/cost_dict — i.e.
-            # the requester would have no way to read the actual result.
+            # to serve. The committed object is kept verbatim under "response"
+            # (the exact bytes hashed into content_cid); request_id / status /
+            # content_cid are envelope fields kept outside it so they don't
+            # perturb the hash. Without this the poll falls back to the done_task,
+            # which carries the CID/multihash but not the result/prompt/cost_dict.
             self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
                 cast(str, req_id)
             ] = {
-                **response,
                 "request_id": cast(str, req_id),
                 "status": "ok",
                 "content_cid": local_cid,
+                "response": response,
             }
             self._finalize_done_task(local_cid)
             return
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -63,6 +63,7 @@ from packages.valory.skills.task_execution.utils.ipfs import (
     get_ipfs_file_hash,
     to_multihash,
 )
+from packages.valory.skills.task_execution.utils.local_cid import compute_cidv1
 from packages.valory.skills.task_execution.utils.task import AnyToolAsTask
 
 PENDING_TASKS = "pending_tasks"
@@ -897,6 +898,21 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 reason=response["result"],
             )
 
+        if is_offchain:
+            # Off-chain path: skip the IPFS upload so the response stays private,
+            # compute the same CID a real IPFS upload would have produced, and
+            # finalize the done_task synchronously. The on-chain commitment
+            # (``task_result`` as multihash hex, derived inside _finalize_done_task)
+            # is unchanged.
+            response_bytes = json.dumps(response).encode("utf-8")
+            local_cid = compute_cidv1(response_bytes)
+            self.context.logger.info(
+                f"Off-chain response for request {req_id} kept private; "
+                f"local CID {local_cid}."
+            )
+            self._finalize_done_task(local_cid)
+            return
+
         msg, dialogue = self._build_ipfs_store_file_req(
             {str(req_id): json.dumps(response)}
         )
@@ -1269,13 +1285,30 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             )
             return
 
-        executing_task = self._executing_task
-        # if sender is not present, use mech address
-        req_id = executing_task["requestId"]
         ipfs_hash = to_v1(message.ipfs_hash)
+        req_id = self._executing_task["requestId"]
         self.context.logger.info(
             f"Response for request {req_id} stored on IPFS with hash {ipfs_hash}."
         )
+        self._finalize_done_task(ipfs_hash)
+
+    def _finalize_done_task(self, cid: str) -> None:
+        """Apply the CID to the in-flight done_task and publish it.
+
+        :param cid: the CIDv1 string (multibase-encoded) to record on chain via
+            ``task_result``. Origin is either a real IPFS store response (on
+            the on-chain path) or a locally-computed CID (off-chain path).
+        """
+        # Shared finalization for both the IPFS-backed on-chain path (CID from
+        # a real IPFS store response) and the off-chain path (CID computed
+        # locally). Updates pricing, mech address, task_result, appends to
+        # done_tasks under the lock, and resets the executing-task state so the
+        # next task can be picked up. The CID is converted to the multihash hex
+        # shape the contract expects only after the done_task validity check
+        # passes — this matches the previous ordering so callers that pass a
+        # malformed CID alongside an invalid done_task still take the
+        # early-reset branch cleanly.
+        req_id = cast(Dict[str, Any], self._executing_task)["requestId"]
         self.set_last_executed_task(req_id)
         done_task = cast(Dict[str, Any], self._done_task)
         if done_task is None or not isinstance(done_task, Dict):
@@ -1295,7 +1328,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             self._async_result = None
             return
 
-        task_result = to_multihash(ipfs_hash)
+        task_result = to_multihash(cid)
         tool = str(done_task.get("tool"))
         dynamic_tool_cost = self._tools_to_pricing.get(tool)
         if dynamic_tool_cost is not None:
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -910,15 +910,17 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             # path's directory-wrapped CID (see utils/local_cid.py). The on-chain
             # commitment derivation (to_multihash, inside _finalize_done_task) is
             # otherwise identical.
-            response_bytes = json.dumps(response).encode("utf-8")
             try:
+                response_bytes = json.dumps(response).encode("utf-8")
                 local_cid = compute_cidv1(response_bytes)
-            except ValueError as exc:
-                # Response too large for a single block: we cannot produce a CID a
-                # real IPFS upload would match. Fail cleanly so the executing-task
-                # slot resets and the requester gets a definitive rejection,
-                # rather than crashing the agent (propagate policy) or stalling
-                # forever with a non-None _executing_task (just_log policy).
+            except (ValueError, TypeError) as exc:
+                # compute_cidv1 raises ValueError above the single-block bound;
+                # json.dumps raises TypeError (non-serializable) / ValueError
+                # (circular ref). Both must stay inside this guard, or the crash
+                # this fix targets still escapes through act(). Fail cleanly so
+                # the executing-task slot resets and the requester gets a
+                # definitive rejection, rather than crashing the agent
+                # (propagate policy) or stalling forever (just_log policy).
                 self.context.logger.error(
                     f"Off-chain CID computation failed for request {req_id}: {exc}"
                 )
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -932,6 +932,19 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 f"Off-chain response for request {req_id} kept private; "
                 f"local CID {local_cid}."
             )
+            # Off-chain return channel: the response was not uploaded to IPFS, so
+            # persist it under offchain_request_responses for /fetch_offchain_info
+            # to serve. Without this the poll falls back to the done_task, which
+            # carries the CID/multihash but not the result/prompt/cost_dict — i.e.
+            # the requester would have no way to read the actual result.
+            self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
+                cast(str, req_id)
+            ] = {
+                **response,
+                "request_id": cast(str, req_id),
+                "status": "ok",
+                "content_cid": local_cid,
+            }
             self._finalize_done_task(local_cid)
             return
 
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -79,6 +79,11 @@ LAST_SUCCESSFUL_EXECUTED_TASK = "last_successful_executed_task"
 LAST_READ_ATTEMPT_TS = "last_read_attempt_ts"
 INFLIGHT_READ_TS = "inflight_read_ts"
 REQUEST_ID_TO_DELIVERY_RATE_INFO = "request_id_to_delivery_rate_info"
+# Shared-state keys owned by the MechHttpHandler (handlers.py); mirrored here
+# because the off-chain finalize path writes the rejection signal and clears the
+# buffered request metadata from the behaviour side.
+OFFCHAIN_REQUEST_RESPONSES = "offchain_request_responses"
+IN_MEMORY_REQUESTS = "in_memory_requests"
 INITIAL_DEADLINE = 1200.0  # 20mins of deadline
 SUBSEQUENT_DEADLINE = 300.0  # 5min of deadline
 STATUS_CHECK_INTERVAL = 600.0  # 10min interval
@@ -899,13 +904,28 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             )
 
         if is_offchain:
-            # Off-chain path: skip the IPFS upload so the response stays private,
-            # compute the same CID a real IPFS upload would have produced, and
-            # finalize the done_task synchronously. The on-chain commitment
-            # (``task_result`` as multihash hex, derived inside _finalize_done_task)
-            # is unchanged.
+            # Off-chain path: skip the IPFS upload so the response stays private
+            # and commit a locally-computed CIDv1 instead. This is a bare
+            # single-block file CID; it intentionally differs from the on-chain
+            # path's directory-wrapped CID (see utils/local_cid.py). The on-chain
+            # commitment derivation (to_multihash, inside _finalize_done_task) is
+            # otherwise identical.
             response_bytes = json.dumps(response).encode("utf-8")
-            local_cid = compute_cidv1(response_bytes)
+            try:
+                local_cid = compute_cidv1(response_bytes)
+            except ValueError as exc:
+                # Response too large for a single block: we cannot produce a CID a
+                # real IPFS upload would match. Fail cleanly so the executing-task
+                # slot resets and the requester gets a definitive rejection,
+                # rather than crashing the agent (propagate policy) or stalling
+                # forever with a non-None _executing_task (just_log policy).
+                self.context.logger.error(
+                    f"Off-chain CID computation failed for request {req_id}: {exc}"
+                )
+                self._record_offchain_failure(
+                    cast(str, req_id), f"cid computation failed: {exc}"
+                )
+                return
             self.context.logger.info(
                 f"Off-chain response for request {req_id} kept private; "
                 f"local CID {local_cid}."
@@ -1295,37 +1315,39 @@ class TaskExecutionBehaviour(SimpleBehaviour):
     def _finalize_done_task(self, cid: str) -> None:
         """Apply the CID to the in-flight done_task and publish it.
 
-        :param cid: the CIDv1 string (multibase-encoded) to record on chain via
-            ``task_result``. Origin is either a real IPFS store response (on
-            the on-chain path) or a locally-computed CID (off-chain path).
+        Shared by the on-chain path (CID from a real IPFS store response) and the
+        off-chain path (CID computed locally). Updates pricing / mech address /
+        task_result, appends to done_tasks under the lock, and resets the
+        executing-task slot. The CID is converted to the multihash shape only
+        after the done_task validity check passes, so an invalid done_task takes
+        the early-reset branch cleanly.
+
+        :param cid: the multibase-encoded CIDv1 to record on chain via
+            ``task_result``.
         """
-        # Shared finalization for both the IPFS-backed on-chain path (CID from
-        # a real IPFS store response) and the off-chain path (CID computed
-        # locally). Updates pricing, mech address, task_result, appends to
-        # done_tasks under the lock, and resets the executing-task state so the
-        # next task can be picked up. The CID is converted to the multihash hex
-        # shape the contract expects only after the done_task validity check
-        # passes — this matches the previous ordering so callers that pass a
-        # malformed CID alongside an invalid done_task still take the
-        # early-reset branch cleanly.
-        req_id = cast(Dict[str, Any], self._executing_task)["requestId"]
+        executing_task = self._executing_task
+        if executing_task is None:
+            # Two entry points call this now; if the slot was already cleared
+            # there is nothing to finalize. Guard turns a None-deref into a log.
+            self.context.logger.warning(
+                "_finalize_done_task called with no executing task; skipping."
+            )
+            return
+        req_id = cast(Dict[str, Any], executing_task)["requestId"]
+        is_offchain = bool(executing_task.get("is_offchain", False))
         self.set_last_executed_task(req_id)
         done_task = cast(Dict[str, Any], self._done_task)
         if done_task is None or not isinstance(done_task, Dict):
             self.context.logger.error(
                 f"Invalid done task format. Expected Dict. Actual: {done_task}"
             )
-            # Prometheus has no way to remove/clear metrics, so we set to default 0
-            self.mech_metrics.set_gauge(
-                self.mech_metrics.mech_tasks_inflight,
-                0,
-            )
-            self._executing_task = None
-            self._done_task = None
-            self._invalid_request = False
-            self._ipfs_error_reason = None
-            self._request_handling_deadline = None
-            self._async_result = None
+            if is_offchain:
+                # On-chain has other fallbacks; off-chain the polling client would
+                # otherwise never learn this failed, so emit a definitive
+                # rejection (which also resets the task slot).
+                self._record_offchain_failure(req_id, "invalid done task")
+                return
+            self._reset_executing_task()
             return
 
         task_result = to_multihash(cid)
@@ -1353,20 +1375,42 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         with self.done_tasks_lock:
             self.done_tasks.append(done_task)
         self.context.logger.info(f"Done task added to done tasks list: {done_task}.")
+        # Off-chain success: drop the buffered request metadata so it can't grow
+        # unbounded (on-chain tasks never populate it, so this is a no-op there).
+        # .get keeps this safe if the handler-owned key was never initialised.
+        self.context.shared_state.get(IN_MEMORY_REQUESTS, {}).pop(req_id, None)
+        self._reset_executing_task()
 
-        # Prometheus has no way to remove/clear metrics, so we set to default 0
-        self.mech_metrics.set_gauge(
-            self.mech_metrics.mech_tasks_inflight,
-            0,
-        )
-
-        # reset tasks
+    def _reset_executing_task(self) -> None:
+        """Reset the in-flight task slot so the next task can be picked up."""
+        # Prometheus has no clear; set the in-flight gauge back to 0.
+        self.mech_metrics.set_gauge(self.mech_metrics.mech_tasks_inflight, 0)
         self._executing_task = None
         self._done_task = None
         self._invalid_request = False
         self._ipfs_error_reason = None
         self._request_handling_deadline = None
         self._async_result = None
+
+    def _record_offchain_failure(self, req_id: str, reason: str) -> None:
+        """Record an off-chain request failure and reset the task slot.
+
+        Writes the same ``{request_id, status, reason}`` shape the HTTP handler's
+        rejection path uses, so a poll of ``GET /fetch_offchain_info`` returns a
+        definitive rejection instead of an empty "still processing" body forever.
+
+        :param req_id: the off-chain request id that failed.
+        :param reason: a short human-readable failure reason.
+        """
+        # setdefault / get keep this safe if the handler-owned shared-state keys
+        # were never initialised (e.g. on-chain-only deployments, unit tests).
+        self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[req_id] = {
+            "request_id": req_id,
+            "status": "rejected",
+            "reason": reason,
+        }
+        self.context.shared_state.get(IN_MEMORY_REQUESTS, {}).pop(req_id, None)
+        self._reset_executing_task()
 
     def _handle_ipfs_tasks_response(
         self, message: IpfsMessage, dialogue: Dialogue
--- a/packages/valory/skills/task_execution/handlers.py
+++ b/packages/valory/skills/task_execution/handlers.py
@@ -554,12 +554,13 @@ class MechHttpHandler(AbstractResponseHandler):
 
     @property
     def in_memory_requests(self) -> Dict[str, str]:
-        """Get in-memory off-chain request metadata buffered by request id.
+        """Get in-memory off-chain request payloads buffered by request id.
 
-        :return: a dict keyed by request_id with the raw request JSON string.
+        :return: a dict keyed by request_id whose values are the request's
+            ``ipfs_data`` payload (not the full request envelope). Populated when
+            the off-chain path skips the IPFS upload; cleared when the task
+            finalizes.
         """
-        # Populated when the off-chain path skips the IPFS upload and keeps the
-        # request JSON locally instead.
         return self.context.shared_state[IN_MEMORY_REQUESTS]
 
     @property
@@ -596,6 +597,28 @@ class MechHttpHandler(AbstractResponseHandler):
         :param http_msg: the HttpMessage instance
         :param http_dialogue: the HttpDialogue instance
         """
+        # Phase 1 ships dark: the off-chain HTTP path is disabled by default and
+        # enabled per deployment in the Phase 2 rollout (use_offchain). When off,
+        # nothing about the on-chain + IPFS flow changes; off-chain requests are
+        # refused so none of the new off-chain code path runs.
+        if not self.params.use_offchain:
+            self.context.logger.info(
+                "Off-chain request received but the off-chain path is disabled "
+                "(use_offchain=false); refusing."
+            )
+            http_response = http_dialogue.reply(
+                performative=HttpMessage.Performative.RESPONSE,
+                target_message=http_msg,
+                version=http_msg.version,
+                status_code=HttpCode.SERVICE_UNAVAILABLE_CODE.value,
+                status_text="Service unavailable",
+                headers=f"{self.json_content_header}{http_msg.headers}",
+                body=json.dumps({"error": "offchain path disabled"}).encode(
+                    ENCODING_UTF8
+                ),
+            )
+            self.context.outbox.put_message(message=http_response)
+            return
 
         try:
             data = self._parse_http_body(http_msg)
@@ -778,6 +801,10 @@ class MechHttpHandler(AbstractResponseHandler):
             terminated by ``\n``) prepended to the response. Callers use this to
             add audit headers (e.g. ``Payment-Receipt``) without rewriting body.
         """
+        # Each header line must be newline-terminated, else it would silently
+        # merge with the Content-Type line that follows.
+        if extra_headers and not extra_headers.endswith("\n"):
+            raise ValueError("extra_headers must be empty or newline-terminated")
         http_response = http_dialogue.reply(
             performative=HttpMessage.Performative.RESPONSE,
             target_message=http_msg,
@@ -814,6 +841,10 @@ class MechHttpHandler(AbstractResponseHandler):
         :param extra_headers: optional pre-formatted header block to prepend.
         :param body_extras: optional dict merged into the JSON response body.
         """
+        # Each header line must be newline-terminated, else it would silently
+        # merge with the Content-Type line that follows.
+        if extra_headers and not extra_headers.endswith("\n"):
+            raise ValueError("extra_headers must be empty or newline-terminated")
         # ``body_extras`` is merged into the JSON body alongside the canonical
         # ``{request_id, status, reason}`` keys, so legacy clients that only
         # read ``reason`` keep working while new clients can pick up structured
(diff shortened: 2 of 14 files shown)


OjusWiZard (commented):
Multi-perspective review (correctness, silent failures, tests, types, comments). Findings posted inline: 1 Critical, 5 Important, plus a few advisories. Two advisory items without a good inline anchor:

- **[Advisory]** The `extra_headers: str` contract ("each header terminated by `\n`") is enforced only by convention at the two composition points in `_send_ok_response`/`_send_rejection_response`; a cheap assertion (`extra_headers == "" or extra_headers.endswith("\n")`) would turn a silently merged header line into a loud failure.
- **[Advisory]** Several new inline comment blocks restate their docstrings (`behaviours.py` `_finalize_done_task` body comment, `handlers.py` `_send_rejection_response` / `_build_402_challenge` / `_build_payment_receipt_header` body comments) and use PR-review phrasing ("is unchanged") that loses meaning post-merge — worth condensing.

No security vulnerabilities found. The extracted `_finalize_done_task`, the golden-CID fixtures, and the on-chain regression gate tests are all solid.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -897,6 +898,21 @@ def _handle_done_task(self, task_result: Any) -> None:
...
             )
 
+        if is_offchain:
+            # Off-chain path: skip the IPFS upload so the response stays private,
+            # compute the same CID a real IPFS upload would have produced, and
+            # finalize the done_task synchronously. The on-chain commitment
+            # (``task_result`` as multihash hex, derived inside _finalize_done_task)
+            # is unchanged.
+            response_bytes = json.dumps(response).encode("utf-8")
+            local_cid = compute_cidv1(response_bytes)
```
OjusWiZard: **[Critical] Unhandled `ValueError` here stalls the agent and leaves the off-chain requester polling forever.**

`compute_cidv1` raises `ValueError` for content over the block bound, and nothing catches it at any depth (`_handle_done_task` ← `_execute_task` / `_handle_timeout_task` ← `act()`). Under the default `skill_exception_policy: propagate` the agent crashes; under `just_log`, `_executing_task` stays non-`None` forever so no new task is ever picked up. Either way nothing is written to `offchain_request_responses` or `done_tasks`, so the client polls `{}` indefinitely. A tool response whose JSON exceeds the bound (base64 image, verbose output) triggers this in production.

Suggested fix: wrap in try/except `ValueError`; on failure log at ERROR, set `_invalid_request` / `_ipfs_error_reason`, and route through the existing failure finalization so state resets and the requester gets a definitive failure. A test asserting `_executing_task is None` after an oversized response would lock this in.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -897,6 +898,21 @@ def _handle_done_task(self, task_result: Any) -> None:
                 reason=response["result"],
             )
 
+        if is_offchain:
+            # Off-chain path: skip the IPFS upload so the response stays private,
```
OjusWiZard: **[Important] This is not "the same CID a real IPFS upload would have produced" — the mech's own on-chain pipeline wraps in a directory.**

The on-chain path stores `{str(req_id): json.dumps(response)}` and the IPFS connection calls `IPFSTool.add(path)` with the default `wrap_with_directory=True` (`connections/ipfs/connection.py:203`), so it commits the **wrapping directory's** CID, which also covers the filename (`req_id`). `compute_cidv1` produces a bare UnixFS file CID over the raw JSON bytes. Identical response content therefore commits structurally different `delivery_data` on the two paths, and tooling that resolves `ipfs://<cid>/<req_id>` (the on-chain shape) cannot do so for off-chain commitments.

This may be intentional (the content is private by design), but it contradicts this comment and the `local_cid.py` docstring's parity claim, and it likely matters for PR 3 / mech-client verification. Please either match the directory-wrapped shape or document the divergence as a deliberate decision.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1269,13 +1285,30 @@ def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> No
...
+        # Shared finalization for both the IPFS-backed on-chain path (CID from
+        # a real IPFS store response) and the off-chain path (CID computed
+        # locally). Updates pricing, mech address, task_result, appends to
+        # done_tasks under the lock, and resets the executing-task state so the
+        # next task can be picked up. The CID is converted to the multihash hex
+        # shape the contract expects only after the done_task validity check
+        # passes — this matches the previous ordering so callers that pass a
+        # malformed CID alongside an invalid done_task still take the
+        # early-reset branch cleanly.
+        req_id = cast(Dict[str, Any], self._executing_task)["requestId"]
```
OjusWiZard: **[Important] Off-chain finalize failure leaves the requester with no failure signal** — and the `cast` hides a possible `None` deref.

When `done_task` is `None`/malformed, the early-reset branch below logs and resets state but writes nothing to `offchain_request_responses` or `done_tasks`. On-chain there are other fallbacks; off-chain the polling client can never distinguish "still processing" from "failed". Suggest writing a rejected-status entry to `offchain_request_responses[req_id]` in that branch.

*[Advisory]* Relatedly: `_handle_store_response` guards `if not self._executing_task` before calling this, but the new off-chain call site relies on the caller too. Now that `_finalize_done_task` has two entry points, a two-line guard here (mirroring `_handle_store_response`) would turn a `TypeError: 'NoneType' object is not subscriptable` into an observable, recoverable error.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -826,12 +970,7 @@ def _enqueue_offchain_request(
         }
         try:
             self.pending_tasks.append(req)
-            self.ipfs_tasks.append(
-                {
-                    RequestKey.REQUEST_ID.value: request_id,
-                    RequestKey.IPFS_DATA.value: data[RequestKey.IPFS_DATA.value],
-                }
-            )
+            self.in_memory_requests[request_id] = data[RequestKey.IPFS_DATA.value]
```
OjusWiZard, line 1059: **[Important] `in_memory_requests` grows without bound — entries are never removed on success, and nothing reads them.**

The only removal site is the rollback path (`_rollback_offchain_enqueue`); there are zero consumers of `IN_MEMORY_REQUESTS` anywhere in the skills. Every accepted off-chain request leaves its `ipfs_data` payload (up to ~1 MiB) in memory for the life of the process. I understand the kv_store buffer + sweeper is deferred to PR 3, but this PR still needs a success-path cleanup — e.g. pop the entry when the off-chain task finalizes. A test asserting the entry is absent after a successful off-chain delivery would lock the lifecycle in.

*Minor:* the `in_memory_requests` property docstring says "raw request JSON string", but only the `ipfs_data` field is stored here, not the full request envelope.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -760,41 +800,133 @@ def _send_rejection_response(
...
             status_code=status_code,
             status_text=status_text,
-            headers=f"{self.json_content_header}{http_msg.headers}",
+            headers=f"{extra_headers}{self.json_content_header}{http_msg.headers}",
             body=json.dumps(response_payload).encode(ENCODING_UTF8),
         )
         self.context.logger.info("Responding with: {}".format(http_response))
         self.context.outbox.put_message(message=http_response)
 
+    def _build_402_challenge(
```
OjusWiZard, line 917: **[Advisory]** This method is safe at its current call site (gated on `status == OK`), but if it's ever fed an error-shaped dict (e.g. from `_make_unavailable_balance_response`, which lacks the three new keys), the `.get(..., default)` fallbacks silently emit `payTo = 0x000…0`, `chainId: 0`, `currentBalance: "0"` — syntactically valid payment instructions pointing at the zero address. A guard that raises when `balance_tracker_address` is absent (or a `TypedDict` return for `_check_offchain_requester_balance`) would make future misuse loud instead of silent.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -760,41 +800,133 @@ def _send_rejection_response(
...
+        # balance-check result and pairs them with deposit instructions the
+        # client can use to construct an on-chain top-up. Native-asset payment
+        # models surface the zero address for ``asset``; clients that see the
+        # zero address must skip the ERC20 approve step.
+        balance_tracker_address = cast(
+            str,
+            balance_check.get(ResponseKey.BALANCE_TRACKER_ADDRESS.value, ZERO_ADDRESS),
+        )
+        payment_type = cast(str, balance_check.get(ResponseKey.PAYMENT_TYPE.value, ""))
+        asset_address = self.params.payment_type_to_asset_address.get(
```
OjusWiZard, line 948: **[Advisory]** No key-case normalization on this lookup: `payment_type` comes from the contract side while the map keys come from operator YAML. A checksummed-vs-lowercase hex mismatch silently falls back to `ZERO_ADDRESS`, signalling a native-asset deposit for what is actually an ERC-20 payment model. A `.lower()` normalization of both the map keys (at load time in `models.py`) and the lookup key here closes it. The test fixtures use all-lowercase keys, so this isn't currently exercised.

packages/valory/skills/task_execution/utils/local_cid.py
```diff
@@ -0,0 +1,162 @@
...
+Single-block only: content must fit in one IPFS block. IPFS chunks at 256 KiB by
+default; payloads above that bound need a chunked / linked DAG and are not handled
+here. Hash sanity asserts the bound at runtime so a future oversized response fails
+loudly rather than silently producing a CID a real IPFS upload would not match.
+"""
+
+import base64
+import hashlib
+
+_MAX_BLOCK_BYTES = 1024 * 1024  # IPFS default chunker is 256 KiB; one MiB is plenty.
```
OjusWiZard: **[Important] The 1 MiB guard doesn't match the 256 KiB single-block reality — content in (256 KiB, 1 MiB] silently gets a CID real IPFS would never produce.**

The module docstring correctly says IPFS chunks at 256 KiB, but the guard only rejects above 1 MiB. For content between those bounds, `compute_cidv1` returns a single-block CID while a real `ipfs add --cid-version=1 --raw-leaves=false` would build a multi-block DAG with a different root — exactly the silent divergence this guard exists to prevent. The largest test fixture is 100 KB, so the boundary is untested.

Suggested fix: set `_MAX_BLOCK_BYTES = 256 * 1024`, add a golden-CID fixture at exactly 256 KiB and a rejection test just above it.

*Also (docstring accuracy, line 27):* "Hash sanity **asserts** the bound" — the code (correctly) raises `ValueError`, not `assert`; worth saying so, since `assert` implies the check is strippable under `python -O`.

DhairyaPatel7:
Pushed a commit addressing the review + gating the offchain path behind a default-off flag.

**Correctness**
- **[Critical] oversized response** — `compute_cidv1` is now wrapped; on `ValueError` it writes a `rejected` entry to `offchain_request_responses` and resets the executing-task slot, so an over-the-block-bound response no longer crashes (propagate) or stalls (just_log) the agent. Test added (`test_handle_done_task_offchain_oversized_records_failure_and_resets`).
- **[Important] finalize-failure signal + None deref** — `_finalize_done_task` now guards a `None` executing task and, on an invalid `done_task` for an offchain request, writes a definitive rejection so the poller can distinguish failed from still-processing.
- **[Important] `in_memory_requests` unbounded** — popped on success (and failure); property docstring corrected (it holds `ipfs_data`, not the full envelope). Test added.
- **[Important] 256 KiB bound** — `_MAX_BLOCK_BYTES` tightened to `256 * 1024`; boundary acceptance test added; docstring fixed (`raises ValueError`, not `assert`).
- **[Advisory] 402 zero-address** — `_build_402_challenge` now raises rather than emitting zero-address deposit instructions when the balance check lacks a balance-tracker address; `payment_type` is case-normalised (keys lower-cased at load in `models.Params`).
- **[Advisory] `extra_headers` contract** — both senders now reject a non-newline-terminated header block.

**[Important] CID parity** — kept as a **bare single-block file CID** and documented the deliberate divergence from the on-chain directory-wrapped CID (offchain content is never published, so verification re-hashes the raw bytes). Flagging for the mech-client / PR3 verification side to match this contract — happy to switch to directory-wrapped instead if you'd prefer parity.

**Feature flag (Phase 1 ships dark)** — added `use_offchain` (default **false**). When off, the offchain HTTP endpoint returns 503 and the on-chain + IPFS path is untouched; Phase 2 enables it per deployment. 503-gate test added.

Comment blocks condensed per the advisory. Package hashes relocked. Local: black/isort/flake8/mypy/pylint/darglint, check-hash/check-packages/check-handlers/check-abci-docstrings/check-abciapp-specs, copyright, and the task_execution + dependent-skill test suites all pass.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -897,6 +903,36 @@ def _handle_done_task(self, task_result: Any) -> None:
...
             )
 
+        if is_offchain:
+            # Off-chain path: skip the IPFS upload so the response stays private
+            # and commit a locally-computed CIDv1 instead. This is a bare
+            # single-block file CID; it intentionally differs from the on-chain
+            # path's directory-wrapped CID (see utils/local_cid.py). The on-chain
+            # commitment derivation (to_multihash, inside _finalize_done_task) is
+            # otherwise identical.
+            response_bytes = json.dumps(response).encode("utf-8")
```
OjusWiZard: **[Important] The `ValueError` catch below doesn't cover this line — the crash the fix targets can still happen.**

`json.dumps(response)` sits one line *above* the `try`, but it can raise `TypeError` (a non-serializable value in `response`) or `ValueError` (circular reference). Either escapes uncaught straight through `act()`, reproducing exactly the agent-crash / stuck-slot scenario this commit set out to prevent — the `except ValueError` only wraps `compute_cidv1`.

Bounded by two mitigants: the on-chain path serializes the same `response` object (`_build_ipfs_store_file_req({str(req_id): json.dumps(response)})`), so responses are de-facto serializable in practice, and the path ships dark behind `use_offchain`. So Important, not Critical.

Fix: move this line inside the `try` and broaden to `except (ValueError, TypeError)` so the "fail cleanly" guarantee actually holds.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1706,6 +1810,54 @@ def test_handle_store_response_invalid_done_task_resets_state(
...
+    monkeypatch.setattr(behaviour.mech_metrics, "set_gauge", MagicMock())
+    monkeypatch.setattr(behaviour.mech_metrics, "inc_counter", MagicMock())
+    monkeypatch.setattr(behaviour.mech_metrics, "observe_histogram", MagicMock())
+
+    behaviour._handle_done_task(task_result=None)
+
+    assert send_calls == []
+    assert shared_state[beh_mod.DONE_TASKS] == []
+    assert behaviour._executing_task is None
+    rejection = shared_state[beh_mod.OFFCHAIN_REQUEST_RESPONSES][13]
```
OjusWiZard: **[Important] This test (and the cleanup test below) use integer keys, but production keys these dicts by the string `request_id` — so they pass for the wrong reason.**

The handler always stores `offchain_request_responses` / `in_memory_requests` under the **string** `request_id` parsed from the HTTP body (the bytes→int conversion at `behaviours.py:665` fires only for `bytes`, never for an off-chain str id). Seeding `req_id=13` (int) and looking up `[13]` is internally self-consistent and green, but it never exercises the real str→str path, so a future str/int regression in the cleanup/rejection write would slip through.

No production bug today — the live types are consistently str. This is a coverage gap. Fix: use a string id (e.g. `req_id="req-13"`) here and in the cleanup test (`{"req-14": ...}` / `_seed_executing_task(..., req_id="req-14")`).

packages/valory/skills/task_execution/handlers.py
```diff
@@ -636,6 +682,10 @@ def _handle_signed_requests(
                 reason="insufficient balance",
                 status_code=HttpCode.PAYMENT_REQUIRED_CODE.value,
                 status_text="Payment required",
+                extra_headers=self._build_www_authenticate_header(),
+                body_extras=self._build_402_challenge(
```
OjusWiZard: **[Advisory]** `_build_402_challenge` now raises `ValueError` (when `balance_tracker_address` is missing), and `_send_ok_response`/`_send_rejection_response` raise on non-newline `extra_headers`. All three are invoked here inside `_handle_signed_requests` with no surrounding try/except, and the `mech_abci` handler dispatch doesn't catch them either. Unreachable today (every OK balance check includes the address; both header builders emit a trailing `\n`), but since these guards exist precisely as future safety nets, if one ever fires the `ValueError` escapes with *no HTTP reply* and the client hangs. Converting them to a 503/500 response would be more robust than raising on a request-handling path.

OjusWiZard (commented):
Re-review of the remediation commit (6367d88b). The prior 7 findings are all genuinely addressed — thanks. These are **new** observations on the fix itself.

Good: `_reset_executing_task` is a faithful extraction (all 6 fields + gauge), the 503 gate short-circuits before any state mutation, conftest sets `use_offchain=True` so the existing off-chain tests still hit the real path (not trivially bypassed via 503), and the 256 KiB boundary fix is correct and tested.

Two items worth fixing (one runtime, one test), plus advisories inline. One advisory without a clean anchor:

- **[Advisory]** No guard tests for the newly-added defensive paths: the `_build_402_challenge` missing-address raise, the `extra_headers` newline raise, end-to-end `payment_type` case-insensitivity, and the `_finalize_done_task` None-guard. Reverting any of those guards would not fail a test today.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -760,41 +827,147 @@ def _send_rejection_response(
...
+                "balance_tracker_address; refusing to emit zero-address deposit "
+                "instructions"
+            )
+        balance_tracker_address = cast(
+            str, balance_check[ResponseKey.BALANCE_TRACKER_ADDRESS.value]
+        )
+        # Normalize the lookup key — the map keys are lower-cased at load time
+        # (models.Params) so a checksummed payment_type still resolves.
+        payment_type = cast(
+            str, balance_check.get(ResponseKey.PAYMENT_TYPE.value, "")
```
OjusWiZard: **[Advisory]** `cast(str, ...)` is a runtime no-op, so if `PAYMENT_TYPE` is present with a `None` value, `.get(..., "")` returns `None` and the `.lower()` on the next line raises `AttributeError`. Unreachable today (a `None` payment_type takes the early unavailable return in `_check_offchain_requester_balance`), but cheap to harden: `(balance_check.get(ResponseKey.PAYMENT_TYPE.value) or "").lower()`.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -573,6 +597,28 @@ def _handle_signed_requests(
         :param http_msg: the HttpMessage instance
         :param http_dialogue: the HttpDialogue instance
         """
+        # Phase 1 ships dark: the off-chain HTTP path is disabled by default and
+        # enabled per deployment in the Phase 2 rollout (use_offchain). When off,
+        # nothing about the on-chain + IPFS flow changes; off-chain requests are
+        # refused so none of the new off-chain code path runs.
+        if not self.params.use_offchain:
```
OjusWiZard, line 604: **[Advisory]** The 503 gate returns before parsing the body, so it stores nothing in `offchain_request_responses`. A client that then polls `GET /fetch_offchain_info` with that `request_id` gets `{}` indefinitely rather than a definitive rejection. Acceptable as a Phase-1 protocol constraint (the client should treat the 503 as terminal, and `request_id` isn't available pre-parse anyway) — just flagging the "empty poll forever" semantic so it's a deliberate choice.

DhairyaPatel7:
Thanks for the re-review — pushed a follow-up (40452dd) addressing all of it.

**[Important]**
- `json.dumps(response)` moved inside the off-chain `try`, broadened to `except (ValueError, TypeError)`. The serialize step sat above the try, so a non-serializable / circular response could still escape through `act()` — exactly the stall the guard targets. Now caught.
- The new tests used **int** request_ids, but production keys `offchain_request_responses` / `in_memory_requests` by the **str** `request_id` — so they passed for the wrong reason. Switched the oversized + cleanup tests to string ids (`"req-13"` / `"req-14"`) and widened the `_seed_executing_task` hint.

**[Advisory]**
- The 402 / header-builder guards no longer escape with no HTTP reply: `_handle_signed_requests` now catches them and replies **500** via a new `_send_internal_error` helper (passing no extra_headers so the 500 itself can't re-trip the newline guard).
- `payment_type` lookup hardened against a present-but-None value: `(... or "").lower()`.
- Added the missing **guard tests** — the 402 missing-address raise, `payment_type` case-insensitivity + None-safety, the `extra_headers` newline guard on both senders, the handler 500-on-guard-fire path, and the `_finalize_done_task` None-guard. Reverting any guard now fails a test.
- 503 empty-poll semantics (`request_id` unavailable pre-parse): left as the deliberate Phase-1 choice you flagged — the client treats the 503 as terminal.

Local: black/isort/flake8/mypy/pylint/darglint, check-hash/check-packages/check-handlers/check-abci-docstrings/check-abciapp-specs, copyright, and the task_execution + dependent-skill suites all pass (285 / 283). Hashes relocked.

OjusWiZard (dismissed):
Approving. ✅

Reviewed across three rounds (initial + two remediations). The off-chain skip-IPFS / structured-402 / Payment-Receipt work is solid:

- All prior findings addressed and verified against the code, not just the commit messages — the `compute_cidv1` crash guard, the `_finalize_done_task` None-guard + off-chain rejection signal, `in_memory_requests` cleanup, the 256 KiB single-block bound, the structured-402 raise-instead-of-zero-address guard, payment_type case/None safety, and the `extra_headers` newline guard with handler-level 500 fallback.
- On-chain (autonomous-polling) path is unchanged; the whole off-chain path ships dark behind `use_offchain=false`, so Phase 1 is low-risk.
- Test coverage is real — the guard tests fail if their guards are reverted, and the int→str shared-state key fix means the cleanup tests now exercise the production contract.

One remaining **near-theoretical advisory** (non-blocking): on the post-enqueue OK path, `_send_internal_error` persists a rejection to `offchain_request_responses`, which `_handle_offchain_request_info` serves with precedence over `done_tasks` — so *if* the receipt-header build ever raised, a client could poll "rejected" while the task actually delivered on-chain. It's effectively unreachable (str request_id + int amount through json/base64/datetime cannot raise), so it's fine to land as-is; worth a follow-up only if that branch ever gains a real trigger.

Note: I could not run the full pytest suite locally (the unrelated `mech_marketplace` contract isn't synced from IPFS in this environment). Verified the changed files compile and validated `local_cid` + the new logic directly. Please confirm CI is green before merge.

bennyjo:
> PR 1 of 3 for the mech prepay migration (see docs/mech_prepay_spec.md in valory-xyz/autonolas-marketplace).

I [can't find ](https://github.com/valory-xyz/autonolas-marketplace/tree/main/docs)this spec. Where is it?

Since this is PR 1 in a series of 3. I'd like to understand all three and how they connect.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1320,21 +1377,43 @@ def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> No
...
-        self.mech_metrics.set_gauge(
-            self.mech_metrics.mech_tasks_inflight,
-            0,
-        )
-
-        # reset tasks
+        # Off-chain success: drop the buffered request metadata so it can't grow
+        # unbounded (on-chain tasks never populate it, so this is a no-op there).
+        # .get keeps this safe if the handler-owned key was never initialised.
+        self.context.shared_state.get(IN_MEMORY_REQUESTS, {}).pop(req_id, None)
```
bennyjo, line 1420: The prepay spec (mech-side changes) says: "Skip IPFS upload of response in offchain mode; **keep in `offchain_request_responses` dict**." This success path skips the IPFS upload but doesn't populate `offchain_request_responses` — that dict only gets written on the rejection path now. Since `/fetch_offchain_info` returns `offchain_request_responses[request_id]` if present and otherwise falls back to the `done_task` (which carries the CID/multihash + metadata but not `result`/`prompt`/`cost_dict`), an accepted offchain request ends up with the result reachable nowhere: the CID no longer resolves on public IPFS, and the content isn't in `offchain_request_responses`.

Is response retention into `offchain_request_responses` deferred to the kv_store PR (PR 3), or was this spec line missed here? As written, the offchain happy path has no return channel for the actual result. Happy to defer if it's tracked — just want to confirm it isn't an accidental drop in the `_finalize_done_task` extraction.

packages/valory/skills/task_execution/utils/local_cid.py
```diff
@@ -0,0 +1,173 @@
...
+Single-block only: content must fit in one IPFS block. IPFS chunks at 256 KiB by
+default, so the single-block bound is exactly 256 KiB; payloads above it need a
+chunked / linked DAG and are not handled here. ``compute_cidv1`` raises
+``ValueError`` (not ``assert``, which ``python -O`` would strip) above the bound,
+so a future oversized response fails loudly rather than silently producing a CID
+a real IPFS upload would not match.
+
+Shape note: this is a *bare* single-block file CID, matching ``ipfs add`` without
+``-w``. The mech's on-chain delivery path instead uploads via the IPFS connection
+with ``wrap_with_directory=True``, committing the *directory* CID. The two paths
```
bennyjo, line 34: Per the prepay spec's verification model ("the same content CID IPFS would have produced… anyone with the original content can re-hash and verify"), committing a **bare-file** CID here is the right call — re-hashing raw content reproduces it. But it's worth pinning the consequence explicitly, because the on-chain/IPFS path commits a **directory-wrapped** CID (`{req_id: response}` is stored as a named file → `message.ipfs_hash` is a directory hash, resolved as `ipfs://<cid>/<req_id>`). So for identical content, `source=mech_offchain` rows and `source=ipfs_historical` rows carry structurally different CIDs, and the spec's `content_cid` ("same as contract sees") treats it as singular.

Is the plan that every verifier — mech-client, mech-interact, and the historical ETL that validates offchain `content_cid` against subgraph `ParsedRequest` — branches on path and uses the bare-file derivation for offchain? A verifier that re-derives the directory form uniformly will reject every offchain commitment once `use_offchain` flips on. Not blocking for this PR (ships dark) — flagging so it's nailed down in the spec / consumer PRs before Phase 2.

bennyjo (commented):
Two questions framed against `docs/mech_prepay_spec.md` — the offchain result-return channel and the bare-file vs directory-wrapped CID contract. Neither blocks this PR (ships dark behind `use_offchain=false`); both want pinning before Phase 2.

packages/valory/skills/task_execution/behaviours.py
DhairyaPatel7: Good catch — this is a real drop, not deferred to PR 3.

PR 3 (mech#455) writes the preimage to the `kv_store` connection, but that's the **operator-side** durable audit buffer (retention / commit-reveal proof) — a different channel from the **requester-facing** `/fetch_offchain_info`, which reads the in-process `offchain_request_responses` dict. So PR 3 doesn't close this; as written, the offchain happy path has no return channel.

The fix belongs here in #454: the offchain success branch (where it computes `response` / `local_cid` before `_finalize_done_task`) should write the result into `offchain_request_responses[request_id]`, mirroring the rejection path at `handlers.py:945` — e.g. `{request_id, status: "ok", result, prompt, cost_dict, metadata, content_cid: local_cid}` — so `/fetch_offchain_info` returns the actual content instead of falling back to the `done_task` (which carries the CID/multihash but not `result`/`prompt`/`cost_dict`). It's behind `use_offchain=false`, so low risk.

Want me to push that to this branch?

packages/valory/skills/task_execution/utils/local_cid.py
DhairyaPatel7: Agreed on both points — the bare-file CID is the intentional, correct choice (re-hash-to-verify), and yes, every verifier has to branch on path.

Confirming it's tracked rather than accidental: on the data-lake side the divergence is already disambiguated by the `source` column — `source='mech_offchain'` rows carry the bare-file CID and `source='ipfs_historical'` rows the directory-wrapped one — so the historical-ETL validation (predict-api#160's `mech_requests.content_cid`) branches on `source` rather than re-deriving uniformly. The mech-client / mech-interact verifiers (PR 2 + consumer PRs) need the same path-branch before `use_offchain` flips on; flagging it there so it's nailed down in the spec before Phase 2. Nothing to change in this PR.

DhairyaPatel7:
@bennyjo on the series (I own PR 3) — here's how the three connect:

- **PR 1/3 — this PR (#454):** mech server offchain path. In offchain mode, skip the IPFS upload of request + response (keep them private) and commit a locally-computed bare-file CID instead; plus the structured 402 challenge + `Payment-Receipt` header. Ships dark behind `use_offchain=false`.
- **PR 2/3 — valory/kv-store#16:** adds `DELETE_REQUEST` / `LIST_REQUEST` performatives to the `valory/kv_store` protocol. Pure enabler — no mech wiring.
- **PR 3/3 — mech#455** (just opened): the operator-side **preimage retention buffer** — mirrors each offchain `(request, response)` into the `kv_store` connection keyed by `request_id`, with a background sweeper that LISTs by prefix + DELETEs entries past a retention window (uses PR 2's performatives). Ships dark behind `preimage_retention_enabled=false`. It's a draft, blocked on PR 2 publishing.

So: PR 1 makes offchain private, PR 2 gives `kv_store` the delete/list verbs, PR 3 uses them for durable retention.

On the spec link — `mech_prepay_spec.md` isn't under `autonolas-marketplace/docs` (the path in the PR description looks stale); @DIvyaNautiyal07 can point you to the current canonical location. Happy to walk through any of the three.

packages/valory/skills/task_execution/behaviours.py
DhairyaPatel7: Done — pushed in 2f0ab30a. The off-chain success path now writes the response into `offchain_request_responses[request_id]` (mirroring the rejection path's shape, with `status: "ok"` + `content_cid`), so `/fetch_offchain_info` serves the actual result instead of falling back to the `done_task` (which has the CID but not `result`/`prompt`/`cost_dict`). Behind `use_offchain=false`, so on-chain behaviour is unchanged. Extended the off-chain finalize test to assert the write.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -897,6 +903,51 @@ def _handle_done_task(self, task_result: Any) -> None:
...
+            )
+            # Off-chain return channel: the response was not uploaded to IPFS, so
+            # persist it under offchain_request_responses for /fetch_offchain_info
+            # to serve. Without this the poll falls back to the done_task, which
+            # carries the CID/multihash but not the result/prompt/cost_dict — i.e.
+            # the requester would have no way to read the actual result.
+            self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
+                cast(str, req_id)
+            ] = {
+                **response,
```
bennyjo: Follow-up to 2f0ab30a — two contracts to pin **before `use_offchain` flips on** (not blocking the dark merge).

**1. Committed bytes ≠ served bytes, so offchain deliveries are unverifiable.** Line 914 commits `compute_cidv1(json.dumps(response))` (bare response); this branch serves the envelope `{**response, "request_id", "status", "content_cid"}`, so `compute_cidv1(served) != content_cid`. **Impact:** every verifier that re-hashes the served payload (mech-client, mech-interact, historical ETL) rejects every offchain commitment the moment `use_offchain` is on — i.e. the feature is dead on arrival in Phase 2. Could we serve the exact committed bytes (keep `json.dumps(response)` verbatim, hang `status`/`content_cid` outside it) plus a round-trip assertion?

**2. A failed task is served as `status: "ok"`.** This runs after the `_invalid_request` block at line 899, so tool failures land in the `"ok"` bucket while CID/done_task failures get `"rejected"`. **Impact:** a paying client keys refund/retry on `status` and can't tell "ran but failed" from "succeeded" — so it pays for and accepts failed deliveries instead of retrying or disputing them. Deliberate, or should `_invalid_request` get its own status?

bennyjo (commented):
Two contracts to pin before `use_offchain` flips on (non-blocking for the dark merge) — details inline.

bennyjo:
> @bennyjo on the series (I own PR 3) — here's how the three connect:
> 
> * **PR 1/3 — this PR ([feat(task_execution): skip IPFS upload on offchain, structured 402, Payment-Receipt header #454](https://github.com/valory-xyz/mech/pull/454)):** mech server offchain path. In offchain mode, skip the IPFS upload of request + response (keep them private) and commit a locally-computed bare-file CID instead; plus the structured 402 challenge + `Payment-Receipt` header. Ships dark behind `use_offchain=false`.
> * **PR 2/3 — valory/kv-store#16:** adds `DELETE_REQUEST` / `LIST_REQUEST` performatives to the `valory/kv_store` protocol. Pure enabler — no mech wiring.
> * **PR 3/3 — mech#455** (just opened): the operator-side **preimage retention buffer** — mirrors each offchain `(request, response)` into the `kv_store` connection keyed by `request_id`, with a background sweeper that LISTs by prefix + DELETEs entries past a retention window (uses PR 2's performatives). Ships dark behind `preimage_retention_enabled=false`. It's a draft, blocked on PR 2 publishing.
> 
> So: PR 1 makes offchain private, PR 2 gives `kv_store` the delete/list verbs, PR 3 uses them for durable retention.
> 
> On the spec link — `mech_prepay_spec.md` isn't under `autonolas-marketplace/docs` (the path in the PR description looks stale); @DIvyaNautiyal07 can point you to the current canonical location. Happy to walk through any of the three.

This is good context to have on in the PR description.

packages/valory/skills/task_execution/behaviours.py
DhairyaPatel7: Both fixed in 0f086ad9 — good catches, neither was deliberate.

**1. Committed bytes ≠ served bytes.** The success write now keeps the committed object verbatim under `response`, with the envelope (`request_id`/`status`/`content_cid`) hung *outside* it rather than merged into the body:

```python
self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[req_id] = {
    "request_id": req_id,
    "status": "ok",
    "content_cid": local_cid,
    "response": response,   # exact bytes hashed into content_cid
}
```

So a consumer re-derives `compute_cidv1(json.dumps(stored["response"])) == content_cid` byte-for-byte — the offchain delivery is now verifiable, and it mirrors the on-chain path which also content-addresses `json.dumps(response)`. The success test asserts the round-trip so it can't silently drift again.

**2. Failed task served as `status: "ok"`.** Fixed — the offchain branch now checks `_invalid_request` *first* and routes execution failures through the same terminal `_record_offchain_failure` channel the CID/done-task failures already use, so they come back as `status: "rejected"` with the failure reason, never `"ok"`. A poll can now distinguish "ran but failed" from "succeeded". Added `test_handle_done_task_offchain_invalid_request_recorded_as_failure`.

One thing worth your eye: I reused the existing `"rejected"` status (so all three offchain failure modes are consistent) rather than introducing a distinct `"failed"`. A failed task therefore isn't finalized/delivered (no `done_task`, no on-chain settlement) — which I think is right for prepay (don't bill a failure), but happy to split out a dedicated `"failed"` status if you'd rather keep "client refused" and "ran but errored" distinct on the wire.

All CI checks pass locally (lock_check, copyright/deps, linters, hashes, abci specs, 569 skill tests).

DhairyaPatel7:
Added the PR-series breakdown to the description (under **PR series**), plus a **Review updates** section covering the off-chain return channel, the verifiable-commitment fix, and the fail-honest status — all addressed in 0f086ad9. The two inline points are resolved there too.

bennyjo (dismissed):
Approving for the dark Phase-1 ship. ✅

The two contracts I flagged last round are correctly addressed at head (`0f086ad9`):
- **Verifiable commitment** — the served body now carries the committed object verbatim under `response`, with the envelope (`request_id`/`status`/`content_cid`) kept outside it, so `compute_cidv1(json.dumps(stored["response"]))` reproduces `content_cid`. The prior envelope-merge made every off-chain delivery unverifiable; that's fixed.
- **Fail-honest status** — a `_invalid_request` off-chain task now routes through the rejection channel (`status: "rejected"`) instead of being served as `status: "ok"` with the error string as the result.

The dark gate is real: `is_offchain` is only set on tasks enqueued past the `use_offchain` 503 gate, so with the default `false` none of the new behaviour-side code runs and the on-chain path is unchanged.

Two non-blocking items worth pinning before `use_offchain` flips on:

1. **Tighten the verification round-trip test** (`test_handle_done_task_offchain_skips_ipfs_and_finalizes_locally`). It re-hashes the same in-memory `response` object that produced `content_cid`, so it only proves `compute_cidv1` is deterministic — not the parse→serve→reparse→re-hash path real consumers use. A `served = json.loads(json.dumps(stored))` round-trip before re-hashing would exercise the step that actually matters (key order / float reformatting). Cheap to do now.
2. **JSON canonicalization** — the verifiability fix holds for CPython consumers on `json.dumps` defaults (separators / `ensure_ascii` / float repr / key order). A non-Python verifier can't reproduce `content_cid`. Pin the canonical encoding and share it with mech-client / mech-interact / the PR3 ETL before they verify against it.

Framework audits (FSM / resilience) are N/A: no `fsm_specification`, no ABCI round changes, no outbound `get_http_response` calls.

bennyjo (approved):

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -897,6 +903,75 @@ def _handle_done_task(self, task_result: Any) -> None:
...
+                # content_bytes is exactly what the CID commits to. The serving
+                # payload below carries this same object verbatim under
+                # "response" (envelope fields hang outside it), so a client that
+                # re-derives compute_cidv1(json.dumps(stored["response"]))
+                # reproduces content_cid byte-for-byte and can verify the
+                # delivery against the on-chain commitment. json.dumps round-trips
+                # stably here (default separators, dict insertion order is
+                # preserved through JSON), matching the on-chain path which also
+                # content-addresses json.dumps(response).
+                content_bytes = json.dumps(response).encode("utf-8")
```
OjusWiZard, line 936: **[Advisory — Phase 2, non-blocking]** The re-hash verification contract isn't serialization-canonical across consumers. `compute_cidv1` hashes raw bytes with no JSON canonicalization, and this commits Python's default `json.dumps` output — which emits `', '`/`': '` separators and `ensure_ascii=True`. So only a verifier that reproduces *that exact* serialization gets a matching CID; a JS `JSON.stringify` re-hash (no spaces, non-ASCII unescaped) would reject every valid offchain delivery.

This matches the existing on-chain path (same `json.dumps`), so it's **not a regression** — but the offchain feature is the first to advertise "anyone with the content can re-hash and verify," which makes the gap matter here. Before `use_offchain` flips on, worth either committing a canonical encoding (`json.dumps(response, sort_keys=True, separators=(',', ':'), ensure_ascii=False)`) or documenting "verifiers must use Python-compatible serialization."

OjusWiZard (approved):
Re-approving. ✅

Reviewed the two new offchain commits (`2f0ab30a`, `0f086ad9`) that landed after my Jun-15 approval (auto-dismissed by the pushes). They directly resolve the near-theoretical advisory I raised at approval — the offchain return channel that could serve a rejection while the task delivered.

Verified against the code, not the commit messages:

- **Return channel**: offchain success now persists `{request_id, status:"ok", content_cid, response}` to `offchain_request_responses`, so `/fetch_offchain_info` serves the real result instead of the metadata-only `done_task` fallback.
- **Committed == served bytes**: the hashed object is stored verbatim under `"response"` with envelope fields hung outside it; the success test asserts `compute_cidv1(json.dumps(stored["response"])) == content_cid`, so it can't silently drift.
- **Fail-honest**: `_invalid_request` is checked before CID computation, so a "ran but failed" task routes through `_record_offchain_failure` → `status:"rejected"` and can never land in the `"ok"` bucket. New guard test covers it. All three offchain failure modes now share the one `"rejected"` channel.

One **non-blocking advisory** posted inline (serialization canonicalization) — worth pinning before `use_offchain` flips on in Phase 2, but consistent with the existing on-chain path and not a regression, so it does not block the dark merge.

PR #455, opened 2026-06-16
[3/3] feat(task_execution): off-chain preimage retention buffer (kv_store)

DhairyaPatel7:
**PR 3/3 of the off-chain mech series.** PRs 1 + 2 have merged; this is the last piece.

| | |
|---|---|
| **PR 1/3** | #454 — skip-IPFS off-chain + local CID + structured 402 + `Payment-Receipt`. **Merged.** |
| **PR 2/3** | [valory-xyz/kv-store#16](https://github.com/valory-xyz/kv-store/pull/16) — adds `DELETE_REQUEST` / `LIST_REQUEST` performatives + pagination. **Merged**, released as `v0.7.0-rc1`. |
| **PR 3/3** | **this PR** — kv_store-backed preimage buffer + retention sweeper. |

## What this adds

In off-chain mode the mech keeps the request/response **private** (skip-IPFS + local CID, PR1), so there is no public record an operator can use to prove what was requested and delivered. This adds a durable, operator-private **preimage buffer**: each off-chain `(request, response)` pair is mirrored into the persistent `valory/kv_store` connection keyed by `request_id`, and a background **sweeper** prunes entries past a configurable retention window (default 24h).

`task_execution` is a `SimpleBehaviour` (outbox + handler, not ABCI generators), so:

- **`utils/preimage.py`** — pure, unit-tested bookkeeping over `shared_state` (record accept/settle, write-queue de-dup, sweeper expiry).
- **`behaviours.py`** — a write/delete/sweep loop draining **one kv op at a time** via a flag separate from the task executor's `in_flight_req`, so preimage I/O never blocks task execution. Records the request on accept, the response on settle (delivered/rejected).
- **`handlers.py` `KvStoreHandler`** — on `LIST_RESPONSE` queues expired keys for deletion; on `ERROR` re-queues a failed write so a transient kv_store failure can't lose a preimage.
- **wiring** — `valory/kv_store` connection + protocol in `skill.yaml` and agent `aea-config.yaml`; `store_path` → the agent's persistent volume (`/data/mech_preimages`).

## Ships dark

`preimage_retention_enabled` defaults **False** (mirrors `use_offchain`), so on-chain + IPFS behaviour is unchanged until enabled in the Phase 2 rollout.

## Integration status

- Merged `origin/main`. The fingerprint conflicts and the semantic conflict in `_handle_done_task` (PR1 added the in-memory `offchain_request_responses` return channel here; this branch adds the durable buffer write here) are both resolved — the in-memory channel is populated first so `/fetch_offchain_info` works immediately, then the buffer mirrors the same payload for audit + restart-survival.
- Vendored `valory/kv_store` synced to `v0.7.0-rc1` (kv-store main, the merge of #16). `[tool.tomte] upstream_pins` pins `valory-xyz/kv-store@0.7.0-rc1` so `check-third-party-hashes` resolves — same shape optimus and meme-ooorr already declare for kv-store / mech-interact / genai / funds-manager.
- Added `peewee` to `pyproject.toml` and `tox.ini` (used by the vendored kv_store SQLite connection) and allowlisted it in `liccheck` (MIT-licensed but no PyPI Trove classifier, same pattern as the existing anchorpy / based58 / jsonalias entries).
- `KV_STORE_CONNECTION_PUBLIC_ID` is now imported from the connection module so `check-handlers` no longer flags the connection package as unused. The previous string-only `PublicId.from_str("valory/kv_store:0.1.0")` didn't count.
- All CI green: lint (black, isort, flake8, mypy, pylint, darglint), packaging (`check-hash`, `check-packages`, `check-dependencies`, `check-third-party-hashes`, `check-handlers`, `check-abciapp-specs`, `check-abci-docstrings`, `lock_check`), tests across the build matrix, copyright, gitleaks, liccheck.

## Tests

- `tests/utils/test_preimage.py` — pure buffer/sweep bookkeeping (accept, settle-with/without-accept, write-queue de-dup, expiry by `settled_at` / `accepted_at`, malformed-skip).
- `tests/test_preimage_buffer.py` — behaviour flush/sweep loop (first-tick sweep, write flush, batched deletes, in-flight gating, disabled no-op) and `KvStoreHandler` (LIST_RESPONSE→delete-queue, terminal-write pop, processing-write keep, ERROR re-queue).


DhairyaPatel7:
**CI status note (blocked draft):** the two red checks — `linter_checks` → *Check packages* and `copyright_and_dependencies_check` → *Check dependencies* — are both `autonomy packages sync` failing to download the **unpublished** `valory/kv_store` (the documented blocker). `lock_check` passes on all 3 OS, which confirms the vendored kv_store + pinned hashes are internally consistent; the lint tools themselves are green locally. Both go green once kv-store#16 publishes and the repin in the description runs.

Also re-vendored kv_store at the de-guarded protobuf hashes (protocol `bafybeiak5ezgop…`, connection `bafybeibbkpotdes…`) per benny's protobuf-5.29 finding on kv-store#16.

OjusWiZard (commented):
Review of the off-chain preimage retention buffer (author-written `task_execution` code; vendored `kv_store` protocol/connection treated as out-of-scope upstream copies). One blocking correctness bug in the write/settle interleaving, plus kv-op lifecycle robustness gaps and the matching test gaps. Inline comments below. Findings are from static review (suite not run locally).

packages/valory/skills/task_execution/handlers.py
```diff
@@ -516,6 +518,68 @@ def handle(self, message: Message) -> None:
...
+            inflight = shared_state.get(preimage_buffer.PREIMAGE_INFLIGHT_WRITE)
+            if inflight is not None:
+                records = shared_state.get(preimage_buffer.PREIMAGE_RECORDS, {})
+                record = records.get(inflight)
+                if (
+                    record is not None
+                    and record.get("settlement_status")
+                    in preimage_buffer.TERMINAL_STATUSES
+                ):
+                    records.pop(inflight, None)
```
OjusWiZard, line 633: **🔴 Critical — settlement during an in-flight `processing` write silently loses the terminal preimage.**

The SUCCESS branch decides whether to drop the record by inspecting its **current** status, not the status that was actually sent. Race:

1. `_process_preimage_buffer` flushes the `processing` record — `_send_kv_write` serializes it *at send time*, sets `INFLIGHT_WRITE=req_id`, `KV_IN_FLIGHT=True`, and the id is `pop`'d off the write queue.
2. Before this SUCCESS is handled (≥1 act cycle later), the task settles. `record_settlement` mutates `records[req_id]` **in place** to `delivered`/`rejected` and re-enqueues `req_id`.
3. This SUCCESS arrives → `records[req_id]` is now terminal → `records.pop(inflight)` (line 568) deletes it. But kv_store only got the `processing` snapshot from step 1.
4. Next tick drains the queue, pops `req_id`, `records.get(req_id)` is `None` → `continue` (behaviours.py:1553). The terminal write never happens.

Net: kv keeps a stale `processing` row (eventually swept by `accepted_at`); the response/CID are never recorded — defeating the audit-copy purpose. Single-threaded loop + I/O-bound kv round-trip makes the window real, especially for fast tools and the fast-fail invalid-request path.

**Minimal fix:** snapshot the sent status alongside `INFLIGHT_WRITE` and only pop when the live status still equals it (else leave it — the re-enqueued write persists the terminal state). Equivalent guard: don't pop if `req_id` is still in `PREIMAGE_WRITE_QUEUE`. This path is untested (see the handler tests below).

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1448,9 +1473,130 @@ def _record_offchain_failure(self, req_id: str, reason: str) -> None:
...
+
+        write_queue: List[str] = shared_state.get(
+            preimage_buffer.PREIMAGE_WRITE_QUEUE, []
+        )
+        while write_queue:
+            request_id = write_queue.pop(0)
+            record = shared_state.get(preimage_buffer.PREIMAGE_RECORDS, {}).get(
+                request_id
+            )
+            if record is None:
```
OjusWiZard, line 1666: Tail of the Critical race (see handlers.py SUCCESS branch): after the handler pops the record on the in-flight `processing` write, the re-enqueued terminal write lands here with `record is None` and is silently skipped — so the `delivered`/`rejected` preimage is never persisted.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1448,9 +1473,130 @@ def _record_offchain_failure(self, req_id: str, reason: str) -> None:
...
+        kv_store request is in flight at a time (a flag separate from the task
+        executor's ``in_flight_req``), so this never blocks task execution.
+        Per tick it does the first of: run a due sweep (LIST by prefix), flush a
+        batch of expired deletes, or flush one queued write.
+        """
+        if not self.params.preimage_retention_enabled:
+            return
+        shared_state = self.context.shared_state
+        if shared_state.get(preimage_buffer.PREIMAGE_KV_IN_FLIGHT):
+            return
```
OjusWiZard, line 1624: **🟠 Important — no watchdog for a stuck `PREIMAGE_KV_IN_FLIGHT`.**

`KV_IN_FLIGHT` is set `True` by every `_send_kv_*` and cleared **only** in `KvStoreHandler.handle()`. If a kv reply (SUCCESS *or* ERROR) is never delivered — connection drop, lost envelope, or an exception in the connection's send path that swallows the reply — this gate returns forever (silently). Meanwhile `record_accept`/`record_settlement` keep filling `PREIMAGE_RECORDS` / `PREIMAGE_WRITE_QUEUE` in memory → unbounded growth until restart, no diagnostics.

The payment-model path already establishes the fix pattern (`PAYMENT_MODEL_REQUEST_TIMEOUT`, ~lines 98 / 269-282): stamp an `in_flight_since` time and reset the flag (and retry) past a timeout. Lower probability than the race, but high blast radius.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1448,9 +1473,130 @@ def _record_offchain_failure(self, req_id: str, reason: str) -> None:
...
+
+        now = time.time()
+        last_sweep = shared_state.get(preimage_buffer.PREIMAGE_LAST_SWEEP, 0.0)
+        if now - last_sweep >= self.params.preimage_sweep_interval:
+            shared_state[preimage_buffer.PREIMAGE_LAST_SWEEP] = now
+            self._send_kv_list()
+            return
```
OjusWiZard, line 1634: **🟡 Advisory — sweep timer advanced before the LIST is sent.** `PREIMAGE_LAST_SWEEP` is set to `now` here, before `_send_kv_list`. A transient LIST `ERROR` is then indistinguishable from "nothing expired," so sweeping stalls for a full `preimage_sweep_interval` (default 1h) on any transient failure. Advance the timestamp only on a successful `LIST_RESPONSE`.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -516,6 +518,68 @@ def handle(self, message: Message) -> None:
...
+        elif performative == KvStoreMessage.Performative.ERROR:
+            self.context.logger.warning(f"kv_store error: {kv_msg.message}")
+            inflight = shared_state.get(preimage_buffer.PREIMAGE_INFLIGHT_WRITE)
+            if inflight is not None:
+                # Re-queue the failed write so the preimage isn't lost.
+                shared_state.setdefault(
+                    preimage_buffer.PREIMAGE_WRITE_QUEUE, []
+                ).append(inflight)
```
OjusWiZard: **🟡 Advisory — ERROR re-queue uses `append`, not the dedup'd `_enqueue_write`.** Can produce duplicate write-queue entries (harmless — the dup hits the `record is None` skip — but inconsistent with the rest of the queue discipline). Route through `_enqueue_write` instead.

Note the surrounding branch only re-queues when `INFLIGHT_WRITE is not None`, i.e. writes. A failed DELETE/LIST (where `INFLIGHT_WRITE` is `None`) is dropped here — that part is self-healing via the next sweep (keys remain in kv, still expired), so it's not data loss, but it's currently untested.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -0,0 +1,198 @@
...
+        try:
+            record = json.loads(raw)
+            stamp = record.get("settled_at") or record.get("accepted_at")
+        except (ValueError, TypeError, AttributeError):
+            continue
+        if stamp is None:
+            continue
+        if now - float(stamp) > retention_seconds:
+            expired.append(key)
+    return expired
```
OjusWiZard, line 317: **🟡 Advisory — silent skips have no observability.** Swallowing parse/timestamp errors with bare `continue` is correct for delete-safety, but emits nothing. If the stored schema ever drifts (e.g. `serialize` changes), every entry silently skips, retention pruning stops, and kv grows unbounded with no signal. A post-loop `WARNING` with the skipped count would make a degraded/garbled namespace visible at near-zero cost when count is 0.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -0,0 +1,198 @@
...
+    record.update(
+        {
+            "response": response_payload,
+            "response_cid": response_cid,
+            "settled_at": int(now),
+            "settlement_status": status,
+        }
+    )
+    records[request_id] = record
+    _enqueue_write(shared_state, request_id)
```
OjusWiZard: **🟡 Advisory — double-settle overwrites a good kv row with a minimal one.** A second `record_settlement` for the same id *after* the first terminal write was popped from `PREIMAGE_RECORDS` lands in the `records.get(...) or {...}` fallback (`request=None`, `accepted_at=None`) and re-queues it — overwriting the full kv record with a stripped one. Low likelihood, but worth an early-return + WARNING when the existing record is already in a terminal status.

packages/valory/skills/task_execution/tests/test_preimage_buffer.py
```diff
@@ -0,0 +1,209 @@
...
+    handler = _handler(handler_context)
+    ss = handler_context.shared_state
+    preimage.record_accept(ss, "r1", "req", now=1.0)
+    ss[preimage.PREIMAGE_INFLIGHT_WRITE] = "r1"
+    ss[preimage.PREIMAGE_KV_IN_FLIGHT] = True
+    handler.handle(
+        SimpleNamespace(performative=KvStoreMessage.Performative.SUCCESS, message="ok")
+    )
+    assert "r1" in ss[preimage.PREIMAGE_RECORDS]
+
```
OjusWiZard, line 201: **🟠 Important — test gaps that would have caught the Critical race + the watchdog gap.** These two SUCCESS tests cover clean snapshots only. Missing, in priority order:

1. **Settle-during-in-flight** (the Critical bug): accept → flush sets `INFLIGHT_WRITE="r1"` → `record_settlement` mutates the record to terminal → deliver `SUCCESS`; assert the terminal preimage is **still persisted**, not lost.
2. **Write loop skipping a `None`/pruned record** (behaviours.py:1553) and proceeding to the next queued id.
3. **DELETE/LIST returning `ERROR`** — only the write-ERROR re-queue (`INFLIGHT_WRITE` set) is tested; the `INFLIGHT_WRITE is None` branch is not.
4. **`_buffer_preimage_settlement` no-op when `preimage_retention_enabled=False`** — call sites at behaviours.py:991 / :1478 are unexercised; a removed/inverted guard would pollute state on retention-off deployments undetected.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Fixed in 73b0b59d using your alternative guard (the simpler of the two): the SUCCESS branch now refuses to pop the record when `inflight in PREIMAGE_WRITE_QUEUE` — that's exactly the signal a re-enqueue happened after the in-flight write was sent.

```python
if (
    record is not None
    and record.get("settlement_status") in preimage_buffer.TERMINAL_STATUSES
    and inflight not in write_queue
):
    records.pop(inflight, None)
```

Walkthrough:
- Clean path (no race): the behaviour loop already `pop(0)`'d the id before sending, so `inflight not in write_queue` → record gets popped. ✓
- Racy path: `record_settlement` ran while in flight and re-enqueued the id, so `inflight in write_queue` → record stays. Next tick flushes the terminal record. On its SUCCESS the queue is empty again → record pops. ✓

`test_handler_success_keeps_record_if_settle_raced_with_write` walks the full four-step race (accept → flush → settle → SUCCESS-for-processing arrives → assert terminal record + write_queue retained).

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Tail addressed by the same fix at handlers.py SUCCESS: the terminal record is no longer popped on the in-flight write's SUCCESS, so the re-enqueued terminal write the next tick sends finds the record present and persists. `test_handler_success_keeps_record_if_settle_raced_with_write` covers the full path; `test_process_buffer_skips_none_record_and_continues` pins the loop's `record is None` skip so a regression that put the loop in that branch (e.g. another race) wouldn't silently lose preimages.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Done in 73b0b59d, mirroring the payment-model pattern you flagged. Added `PREIMAGE_KV_REQUEST_TIMEOUT = 60.0` and a new `PREIMAGE_INFLIGHT_SENT_AT` shared-state key. Every `_send_kv_*` stamps the timestamp; the handler clears all three of `KV_IN_FLIGHT` / `INFLIGHT_WRITE` / `INFLIGHT_SENT_AT` on any reply. `_process_preimage_buffer` checks the timestamp at the top of the tick and, if the in-flight stretch is over the timeout, logs a WARNING and clears the gate so the loop can resume (any pending write stays in the queue, so the next tick retries it).

`test_process_buffer_resets_stuck_kv_in_flight` pins the reset + that no fresh op fires immediately after (pinning the sweep clock to recent in the fixture so the reset isn't masked).

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Done in 73b0b59d. Moved `PREIMAGE_LAST_SWEEP = now` from `_process_preimage_buffer` (where it stamped on send, ahead of the round-trip) into the LIST_RESPONSE branch of `KvStoreHandler` (stamps on a successful response only). A transient LIST ERROR now leaves the timestamp untouched, so the next tick is eligible to retry. `test_handler_list_response_advances_sweep_timestamp` pins both directions (LIST_RESPONSE advances; ERROR does not).

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Done in 73b0b59d. Made `_enqueue_write` public as `enqueue_write` (the only API caller is the ERROR branch in the handler) and routed the ERROR re-queue through it. Queue stays dedup'd. `test_handler_error_requeue_dedupes_via_enqueue_write` pins it (the test seeds "r1" already in the queue and asserts the queue stays length 1 after the ERROR).

The DELETE/LIST ERROR path (where `INFLIGHT_WRITE is None`) was untested too; added `test_handler_delete_or_list_error_self_heals` — asserts no re-queue (correct, no in-flight write to retry), no in-flight flag stuck, and the next sweep tick picking up the same expired keys is the self-heal mechanism.

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: Done in 73b0b59d. `expired_keys` now collects the skipped keys in a list and logs a single WARNING post-loop when non-zero:

```
Preimage sweep: 17 entries skipped (unparseable or timestamp-less).
First skipped key: 'mech_preimage/abc'
```

Zero-cost when the namespace is healthy (no log line emitted); turns a stuck retention pruner into a visible signal as soon as a single row starts skipping. Kept the existing safety semantics (skipped rows still aren't deleted).

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: Done in 73b0b59d. `record_settlement` now early-returns + WARNINGs when the existing record is already in `TERMINAL_STATUSES`:

```python
if existing is not None and existing.get("settlement_status") in TERMINAL_STATUSES:
    _logger.warning(
        "Double-settle ignored for request_id=%s (already %s).",
        request_id, existing.get("settlement_status"),
    )
    return
```

The full kv row stands; the strip-then-overwrite is impossible. Low likelihood overall (the executor doesn't normally settle twice) but the guard means a future regression that called settle twice can't silently strip the row.

packages/valory/skills/task_execution/tests/test_preimage_buffer.py
DIvyaNautiyal07: Done in 73b0b59d. Added all four tests in priority order plus two related ones for the new code:

1. `test_handler_success_keeps_record_if_settle_raced_with_write` — the settle-during-in-flight race (the Critical bug). Walks all four steps.
2. `test_process_buffer_skips_none_record_and_continues` — the loop's `record is None` skip path; seeds a pruned id and a real id in the queue, asserts exactly one write fires for the real id.
3. `test_handler_delete_or_list_error_self_heals` — ERROR with `INFLIGHT_WRITE is None`; asserts no re-queue and in-flight cleared.
4. `test_buffer_settlement_is_noop_when_retention_disabled` — builds a minimal behaviour-shaped object with `preimage_retention_enabled=False`, calls `_buffer_preimage_settlement`, asserts shared_state stays empty.

Plus the new code's own tests: `test_process_buffer_resets_stuck_kv_in_flight` (watchdog), `test_handler_list_response_advances_sweep_timestamp` (sweep clock moves on success only), and `test_handler_error_requeue_dedupes_via_enqueue_write` (queue dedup discipline).

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1448,9 +1480,156 @@ def _record_offchain_failure(self, req_id: str, reason: str) -> None:
...
+                # any pending write stays in PREIMAGE_WRITE_QUEUE so the next
+                # tick retries it.
+                self.context.logger.warning(
+                    "Preimage kv op has been in flight for "
+                    f">{PREIMAGE_KV_REQUEST_TIMEOUT}s with no reply. "
+                    "Resetting PREIMAGE_KV_IN_FLIGHT so the loop can retry."
+                )
+                shared_state[preimage_buffer.PREIMAGE_KV_IN_FLIGHT] = False
+                shared_state[preimage_buffer.PREIMAGE_INFLIGHT_WRITE] = None
+                shared_state[preimage_buffer.PREIMAGE_INFLIGHT_SENT_AT] = None
```
OjusWiZard: **🟠 Follow-up (non-blocking) — the watchdog drops the in-flight write instead of retrying it.**

The comment says *"any pending write stays in `PREIMAGE_WRITE_QUEUE` so the next tick retries it"* — but that's only true for DELETE/LIST. The write-flush loop does `request_id = write_queue.pop(0)` **before** `_send_kv_write`, so an in-flight write's id is **no longer in the queue**. When the watchdog nulls `PREIMAGE_INFLIGHT_WRITE` here without re-enqueuing, the id is in neither the queue nor the in-flight slot.

- **Processing-write timeout** → self-recovers: a later `record_settlement` re-`enqueue_write`s it (record still in `PREIMAGE_RECORDS`).
- **Terminal-write timeout** → **permanent loss**: the record is already `delivered`/`rejected`, and the double-settle guard added in this same commit early-returns on an already-terminal record — closing the only other re-enqueue path. The preimage is never persisted and its record leaks in memory.

`test_process_buffer_resets_stuck_kv_in_flight` doesn't catch it: it sets `INFLIGHT_WRITE="r1"` with no record and an empty queue, so there's nothing to lose.

**Minimal fix** (dedup-safe):
```python
inflight = shared_state.get(preimage_buffer.PREIMAGE_INFLIGHT_WRITE)
if inflight is not None:
    preimage_buffer.enqueue_write(shared_state, inflight)
shared_state[preimage_buffer.PREIMAGE_KV_IN_FLIGHT] = False
shared_state[preimage_buffer.PREIMAGE_INFLIGHT_WRITE] = None
shared_state[preimage_buffer.PREIMAGE_INFLIGHT_SENT_AT] = None
```
and update the test to seed a terminal record for `"r1"` and assert it lands back in `PREIMAGE_WRITE_QUEUE`. Precondition is a kv op that never replies (rare, local kv), so non-blocking — but it's the exact audit-loss class this feature targets.

OjusWiZard (dismissed):
Approving. The fix commit (73b0b59) correctly addresses all 8 of my earlier threads — the settle/write race guard, the watchdog, the sweep-timer move, ERROR dedup, `expired_keys` observability, and the double-settle guard are all sound, and the pure unit suite is green locally.

One non-blocking follow-up below on the watchdog's write-retry path (a narrow data-loss edge the fix's own comment claims is handled but isn't). Worth tightening before the Phase 2 rollout that enables `preimage_retention_enabled`, but it doesn't gate this draft. 🚢

packages/valory/skills/task_execution/handlers.py
```diff
@@ -516,6 +518,68 @@ def handle(self, message: Message) -> None:
...
+        :param message: the incoming kv_store message.
+        """
+        kv_msg = cast(KvStoreMessage, message)
+        self.context.kv_store_dialogues.update(kv_msg)
+        shared_state = self.context.shared_state
+        performative = kv_msg.performative
+
+        if performative == KvStoreMessage.Performative.LIST_RESPONSE:
+            expired = preimage_buffer.expired_keys(
+                dict(kv_msg.data),
```
bennyjo, line 580: **Problem:** the retention sweeper only ever deletes from the first 100 preimages and silently ignores the rest.
**Impact:** once a mech has >100 live preimages in the retention window (trivial over 24h), storage grows without bound and the retention guarantee doesn't actually hold — an abusive requester can push this into a disk-fill DoS. This is also the opposite of what the merge commit claims ("pagination-aware LIST_REQUEST").

**Why:** the sweep reads `kv_msg.data` but never `kv_msg.next_cursor`, and `behaviours.py:1589 _send_kv_list()` sends `LIST_REQUEST` with only `key_prefix` (no `limit`/`cursor`). The connection resolves an unset `limit` to `_LIST_DEFAULT_PAGE_SIZE = 100` and returns the remainder behind `next_cursor` (connection.py `list_request`/`_resolve_page_size`). So each sweep sees one page of 100 keys — ordered by key, not by age — and never pages through the rest.

Could the sweep carry `next_cursor` across ticks until the store reports no more pages? There's also no test for a `next_cursor`-bearing `LIST_RESPONSE`, which is why this passed green.

packages/valory/skills/task_execution/handlers.py
bennyjo, line 633: **Problem:** under a common timing, the response half of a preimage is never persisted — only the request is.
**Impact:** the buffer exists to prove *what was delivered*, but for any request that settles quickly (e.g. fast rejections via `_record_offchain_failure`) the kv_store can end up holding just the "processing" record with no response, then sweep it away. The guarantee the PR sells is quietly lost for those requests.

**Why:** this pop assumes a SUCCESS corresponds to the record in its *current* state, but the record can be settled after the in-flight write was dispatched. Sequence: processing write dispatched (`inflight=r1`) → task settles, so `record_settlement` flips `r1` to terminal and re-queues it → the **processing** write's SUCCESS lands here, sees a terminal status, and pops `r1` → the re-queued terminal write then finds `record is None` in `_process_preimage_buffer` and is skipped.

Would tagging the in-flight write with the status it actually dispatched (and only popping on a match) work? No test covers settle-while-processing-write-in-flight.

packages/valory/skills/task_execution/handlers.py
bennyjo: **Problem:** a failing kv_store write is retried forever, every tick, with no backoff and no cap.
**Impact:** if the kv_store is persistently unhealthy (db locked, disk full), the agent spins on the same write each `act()` tick and never releases the record — a hot loop plus steadily growing memory if several records keep failing. A degraded kv_store should fail quietly, not amplify.

**Why:** the ERROR path re-queues `inflight` with no attempt counter and no delay, and the record is never dropped, so the cycle send→ERROR→re-queue repeats indefinitely.

Could we bound this with an attempt counter + drop-with-loud-log (or backoff)? (Related: a failed DELETE at `_send_kv_delete` isn't re-queued at all — it relies on the next sweep, which finding #1 may prevent from re-surfacing it.)

bennyjo (commented):
Three findings on the preimage sweep/flush loop — one blocking (pagination not consumed, so retention isn't actually enforced), plus a response-write drop race and an unbounded retry. Details inline.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Good catch, this was the blocking one. Fixed in 1c2b6aac.

`_send_kv_list` now passes `limit=preimage_list_page_size` (default 100, matches the server) and `cursor` pulled from `PREIMAGE_LIST_CURSOR`. The LIST_RESPONSE handler reads `next_cursor`: non-empty saves it to shared_state, empty clears it and stamps `PREIMAGE_LAST_SWEEP`. `_process_preimage_buffer` short-circuits to `_send_kv_list` on every tick while the cursor is non-empty, overriding the sweep-interval gate, so the namespace is walked end to end before the next sweep window starts.

Two new tests: `test_process_buffer_keeps_paging_until_cursor_clears` pins the cursor-overrides-time-gate behaviour; `test_handler_list_response_mid_page_saves_cursor_only` pins both the mid-page (save cursor, clock untouched) and final-page (clear cursor, stamp clock) branches.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Right, terminal-write timeout was the gap. Fixed in 1c2b6aac with your exact suggested patch.

The watchdog now calls `enqueue_write(shared_state, inflight)` before clearing the three flags. `enqueue_write` dedupes, so it's safe even if a settle already re-enqueued the id. Renamed the test to `test_process_buffer_resets_stuck_kv_in_flight_and_resends` and reseeded it: terminal record present, id pop'd off the queue, in-flight stamp set to 999s in the past. After one tick the queue gets the id back, the watchdog clears the gate, and the fall-through fires a fresh CREATE_OR_UPDATE for `r1` with a recent stamp. Added `test_process_buffer_watchdog_resets_when_inflight_is_none` to keep the LIST/DELETE branch covered (no re-queue, gate still clears).

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Agreed, fixed in 1c2b6aac.

Added `PREIMAGE_WRITE_ATTEMPTS: Dict[str, int]` shared-state key + `preimage_max_write_attempts` skill param (default 5). The ERROR branch increments the per-id counter; at the cap the record + counter entry are both dropped and a WARN is logged with the kv_store error message. SUCCESS on a terminal write clears the counter so it stays bounded by the live record set, not lifetime traffic. Tests: `test_handler_error_drops_record_after_max_attempts` (cycles through 3 ERRORs with the cap set to 3, asserts cap reached on the third) and `test_handler_success_clears_attempt_counter`.

Re the related DELETE point: kept the existing self-heal. The next sweep tick re-LISTs and re-picks-up the same expired keys (they're still in kv). With pagination correctly consumed now (#1), that path actually completes the walk, so a single dropped DELETE isn't a leak.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Same diagnosis as OjusWiZard's Critical race comment from earlier in this review pass. Already fixed in 73b0b59d (in this PR).

The SUCCESS branch now refuses to pop when `inflight in PREIMAGE_WRITE_QUEUE`. That's exactly the signal a re-enqueue happened after the processing write was sent: `record_settlement` mutates the record to terminal AND calls `enqueue_write`, so seeing the id back in the queue means a settle landed in flight. The race walkthrough you described is pinned by `test_handler_success_keeps_record_if_settle_raced_with_write` (accept → flush → settle → SUCCESS-for-processing arrives → assert terminal record + queue entry both retained → next tick flushes the terminal write).

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1448,9 +1480,188 @@ def _record_offchain_failure(self, req_id: str, reason: str) -> None:
...
+                # An in-flight write has already been pop(0)'d off the write
+                # queue (see below), so it lives in neither the queue nor the
+                # in-flight slot — without an explicit re-enqueue here a
+                # terminal write (settle followed by a stuck CREATE_OR_UPDATE)
+                # would leak: the double-settle guard in record_settlement
+                # blocks the only other re-enqueue path. enqueue_write dedupes,
+                # so this is safe even if a settle already re-enqueued it.
+                inflight = shared_state.get(preimage_buffer.PREIMAGE_INFLIGHT_WRITE)
+                if inflight is not None:
+                    preimage_buffer.enqueue_write(shared_state, inflight)
```
OjusWiZard: **[Important — Phase 2, non-blocking] The new `preimage_max_write_attempts` cap doesn't cover this no-reply path.** The cap added in this commit only fires on an explicit kv_store `ERROR` reply (`handlers.py` ERROR branch increments `PREIMAGE_WRITE_ATTEMPTS`). This watchdog re-enqueue — which fires when the kv_store accepts the envelope but never replies (dropped envelope / connection drop) — re-queues the write **without** touching the attempt counter. So a kv_store stuck in the no-reply failure mode retries the same write forever and never drops the record, reopening the exact unbounded-retry hole the cap was added to close (just via timeout instead of ERROR). It's timeout-gated so not a hot loop, but it is an unbounded stall + unbounded retention for that id.

Suggest counting this re-enqueue toward the same cap:
```python
inflight = shared_state.get(preimage_buffer.PREIMAGE_INFLIGHT_WRITE)
if inflight is not None:
    attempts = shared_state.setdefault(preimage_buffer.PREIMAGE_WRITE_ATTEMPTS, {})
    attempts[inflight] = attempts.get(inflight, 0) + 1
    if attempts[inflight] >= self.params.preimage_max_write_attempts:
        # drop + WARN, mirroring the ERROR branch
        shared_state.get(preimage_buffer.PREIMAGE_RECORDS, {}).pop(inflight, None)
        attempts.pop(inflight, None)
    else:
        preimage_buffer.enqueue_write(shared_state, inflight)
```

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1448,9 +1480,188 @@ def _record_offchain_failure(self, req_id: str, reason: str) -> None:
...
+            else:
+                return
+
+        # A non-empty cursor means the previous LIST_RESPONSE returned
+        # ``next_cursor`` — i.e. the namespace has more pages. Keep paging on
+        # every tick (deletes/writes wait until the namespace is fully walked)
+        # so retention pruning isn't stuck on the first page when the namespace
+        # has more than preimage_list_page_size entries.
+        if shared_state.get(preimage_buffer.PREIMAGE_LIST_CURSOR):
+            self._send_kv_list()
```
OjusWiZard, line 1623: **[Advisory — Phase 2, non-blocking] The LIST sweep path has no failure bound — the same hot-loop just fixed for writes still applies here.** `PREIMAGE_LAST_SWEEP` is stamped only on a successful *final* page, and a LIST `ERROR` carries no `inflight`, so nothing throttles it: a persistently failing LIST re-lists on every `act()` tick with no cap or backoff. With the pagination added in this commit, a page-2+ failure mid-walk does the same — the cursor stays set, so this branch re-requests the failing page every tick indefinitely. This is the identical unbounded-hot-loop class bennyjo flagged for writes (now bounded via `preimage_max_write_attempts`), left unbounded for LIST. The initial-LIST case is partly pre-existing, but the multi-page loop is new here. Lower priority since the buffer is best-effort, but worth a bound/backoff for symmetry before Phase 2.

OjusWiZard (dismissed):
Re-approving. ✅

Reviewed the one new commit since my last pass (`1c2b6aac`). It resolves my watchdog write-retry follow-up (terminal-write timeout now re-enqueues — exactly the patch I suggested) and bennyjo's three threads (first-page-only sweep → real pagination; unbounded write retry → `preimage_max_write_attempts` cap; settle/write race → already guarded in `73b0b59d`). Pagination control flow (in-flight gate → cursor continuation → sweep → delete → write) is correct, `next_cursor` is a required LIST_RESPONSE field so the `or ""` access is safe, and the new tests cover each path. Verified against the code, not the commit messages.

Two non-blocking follow-ups posted inline — both about failure-bound symmetry, neither gates this draft but both want closing before `preimage_retention_enabled` flips on in Phase 2:

- **[Important]** the new write-retry cap is bypassed by the watchdog no-reply path (unbounded retry survives for dropped-envelope failures).
- **[Advisory]** the LIST sweep path still has the unbounded hot-loop that was just fixed for writes.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1448,9 +1480,188 @@ def _record_offchain_failure(self, req_id: str, reason: str) -> None:
...
+            return
+
+        delete_queue: List[str] = shared_state.get(
+            preimage_buffer.PREIMAGE_DELETE_QUEUE, []
+        )
+        if delete_queue:
+            keys = list(delete_queue)
+            delete_queue.clear()
+            self._send_kv_delete(keys)
```
bennyjo, line 1655: **Retention pruning can permanently stall, and then the 24h-deletion promise silently breaks while the db grows without bound.** Each sweep drains *every* expired key across the whole namespace into a single DELETE; once that batch is large enough the delete fails outright, and because failed deletes aren't retried the next sweep rebuilds the identical oversized batch and fails the same way. The trigger isn't exotic — an agent down longer than the retention window comes back, everything expires at once, and the first sweep hits the cliff.

Mechanism: the one-shot DELETE exceeds SQLite's bound-parameter cap (`connection.py:419`, `Store.key.in_(keys)`). Would you consider chunking deletes (one batch per `preimage_list_page_size`, drained over successive ticks)?

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -91,8 +96,19 @@
 IPFS_MAX_TASK_BYTES = 1_048_576  # 1MB cap on attacker-controlled task payload
 MAX_PROMPT_BYTES = 100_000  # 100KB cap on the prompt field
 PAYMENT_MODEL_REQUEST_TIMEOUT = 60.0  # reset stuck payment-model in_flight_req
+# Reset PREIMAGE_KV_IN_FLIGHT if no kv_store reply arrives within this many
+# seconds. Mirrors PAYMENT_MODEL_REQUEST_TIMEOUT: without this watchdog a lost
+# SUCCESS/ERROR (connection drop, dropped envelope, exception in the
+# connection's send path) wedges the flag True forever — the preimage loop
+# stops draining and PREIMAGE_RECORDS / PREIMAGE_WRITE_QUEUE grow unbounded
+# until restart, silently.
+PREIMAGE_KV_REQUEST_TIMEOUT = 60.0
```
bennyjo: **This 60s in-flight timeout is sized for a remote service, but the kv_store here is local SQLite where a healthy reply is sub-millisecond.** Whenever a reply does go missing, the single-in-flight gate is held this entire 60s and the whole preimage pipeline makes no progress in the meantime. Dropping it to something much smaller shrinks that stall window for free and bounds the blast radius of a lost reply. Cheap change, worth doing now.

bennyjo (commented):
Live-readiness review of the preimage buffer: three gaps that surface once the flag is enabled — chunked deletes, an oversized in-flight timeout, and at-rest retention semantics. (Two lower-likelihood items — the never-settled in-memory leak and the timeout-path retry cap — left out as edge cases for now.)

packages/valory/skills/task_execution/models.py
```diff
@@ -129,6 +129,38 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        # today's behaviour (no durable preimage buffer). When enabled, each
+        # off-chain (request, response) pair is mirrored into the kv_store and a
+        # background sweeper prunes entries older than the retention window. Only
+        # meaningful alongside use_offchain (on-chain deliveries are already
+        # public on IPFS).
+        self.preimage_retention_enabled: bool = kwargs.get(
+            "preimage_retention_enabled", False
+        )
+        # Retention window for buffered preimages, in seconds (default 24h).
+        self.preimage_retention_seconds: int = kwargs.get(
```
bennyjo, line 148: **Retention here is logical, not actual — "deleted" preimages stay recoverable from disk past the window, which undercuts the privacy guarantee this feature exists to provide.** A delete at the boundary removes the row logically, but the plaintext request/response (PII-bearing prompts) remains readable in the db files until later writes happen to overwrite it. This isn't probabilistic — it's the behaviour on every delete.

Mechanism: WAL journaling with no `secure_delete` leaves freed pages and old WAL frames intact (`connection.py:172-179`). Worth either enabling `secure_delete` / periodic `VACUUM`, or documenting retention as logical-only and relying on volume-level encryption at rest.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Good catch. Fixed in 17071777 with your exact suggested patch.

The watchdog now mirrors the ERROR-branch logic: increment `PREIMAGE_WRITE_ATTEMPTS[inflight]`, drop the record + WARN when the count reaches `preimage_max_write_attempts`, otherwise `enqueue_write` (dedup'd) and reset the flags. A no-reply kv_store can no longer bypass the per-id cap. Test: `test_watchdog_counts_toward_write_attempts_and_drops_at_cap` walks two ticks with the cap set to 2 — first tick increments + re-enqueues + fall-through resends; second tick hits the cap and the record + counter both purge.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Done in 17071777, symmetric to the write-side cap.

Added `PREIMAGE_LIST_ATTEMPTS` (int) + `preimage_max_list_attempts` param (default 5). The handler's ERROR branch routes by a new `PREIMAGE_INFLIGHT_OP` shared-state key ("list" / "delete" / "write") since LIST and DELETE both have `INFLIGHT_WRITE=None` and were previously indistinguishable. On a LIST ERROR the counter bumps; at the cap the cursor is cleared, `LAST_SWEEP` is stamped, and a WARN is emitted so the next `preimage_sweep_interval` is the natural backoff. A successful LIST_RESPONSE resets the counter to 0 so failures are counted consecutively, not cumulatively. Covers both initial-LIST and page-2+ mid-walk failures. Tests: `test_handler_list_error_caps_consecutive_failures` walks 3 ERRORs with the cap set to 3; `test_handler_list_response_resets_attempt_counter` pins the reset on success.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Confirmed and fixed in 17071777. This was the real disk-fill risk in the buffer.

Changed the DELETE drain to slice at most `preimage_list_page_size` keys per tick instead of the whole queue. The remaining ids stay in `PREIMAGE_DELETE_QUEUE` for subsequent ticks (with the LIST pagination from the previous round, a full namespace walk + drain completes within one sweep cycle). Bounds the request to well below SQLite's `SQLITE_MAX_VARIABLE_NUMBER` cap, so an agent coming back after a long downtime can no longer build an oversized DELETE that fails on every retry and stalls retention pruning forever. Test: `test_delete_queue_chunked_to_page_size_per_tick` seeds 7 keys with page_size=3, asserts exactly one DELETE of 3 keys goes out and the other 4 stay queued in order.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Done in 17071777. `PREIMAGE_KV_REQUEST_TIMEOUT`: 60.0 → 5.0. Sized for a local SQLite kv_store where a healthy reply is sub-millisecond — 5s is huge headroom in the happy case while shrinking the lost-reply stall to one or two `act()` ticks instead of a minute. Comment updated to call out the local-SQLite assumption.

packages/valory/skills/task_execution/models.py
DIvyaNautiyal07: Acknowledged — we're not addressing this in #455. Our retention guarantee here is storage-bound, not privacy-bound: at steady state the db plateaus at "peak 24h worth of preimages" instead of growing unbounded, which is what we need for operator disk budgeting. The plaintext-recoverable-from-freed-pages property is a separate threat model (file-level access to the box) that we explicitly don't claim to defend against in this feature.

If we want the actual privacy property too, it belongs in a follow-up kv-store PR (`PRAGMA secure_delete=ON` on the connection's SQLite init, optionally periodic `VACUUM`) rather than here — the mech doesn't own the .db file lifecycle. Happy to file that as a tracking issue against kv-store, but flagging it as out of scope for this PR so we don't block the storage fix on a different repo.

OjusWiZard (commented):
Round-4 pass (full review stack) over the preimage buffer. Everything prior is resolved; these are new, verified against the code. Two Important, three Advisory — none Critical, none block the dark merge, but the two Important items want closing before `preimage_retention_enabled` flips on in Phase 2. Minor comment-rot, a test gap, and a small dedup suggestion are folded into the inline notes rather than filed separately.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1448,9 +1482,231 @@ def _record_offchain_failure(self, req_id: str, reason: str) -> None:
...
+                # dropped envelope, etc). Clear the gate so the loop can drain.
+                # An in-flight write has already been pop(0)'d off the write
+                # queue (see below), so it lives in neither the queue nor the
+                # in-flight slot — without an explicit re-enqueue here a
+                # terminal write (settle followed by a stuck CREATE_OR_UPDATE)
+                # would leak: the double-settle guard in record_settlement
+                # blocks the only other re-enqueue path. enqueue_write dedupes,
+                # so this is safe even if a settle already re-enqueued it.
+                inflight = shared_state.get(preimage_buffer.PREIMAGE_INFLIGHT_WRITE)
+                if inflight is not None:
```
OjusWiZard, line 1554: **[Important] The retry cap mirrored here only covers write timeouts — the LIST *no-reply* path is still uncapped.** `inflight` is the `PREIMAGE_INFLIGHT_WRITE` slot, which is `None` for a LIST op, so this whole block is skipped on a LIST timeout. The `preimage_max_list_attempts` cap (handlers.py:620) only fires on a LIST **ERROR** reply — a LIST that *times out* (kv_store accepts the request, never replies) never reaches the handler. Net: `PREIMAGE_LIST_CURSOR` stays set and `PREIMAGE_LAST_SWEEP` un-stamped, so the loop re-LISTs every `PREIMAGE_KV_REQUEST_TIMEOUT` (5s) forever — the exact unbounded-retry corner that was just closed for the ERROR path, reopened for no-reply.

Fix: read `PREIMAGE_INFLIGHT_OP` **before** it's cleared at line 1578, and when it's `OP_LIST` increment `PREIMAGE_LIST_ATTEMPTS` + clear the cursor / stamp `LAST_SWEEP` at the cap (mirror handlers.py:620-630). Non-blocking; matters before `preimage_retention_enabled` flips on.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -0,0 +1,261 @@
...
+        try:
+            record = json.loads(raw)
+            stamp = record.get("settled_at") or record.get("accepted_at")
+        except (ValueError, TypeError, AttributeError):
+            skipped.append(key)
+            continue
+        if stamp is None:
+            skipped.append(key)
+            continue
+        if now - float(stamp) > retention_seconds:
```
OjusWiZard: **[Important] `float(stamp)` is outside the try/except above (227-232), so it breaks this function's own contract.** The docstring promises rows "whose value can't be parsed are left untouched — the sweeper must never delete a row it doesn't understand," and the `try` honors that for `json.loads`. But a row with a JSON-parseable yet non-numeric timestamp (e.g. `{"settled_at":"oops"}`) raises `ValueError` here, which escapes `expired_keys` → `KvStoreHandler` → `act()`. Since that row reappears in every LIST of its page, the sweep crashes on **every** cycle: retention pruning stalls permanently and the agent throws each sweep — a poison-pill.

Not reachable today (current writers always store `int(now)`), so this is a defensive gap, not an active bug. Fix: pull `float(stamp)` into the `try` (catch `ValueError`/`TypeError` → `skipped`), and add a malformed-stamp test.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -516,6 +518,155 @@ def handle(self, message: Message) -> None:
...
+                        "Preimage sweep LIST failed %d times in a row; "
+                        "giving up the current walk. Next attempt in "
+                        "preimage_sweep_interval. Last kv_store error: %s",
+                        list_attempts,
+                        kv_msg.message,
+                    )
+                    shared_state[preimage_buffer.PREIMAGE_LIST_CURSOR] = None
+                    shared_state[preimage_buffer.PREIMAGE_LAST_SWEEP] = time.time()
+                    shared_state[preimage_buffer.PREIMAGE_LIST_ATTEMPTS] = 0
+            inflight = shared_state.get(preimage_buffer.PREIMAGE_INFLIGHT_WRITE)
```
OjusWiZard, line 685: **[Advisory] No `OP_DELETE` arm in this ERROR branch — a failed DELETE silently drops its key batch.** A DELETE has `PREIMAGE_INFLIGHT_WRITE=None` and `inflight_op==OP_DELETE`, so it matches neither the `OP_LIST` block above nor this `inflight is not None` block. The keys already sliced off `PREIMAGE_DELETE_QUEUE` (behaviours.py ~1620) are dropped with only the generic `kv_store error` WARN at line 605. It self-heals (next sweep re-LISTs and re-queues), so impact is bounded — but the loss is unobservable. Suggest an `elif inflight_op == preimage_buffer.OP_DELETE:` arm logging the dropped count.

Two related notes: (a) the `PREIMAGE_INFLIGHT_OP` doc comment (preimage.py:63-66) says it routes DELETE "to the right counter" — DELETE has no counter, only LIST does; worth correcting. (b) `test_handler_delete_or_list_error_self_heals` leaves `INFLIGHT_OP=None`, so it never exercises the `OP_DELETE` path — a test asserting `PREIMAGE_LIST_ATTEMPTS` stays `0` after a DELETE error would lock in that DELETE doesn't borrow the LIST counter.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -0,0 +1,261 @@
...
+    :param shared_state: the skill's shared state dict.
+    :param request_id: the off-chain request id.
+    :param response_payload: the delivered response, or the failure reason.
+    :param response_cid: the local CID of the response, when delivered.
+    :param status: one of STATUS_DELIVERED / STATUS_REJECTED.
+    :param now: the current epoch time in seconds.
+    """
+    records = shared_state.setdefault(PREIMAGE_RECORDS, {})
+    existing = records.get(request_id)
+    if existing is not None and existing.get("settlement_status") in TERMINAL_STATUSES:
```
OjusWiZard, line 220: **[Advisory] This double-settle guard only fires while the record is still in `PREIMAGE_RECORDS`.** Once a terminal write flushes and the SUCCESS handler pops it (handlers.py ~597), `existing` is `None`, so a second `record_settlement` for that id falls through to the fallback at 193, builds a stripped record (`request=None, accepted_at=None`), and `enqueue_write` overwrites the good kv row. Low reachability (single task slot; callers return right after settling), so probably fine to leave — but the comment is misleading: it opens by describing the **post-pop** scenario ("already flushed and popped") as if this guard caught it, when that's exactly the case it does *not* cover. Either track settled ids in a never-cleared set, or reword to say the guard is in-memory-only and the post-pop overwrite is intentionally unguarded.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -0,0 +1,261 @@
...
+    }
+    enqueue_write(shared_state, request_id)
+
+
+def record_settlement(
+    shared_state: Dict[str, Any],
+    request_id: str,
+    response_payload: str,
+    response_cid: Optional[str],
+    status: str,
```
OjusWiZard, line 193: **[Advisory] `status: str` accepts any string, including `STATUS_PROCESSING`** — which would write a record the SUCCESS handler never pops (it only pops `TERMINAL_STATUSES`), leaking that id in `PREIMAGE_RECORDS` forever. Both call sites pass a terminal status today; narrowing to `Literal["delivered", "rejected"]` (or asserting `status in TERMINAL_STATUSES` at the top) makes the leak unrepresentable.

packages/valory/skills/task_execution/models.py
bennyjo: Makes sense — agreed this is correctly scoped out of #455. The storage-bound plateau is the property that matters for the disk-budget goal here, and `secure_delete`/`VACUUM` rightly belongs on the kv_store connection (mech only vendors it), not patched into the vendored copy. Please link the kv-store tracking issue here when you file it so the thread closes against something concrete.

One small follow-up unrelated to the code: the PR description frames the buffer as keeping the response "private" / "operator-private," which can read as a privacy guarantee. Since the actual claim is storage-bound (not "wiped from disk after 24h"), worth a one-line clarification in the operator-facing docs so nobody downstream over-reads the retention window as cryptographic erasure.

packages/valory/skills/task_execution/behaviours.py
bennyjo, line 1554: **If the kv_store hangs on a LIST (stops replying rather than returning an error), preimage retention silently dies: the sweep never finishes and no new preimages get written until the agent restarts.**

Scenario: a degraded/overloaded kv_store connection that times out instead of erroring. The careful back-off you added for *errored* LISTs (`preimage_max_list_attempts`) doesn't cover the *no-reply* case — the watchdog only counts write timeouts, so a hung LIST loops every tick forever and blocks the write/delete drain behind it. Healthy local SQLite never hits this, so it's confined to a genuinely stuck connection — but that's the exact failure the watchdog exists for. Worth extending the same LIST back-off to the timeout path.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -516,6 +518,155 @@ def handle(self, message: Message) -> None:
...
+                else:
+                    # Re-queue the failed write so the preimage isn't lost.
+                    # Route through enqueue_write so the queue stays dedup'd
+                    # (matches the discipline of every other enqueue in the
+                    # buffer module).
+                    preimage_buffer.enqueue_write(shared_state, inflight)
+
+        shared_state[preimage_buffer.PREIMAGE_INFLIGHT_WRITE] = None
+        shared_state[preimage_buffer.PREIMAGE_INFLIGHT_OP] = None
+        shared_state[preimage_buffer.PREIMAGE_KV_IN_FLIGHT] = False
```
bennyjo, line 721: **A kv reply that arrives after the 5s watchdog gave up can briefly let two kv ops run at once, corrupting the bookkeeping the loop depends on (wrong op's retry counter incremented, gate opened early).**

Scenario: the connection stalls past 5s, the watchdog moves on and starts the next op, then the slow reply finally lands and gets misattributed to that new op. Very unlikely on local SQLite (sub-ms replies), so fine to leave if that's a conscious call — flagging only because the whole loop is built on the one-op-at-a-time gate, and this is the one path that can violate it. A check that the reply matches the op actually in flight would close it.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -1060,6 +1211,16 @@ def _enqueue_offchain_request(
         except Exception:
             self._rollback_offchain_enqueue(request_id)
             raise
+        # Buffer the request half of the durable preimage (no-op unless off-chain
+        # preimage retention is enabled). The response half is added at
+        # settlement; the behaviour flushes both to the kv_store asynchronously.
+        if self.params.preimage_retention_enabled:
+            preimage_buffer.record_accept(
+                self.context.shared_state,
+                request_id,
```
bennyjo, line 1276: Confirmation, not a defect: the preimage is keyed purely on `request_id`, so if a requester could reuse another requester's `request_id`, their preimage write would overwrite the victim's audit row. That's a non-issue as long as `request_id` is marketplace-derived (hash of request + sender) and not free-form caller input — just want it explicit, since the buffer's integrity rests entirely on that assumption.

bennyjo (commented):
Three findings on the preimage buffer — one resilience gap (LIST timeout path) plus a single-in-flight edge case and a request_id-uniqueness confirmation. Details inline.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in d854e36c — same INFLIGHT_OP routing on the watchdog path that the ERROR branch already uses. Read PREIMAGE_INFLIGHT_OP before the cleanup; if OP_LIST and timed out, bump PREIMAGE_LIST_ATTEMPTS and at the cap clear the cursor + stamp LAST_SWEEP. PREIMAGE_INFLIGHT_DIALOGUE is intentionally NOT cleared in the watchdog (any late reply for the timed-out op now hits the new dialogue-mismatch guard in the handler and is dropped). Test: `test_watchdog_caps_list_no_reply` walks two ticks with the cap set to 2 — first tick bumps the counter and re-fires LIST; second tick hits the cap, cursor clears, LAST_SWEEP stamps, no fresh LIST.

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: Fixed in d854e36c. `float(stamp)` is now inside the `try`, catching `ValueError` / `TypeError` from a JSON-parseable-but-non-numeric timestamp (`{"settled_at": "oops"}`). Such rows go to `skipped` exactly like the unparseable-JSON / missing-timestamp rows already did, honoring the docstring's 'don't touch what we don't understand' contract. Test: `test_expired_keys_treats_non_numeric_stamp_as_skipped` mixes a poison-pill row with a real expired row and asserts only the real one is returned (the poison-pill goes to skipped instead of raising and crashing the sweeper).

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Done in d854e36c. Added an `elif inflight_op == preimage_buffer.OP_DELETE:` arm that logs the dropped batch size: `Preimage sweep DELETE failed; N expired key(s) will be retried on the next sweep. Last kv_store error: ...`. Surfaced via a new `PREIMAGE_INFLIGHT_DELETE_COUNT` shared-state key stamped in `_send_kv_delete(len(keys))` and cleared by the handler + watchdog. Doc comment for `PREIMAGE_INFLIGHT_OP` rewritten to say 'routes ERROR replies to the right arm' (was 'right counter' which only applies to LIST/WRITE). Test: `test_handler_delete_error_logs_dropped_count_and_keeps_list_counter` seeds INFLIGHT_OP=OP_DELETE + count=7, fires ERROR, asserts the WARN message contains '7 expired key(s)' AND that PREIMAGE_LIST_ATTEMPTS is unchanged (DELETE doesn't borrow the LIST counter).

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: Reworded in d854e36c — agreed the old text was misleading. The new comment is explicit: this is an in-memory-only guard that catches the pre-flush double-settle, and the post-pop case is intentionally unguarded because (a) tracking settled ids in a never-cleared set would leak memory unbounded and (b) the executor is single-threaded and returns immediately after settling, so there's no plausible code path to trigger it. Wording the guard's actual scope avoids a future reader assuming defense in depth that isn't there.

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: Done in d854e36c — `assert status in TERMINAL_STATUSES` at the top of `record_settlement` with a descriptive message. Both call sites pass terminal status today; the assertion makes a future regression that forwards STATUS_PROCESSING fail loudly instead of leaking the id in PREIMAGE_RECORDS forever. Test: `test_record_settlement_rejects_non_terminal_status` calls with STATUS_PROCESSING and asserts AssertionError.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in d854e36c. Same watchdog extension OjusWiZard flagged from a different angle: a stuck LIST (no reply) now bumps PREIMAGE_LIST_ATTEMPTS on the timeout path, and at `preimage_max_list_attempts` clears the cursor + stamps LAST_SWEEP so the next attempt is gated by `preimage_sweep_interval` (1h default) instead of `PREIMAGE_KV_REQUEST_TIMEOUT` (5s). Test `test_watchdog_caps_list_no_reply` walks the two-tick scenario end to end.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Closed in d854e36c. Added a dialogue-reference late-reply guard at the top of `KvStoreHandler.handle()`: `_send_kv_*` now stamps the dialogue reference into `PREIMAGE_INFLIGHT_DIALOGUE`; the handler compares the incoming reply's reference against it and ignores anything that doesn't match. The watchdog intentionally leaves `PREIMAGE_INFLIGHT_DIALOGUE` set when it gives up (cleared only by the next `_send_kv_*` or a matching reply), so a late reply for the timed-out op still has a stale-but-non-None expected to mismatch against — exactly the path you flagged. Test: `test_handler_ignores_reply_from_previously_timed_out_op` stamps a 'new op' dialogue ref, fires a reply whose ref doesn't match, asserts all current op state is untouched.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Confirmed: `request_id` is marketplace-derived, not free-form. For marketplace requests it's the on-chain `requestId` emitted by `MechMarketplace.requestForwarded` (deterministic from the requester + nonce + tool args, no caller control over the value). For the off-chain HTTP path it's set by the signed-request handler from the requester signature flow — same uniqueness guarantee. So preimage rows are keyed on a value the requester can't choose, and no cross-requester overwrite is possible. Worth pinning in the test_preimage_buffer integration tests if we add them; tracking separately.

packages/valory/skills/task_execution/models.py
DIvyaNautiyal07: Picked up the doc clarification piece in  — `utils/preimage.py` module docstring + `models.py` comment on `preimage_retention_seconds` both spell out 'storage-bound, not cryptographic erasure' explicitly and point at `secure_delete`/volume encryption as the path to the actual privacy property. Operator-facing docs (handoff README + spec) get the same one-liner when I touch them next.

Kv-store tracking issue still pending — will file against valory-xyz/kv-store with the `PRAGMA secure_delete=ON` + WAL frame retention details and link the issue here when it's up.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -516,6 +518,201 @@ def handle(self, message: Message) -> None:
...
+        # guard exists for the genuinely stuck-connection case.
+        expected = shared_state.get(preimage_buffer.PREIMAGE_INFLIGHT_DIALOGUE)
+        actual = (
+            dialogue.dialogue_label.dialogue_reference if dialogue is not None else None
+        )
+        if (
+            expected is not None
+            and actual is not None
+            and tuple(actual) != tuple(expected)
+        ):
```
OjusWiZard: **Late-reply guard has a `dialogue is None` blind spot.**

When `kv_store_dialogues.update()` returns `None` (unrecognised message), `actual` is set to `None` at line 551, which short-circuits the guard (`actual is not None` is `False`). Execution falls through to the performative dispatch and then to the unconditional cleanup block (L707-712), which sets `PREIMAGE_INFLIGHT_DIALOGUE = None` and `PREIMAGE_KV_IN_FLIGHT = False`. From that point a real in-flight reply arriving next sees `expected = None` and also passes the guard — potentially misapplying a stale LIST_RESPONSE or SUCCESS to the wrong op state.

Fix: add an early return just before the guard:
```python
if dialogue is None:
    self.context.logger.warning(
        "KvStoreHandler: received message with no matching dialogue; dropping."
    )
    return
```
This is the standard AEA handler pattern and removes the ambiguity entirely.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1448,9 +1482,280 @@ def _record_offchain_failure(self, req_id: str, reason: str) -> None:
...
+                )
+                shared_state[preimage_buffer.PREIMAGE_KV_IN_FLIGHT] = False
+                shared_state[preimage_buffer.PREIMAGE_INFLIGHT_WRITE] = None
+                shared_state[preimage_buffer.PREIMAGE_INFLIGHT_OP] = None
+                shared_state[preimage_buffer.PREIMAGE_INFLIGHT_DELETE_COUNT] = 0
+                # NOTE: keep PREIMAGE_INFLIGHT_DIALOGUE around — clearing it
+                # here would silently let a slow reply for the timed-out op
+                # match against the next op (no expected ref → no guard).
+                # The next _send_kv_* overwrites it with the new dialogue ref,
+                # and any reply arriving in between sees a stale expected
```
OjusWiZard, line 1611: **Watchdog deliberate non-clear of `PREIMAGE_INFLIGHT_DIALOGUE` has no test asserting it.**

The comment correctly explains the invariant: clearing the dialogue ref during a watchdog reset would let a slow arriving reply for the timed-out op match the *next* op's guard (no stored reference → guard skips). Existing watchdog tests (`test_watchdog_caps_list_no_reply`, `test_process_buffer_resets_stuck_kv_in_flight_and_resends`) verify the counter and gate behaviour but don't assert that `shared_state[PREIMAGE_INFLIGHT_DIALOGUE]` is *preserved* (non-None) after the watchdog fires.

A future cleanup that adds `shared_state[PREIMAGE_INFLIGHT_DIALOGUE] = None` alongside the other clears at L1604-1606 would look obviously correct, pass all existing tests, and silently hollow out the late-reply guard. Worth a one-line assertion in both watchdog tests.

OjusWiZard (dismissed):
Pagination implementation in 614e8cdf is correct — empty-table path safe, cursor ordering consistent with SQLite BINARY collation, negative limits rejected at the protocol layer, client sets all three required fields. Test suite covers the tricky exact-multiple-of-limit boundary. Two defensive gaps worth addressing before merge.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -516,6 +518,201 @@ def handle(self, message: Message) -> None:
...
+        # and we ignore it. Healthy local SQLite never hits this; the
+        # guard exists for the genuinely stuck-connection case.
+        expected = shared_state.get(preimage_buffer.PREIMAGE_INFLIGHT_DIALOGUE)
+        actual = (
+            dialogue.dialogue_label.dialogue_reference if dialogue is not None else None
+        )
+        if (
+            expected is not None
+            and actual is not None
+            and tuple(actual) != tuple(expected)
```
bennyjo: This guard rejects **every** real kv_store reply, not just late ones. The reference stamped at send time has an empty responder slot (`(nonce, "")`), but `dialogues.update()` completes it to `(nonce, responder_ref)` before this comparison runs — so the full-tuple check never matches a legitimate first reply.

Impact (only when `preimage_retention_enabled=True`, so it ships dark): every SUCCESS / LIST_RESPONSE / DELETE reply is dropped → `PREIMAGE_KV_IN_FLIGHT` clears only via the 5s watchdog, writes hit the retry cap and get dropped, and the sweep never advances → retention pruning stops and the buffer grows unbounded — the exact failure this feature exists to prevent.

Comparing only the initiator nonce (`actual[0] != expected[0]`) keeps the intended "ignore a timed-out op's reply" behavior without rejecting valid ones.

packages/valory/skills/task_execution/tests/conftest.py
```diff
@@ -90,6 +90,29 @@ def create(self, *a: Any, **k: Any) -> Tuple[SimpleNamespace, SimpleNamespace]:
         return msg, dlg
 
 
+class _KvStoreDLG(_DLG):
+    """kv_store dialogue stub whose create() echoes performative + content."""
+
+    def update(self, _msg: Any) -> Any:
+        """Return an object with a stable dialogue reference."""
+        return SimpleNamespace(
+            dialogue_label=SimpleNamespace(dialogue_reference=("nonce-1", "x"))
```
bennyjo: This mock returns the same `dialogue_reference` from both `create()` and `update()`, which masks the one behavior that breaks the guard in production: real open-aea completes the responder slot on `update()` (`(nonce, "")` → `(nonce, responder_ref)`). As a result the handler suite stays green against code that rejects every real reply, and the guard's accept path is never exercised — the other handler tests leave `PREIMAGE_INFLIGHT_DIALOGUE` unset, which short-circuits it.

Suggest having `update()` return a completed, distinct reference and adding a test asserting a normal reply is processed. That test fails today and passes once the guard compares only the initiator nonce.

bennyjo (commented):
Two blocking findings, one shared root cause: the late-reply dialogue guard rejects every real kv_store reply in production, and the conftest mock hides it.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Good find — this would have been worse than a stuck-op edge case. Fixed in dc6b9b8a with the standard AEA early-return pattern: if `dialogue is None` log a WARNING and `return` before the cleanup block runs. The unconditional cleanup at the end of `handle()` now only runs for messages we actually processed, so PREIMAGE_INFLIGHT_DIALOGUE stays intact when an unrecognised message arrives — the next real reply still has a valid expected ref to match against. Test: `test_handler_drops_unrecognised_message_without_touching_state` monkeypatches `update()` to return None, fires a message, asserts INFLIGHT_DIALOGUE and KV_IN_FLIGHT are both preserved.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Done in dc6b9b8a. Both watchdog tests now stamp PREIMAGE_INFLIGHT_DIALOGUE in setup and assert it survives the watchdog reset:
- `test_process_buffer_watchdog_resets_when_inflight_is_none` (the clean LIST/DELETE case where there's no fall-through) gets the explicit assertion you suggested.
- `test_watchdog_caps_list_no_reply` asserts that even after the cap-hit branch fires (cursor cleared, LAST_SWEEP stamped) the dialogue ref is left in place.

A future patch that drops `PREIMAGE_INFLIGHT_DIALOGUE = None` into the watchdog cleanup block alongside the other clears now fails both tests with a clear message about the late-reply guard regression.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Caught a real bug — thank you. Fixed in dc6b9b8a.

You're right: the send-time stamp is `(nonce, "")` and `dialogues.update()` completes the responder slot before the comparison runs, so a full-tuple check rejects every legitimate reply. Switched to comparing the initiator nonce only (`tuple(actual)[0] != tuple(expected)[0]`) — that's the part we generate and the part that uniquely identifies the op.

With `preimage_retention_enabled=True` the broken guard would have: dropped every SUCCESS so PREIMAGE_KV_IN_FLIGHT only cleared via the 5s watchdog, hit `preimage_max_write_attempts` on every write (record + counter purged), and never advanced the sweep. Buffer would have grown unbounded — exactly the storage failure this feature exists to prevent. Good thing the feature ships dark.

Regression test: `test_handler_accepts_reply_when_initiator_matches` stamps `("nonce-1", "")`, fires a reply with `("nonce-1", "responder-ref")`, and asserts the terminal record is popped and the in-flight gate cleared. Fails on the old code, passes on the new.

packages/valory/skills/task_execution/tests/conftest.py
DIvyaNautiyal07: Confirmed and fixed in dc6b9b8a. The `_KvStoreDLG` stub now mirrors open-aea's send/reply reference flow:

- `create()` returns `(nonce, "")` (empty responder slot, send-time state).
- `update()` returns `(nonce, "responder-ref")` (responder completed by the connection).

This is what surfaced the full-tuple bug at all — with the test added per your suggestion (`test_handler_accepts_reply_when_initiator_matches`) the suite now goes red on the broken guard and green only when the comparison is initiator-only. Docstring on `_KvStoreDLG` explains the convention so a future contributor doesn't collapse the two refs back into one.

OjusWiZard (approved):
Re-approving. Clean pass on dc6b9b8a + 88b5e8d9.

The late-reply guard is now correct: dialogue is None gets an early return before the cleanup block can hollow out PREIMAGE_INFLIGHT_DIALOGUE, and the initiator-nonce-only comparison is the right fix for the full-tuple regression (open-aea's .update() fills the responder slot from "" to a real ref, so the send-time stamp would never match the incoming reply on a full comparison). The conftest mock mirrors real open-aea send/reply reference flow. The three new regression tests (nonce-match accept, None-dialogue drop, watchdog preservation) are the right pins. 88b5e8d9 is mechanical darglint compliance only.

bennyjo (approved):
Re-reviewed. Every blocking item from the earlier rounds is fixed and pinned by a discriminating regression test: LIST pagination is now consumed end-to-end (sweep pages until `next_cursor` empties, `LAST_SWEEP` stamped only on the final page), the settle/write race guard, the bounded write/LIST retry caps, and the late-reply misattribution bug (compare initiator nonce only, since open-aea completes the responder slot on the first reply). Security/attacker surface is clean: kv keys/values go through the parameterized peewee/SQLite connection, DELETE is scoped to prefix-LISTed keys only, the poison-row timestamp crash is handled, and resource use is bounded (chunked DELETE, capped retries, 5s in-flight watchdog). Retention is correctly documented as storage-bound rather than cryptographic erasure. CI green across the full matrix; ships dark behind `preimage_retention_enabled`. 🚀

PR #457, opened 2026-06-23
feat(task_submission): PostTxSettlement round emits wildcard data-lake events

DIvyaNautiyal07:
## Summary

After on-chain settlement confirms, the mech bundles every offchain delivery from that FSM round into a single signed POST to the wildcard data lake. The signed typed-data binds a keccak256 hash of the events array, so the signature covers the row content rather than just metadata; the server (predict-api#162) recomputes the hash and rejects on mismatch.

This is the mech-side half of the offchain analytics path. Pairs with predict-api#162 (server EIP-712 swap).

Spec: `docs/mech_prepay_spec.md` → "Authentication and rate limits" and "Mech-side write client (writes to wildcard)". Status: `docs/mech_prepay_status.md` items A + L.

## What changes

- **`task_execution.behaviours._finalize_done_task`** attaches a structured `wildcard_event` dict to each offchain `done_task` BEFORE the existing `IN_MEMORY_REQUESTS` pop. The payload survives FSM consensus replication via `synchronized_data.done_tasks`, so every agent has the data the PostTxSettlement behaviour needs even though only the originating agent had the raw IPFS_DATA in its local shared_state.
- **`task_submission_abci.rounds`** gains `PostTxSettlementRound` + `FinishedPostTxSettlementRound`, declared as an initial state so composition can wire into it from outside the skill. The round is a `CollectSameUntilThresholdRound`: DONE on consensus participation, NO_MAJORITY only when impossible.
- **`task_submission_abci.behaviours.PostTxSettlementBehaviour`** extracts the wildcard event dicts from synced `done_tasks`, computes the batch hash, builds the EIP-712 typed-data, signs via the existing `get_signature(digest, is_deprecated_mode=True)` path (so the signer is the agent's own owner-EOA), and POSTs to `params.wildcard_events_url`. Every agent posts its own copy under its own EOA; the wildcard server is idempotent on `request_id` so multi-agent services don't need a keeper pattern, and the per-EOA rate limiter on the server scales across co-owners.
- **`task_submission_abci.wildcard_write_client`** mirrors the server's canonical-JSON encoder + batch-hash computation bit-for-bit. The contract is narrow on purpose: any drift invalidates every batch.
- **`mech_abci.composition`** routes `FinishedTransactionSubmissionRound → PostTxSettlementRound → FinishedPostTxSettlementRound → ResetAndPauseRound`. The two-edge wiring keeps the settlement→reset path intact while slotting in the analytics-write side effect. Fail-soft by construction: DONE always, so a wildcard outage cannot stall the FSM.
- **`skill.yaml` + `Params`** add `mech_events_enabled` (default `false`; Phase 1 ships dark) and `wildcard_events_url`. No API key, no env-var secret — the EOA signature is the credential.
- `fsm_specification.yaml` updated in both `task_submission_abci` and `mech_abci`; `autonomy analyse fsm-specs` clean.

## Auth chain (paired with predict-api#162)

```
mech: build batch (events[])
  + batch_hash = keccak256(canonical_json(events))
  + sign typed_data{ mech_service_multisig, batch_hash } with agent's EOA
  + POST { typed_data, signature, events }

server: ecrecover EOA
  + allowlist verifyingContract for chainId
  + EOA ∈ Safe.getOwners(mech_service_multisig)   (cached)
  + checkMech(mech_service_multisig)              (cached)
  + recompute batch_hash from events, compare
  + INSERT all rows in one tx, idempotent on request_id
```

## Tests

- `tests/test_rounds.py::TestPostTxSettlementRound`: round transitions (DONE on threshold, NO_MAJORITY when impossible).
- `tests/test_rounds.py::TestTaskSubmissionAbciApp` extended: new state in `initial_states` and `final_states`, new transitions in `transition_function`.
- `tests/test_wildcard_write_client.py`: canonical-JSON contract (key sort, recursive sort, compact separators, UTF-8 preservation), batch-hash stability and order-sensitivity, typed-data shape pinned against the server's Pydantic schema.
- `mech_abci/tests/test_composition.py` updated: `FinishedTransactionSubmissionRound → PostTxSettlementRound`, `FinishedPostTxSettlementRound → ResetAndPauseRound`.

582 tests pass across `task_submission_abci`, `mech_abci`, and `task_execution` locally.

## Test plan

- [x] `uv run pytest packages/valory/skills/task_submission_abci packages/valory/skills/mech_abci packages/valory/skills/task_execution` — 582 pass
- [x] `uv run autonomy packages lock --check` after `lock` — clean
- [x] `uv run autonomy analyse fsm-specs` — exit 0
- [ ] Full mech CI green on this branch (lock-check, copyright, liccheck, check-dependencies, check-third-party-hashes, check-doc-hashes, check-packages, check-hash, black-check, isort-check, flake8, mypy, pylint, darglint, check-abci-docstrings, check-handlers, gitleaks, py3.10-tests across linux/macos/windows)

## Open questions for the reviewer

1. **`mech_service_multisig` source.** The behaviour reads it from `self.synchronized_data.safe_contract_address` with a `params.safe_contract_address` fallback. Both are standard AEA hooks, but if the multi-mech-per-service config means a different field is more accurate, flag it.
2. **`requested_at` placement.** `_build_wildcard_event` pulls the timestamp from `request_data["datetime"]` (the field name the requester's signed header carries today) with a fallback to `executed_at`. If the field has been renamed somewhere in the offchain handler's plumbing, the wildcard row will fall back to executed_at silently rather than fail; happy to switch to a strict lookup if you'd rather see a 422 on a missing timestamp than a slightly-off row.
3. **Replay buffer follow-up.** This PR drops the batch with an error log on transient wildcard failure (5xx / network drop), matching the Phase 1 dark-ship posture. The kv_store-backed replay buffer + drainer stack on top of mech#455 (preimage retention, currently open draft) and land in a follow-up; the spec calls for it but it's not safe to add here without #455's protocol-level `DELETE` / `LIST` shape merged.

## Companion PR

predict-api#162 — server-side EIP-712 swap. The two PRs ship together: this one is a no-op until the server accepts the new shape, and the server accepts a feature-flagged endpoint that does nothing until this PR's mechs start emitting.


packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1805,6 +1813,227 @@ def _num_to_bytes(value: int) -> bytes:
...
+                encode_typed_data,
+            )
+        except ImportError:
+            self.context.logger.error(
+                "eth_account not available; cannot sign wildcard EIP-712. "
+                "Skipping batch."
+            )
+            return
+        signable = encode_typed_data(full_message=typed_data)
+        digest = signable.body
```
bennyjo: **Impact: when `mech_events_enabled` is flipped on, the server will reject every batch as signer-not-owner, so no analytics rows are ever written.**

`encode_typed_data(full_message=...)` returns a `SignableMessage` whose `.body` is only the message `hashStruct` — it does **not** include the domain separator. The actual EIP-712 digest is `keccak(0x19 ‖ 0x01 ‖ header ‖ body)`, where `header` is the domain separator. Because `get_signature(..., is_deprecated_mode=True)` raw-signs whatever 32 bytes you hand it (no re-hashing), signing `signable.body` produces a signature over the message struct alone. A server doing standard EIP-712 `ecrecover` reconstructs the full digest from `typed_data` and recovers a *different* address → 4xx → batch dropped every round.

A second consequence: since the domain (chainId + verifyingContract) isn't bound into the signature, even a matching signature would be replayable across chains/contracts.

Suggested fix:
```python
digest = keccak(b"\x19" + signable.version + signable.header + signable.body)
```
Could you also confirm this against predict-api#162's `mech_sig.py` recover function? If the server intentionally recovers over `hashStruct` only, that's non-standard and worth a comment here — but then the cross-domain replay point still stands. (Note: the `build_typed_data` docstring claims signing goes through the `gnosis_safe.encode` path, which isn't what this code does.)

bennyjo (commented):
Second-reviewer pass: one blocking EIP-712 signing bug plus a few items to address before the feature flag is enabled. FSM + resilience structure looks sound. Details inline.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1805,6 +1813,227 @@ def _num_to_bytes(value: int) -> bytes:
...
+            from eth_account.messages import (  # type: ignore[import-not-found]
+                encode_typed_data,
+            )
+        except ImportError:
+            self.context.logger.error(
+                "eth_account not available; cannot sign wildcard EIP-712. "
+                "Skipping batch."
+            )
+            return
+        signable = encode_typed_data(full_message=typed_data)
```
bennyjo: **Impact: the signing path has no test, so the digest bug flagged just below is invisible to CI.** `test_wildcard_write_client.py` pins the typed-data *shape* but nothing asserts what bytes actually get signed.

A test that either (a) signs a known batch and recovers the address back, or (b) asserts `digest == keccak(0x19 ‖ 0x01 ‖ header ‖ body)`, would have caught this and would protect the contract with the server going forward. Worth adding alongside the fix.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1805,6 +1813,227 @@ def _num_to_bytes(value: int) -> bytes:
...
+        body = json.dumps(
+            {
+                "typed_data": typed_data,
+                "signature": signature_hex,
+                "events": events,
+            }
+        ).encode("utf-8")
+
+        try:
+            response = yield from self.get_http_response(
```
bennyjo: **Impact: a slow or hung wildcard endpoint can pin this behaviour until the round timeout fires, instead of failing fast.** There's no explicit `timeout=` on the POST, so it inherits the framework default and relies on `ROUND_TIMEOUT` to unstick the round. Consider passing a short explicit `timeout=` so a slow endpoint degrades quickly rather than holding the round open.

(Also minor: a 3xx response currently falls through to the `POST OK` log — a redirected POST would be logged as success without writing.)

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1410,6 +1410,20 @@ def _finalize_done_task(self, cid: str) -> None:
...
+        # Attach the offchain wildcard-event payload before the
+        # IN_MEMORY_REQUESTS pop below; the post-settlement behaviour in
+        # task_submission_abci reads it off ``synchronized_data.done_tasks``
+        # (i.e. after consensus replication) and posts it to the wildcard
+        # data lake, but the buffered request metadata it needs is local
+        # to this agent's shared_state and goes away at the pop. Empty /
+        # None on the on-chain path; the post-settlement behaviour skips
+        # entries where this is absent.
+        if is_offchain:
+            done_task["wildcard_event"] = self._build_wildcard_event(
```
bennyjo, line 1496: **Impact: this PR is not actually "dark." Even with `mech_events_enabled=False`, every offchain delivery now builds and consensus-replicates a full event payload to every agent — so the feature flag doesn't gate the cost the PR description says it does.**

This build runs whenever `is_offchain` is true, with no `mech_events_enabled` check (and `task_execution`'s Params doesn't even have that flag — only `task_submission` does). The HTTP POST is gated, but the build/attach/replicate work isn't. So the moment this lands, the `wildcard_event` dict rides FSM consensus into `synchronized_data` on every agent for every offchain request, regardless of rollout state.

Could you gate the build behind the feature flag too (plumb a flag into `task_execution`, or have `task_submission` build lazily from data already on `done_task` instead of pre-attaching)? That makes Phase 1 genuinely a no-op out of the box.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1420,6 +1434,199 @@ def _finalize_done_task(self, cid: str) -> None:
...
+                    str(request_data.get("model"))
+                    if request_data.get("model")
+                    else None
+                ),
+                "tool_params": (
+                    request_data.get("params")
+                    if isinstance(request_data.get("params"), dict)
+                    else None
+                ),
+                "raw_content": (
```
bennyjo: **Impact: a requester can inflate per-request consensus state on every agent.** The `prompt` field is capped at 100KB, but `raw_content` embeds the entire request (and response, just below) dict uncapped, and that rides Tendermint replication into `synchronized_data` on all agents.

Requester-controlled fields (`params`, `model`, arbitrary request body) flow into `raw_content` without a size bound, then get consensus-replicated and stored. Combined with the flag-gating issue above (this happens even with the flag off), a single large request grows state on every node. Could you cap or drop `raw_content` for the consensus copy — e.g. keep it out of the replicated `done_task` and reconstruct it only at POST time, or truncate it the way `prompt` is truncated?

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1420,6 +1474,199 @@ def _finalize_done_task(self, cid: str) -> None:
...
+            except (ValueError, TypeError):
+                request_data = {"raw": request_data_raw}
+        elif isinstance(request_data_raw, dict):
+            request_data = request_data_raw
+        else:
+            request_data = {}
+        response_data_raw = self.context.shared_state.get(
+            OFFCHAIN_REQUEST_RESPONSES, {}
+        ).get(req_id, {})
+        response_data = response_data_raw if isinstance(response_data_raw, dict) else {}
```
OjusWiZard: **[CRITICAL] Wrong dict level — every successful offchain delivery records as `status="failed"` in analytics.**

`OFFCHAIN_REQUEST_RESPONSES[req_id]` stores an envelope `{"status": "ok", "content_cid": ..., "response": response}` (see line 1006 where it's set). The actual result and executed_at live under the inner `"response"` key, not the envelope itself.

Because `response_data.get("result")` and `response_data.get("executed_at")` are always called on the envelope, both always return `None`, which forces `status = "failed"`, `result = null`, `error = null` for every completed task.

```python
# Fix: drill into the inner response dict
envelope = self.context.shared_state.get(OFFCHAIN_REQUEST_RESPONSES, {}).get(req_id, {})
response_data = envelope.get("response", {}) if isinstance(envelope, dict) else {}
```

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -189,6 +190,46 @@ class FinishedWithoutTasksRound(DegenerateRound):
...
+
+    payload_class = PostTxSettlementPayload
+    payload_attribute = "content"
+    synchronized_data_class = SynchronizedData
+    extended_requirements = ()
+
+    def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
+        """Process the end of the block."""
+        if self.threshold_reached:
+            return self.synchronized_data, Event.DONE
```
OjusWiZard: **[CRITICAL] `done_tasks` not cleared on DONE — re-post hazard on NO_MAJORITY / ROUND_TIMEOUT re-entry.**

`TransactionPreparationRound.end_block` explicitly does `synchronized_data.update(done_tasks=[])` on every exit arm. This round returns the existing `synchronized_data` object unchanged. On `ROUND_TIMEOUT` or `NO_MAJORITY` (both self-loop back per the transition function), the next entry re-reads the same stale `done_tasks` and re-fires the wildcard POST. Server-side idempotency absorbs duplicate POSTs today, but the stale data also survives across periods via `cross_period_persisted_keys`, so the next cycle's `PostTxSettlementBehaviour` sees last period's events too.

```python
if self.threshold_reached:
    return self.synchronized_data.update(done_tasks=[]), Event.DONE
```

OjusWiZard (changes requested):
Five agents reviewed this PR from multiple angles (code correctness, Python idioms, test coverage, silent failures, security). One critical data bug, three FSM-safety bugs, and five correctness issues follow as inline comments. Blocking on the first four.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1805,6 +1814,239 @@ def _num_to_bytes(value: int) -> bytes:
...
+        # produce a signature the server's standard ``ecrecover`` recovers
+        # to a different address; the helper exists to keep this exact
+        # bytes contract testable.
+        try:
+            digest = compute_eip712_digest(typed_data)
+        except ImportError:
+            self.context.logger.error(
+                "eth_account not available; cannot sign wildcard EIP-712. "
+                "Skipping batch."
+            )
```
OjusWiZard, line 1967: **[CRITICAL] Only `ImportError` is caught — other exceptions from `compute_eip712_digest` crash the FSM round.**

`encode_typed_data(full_message=typed_data)` raises `ValueError`/`KeyError` on a malformed typed-data dict. The `keccak(b"\x19" + signable.version + ...)` concatenation raises `TypeError` if any attribute is not `bytes`. These propagate through `async_act` and crash a round that was explicitly designed to be fail-soft (settlement already confirmed on-chain; the analytics row should just be dropped, not stall the FSM).

```python
except Exception as exc:   # same pattern as the signing and HTTP try-blocks below
    self.context.logger.error(
        "Wildcard EIP-712 digest failed; dropping batch: %s", exc
    )
    return
```

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1805,6 +1814,239 @@ def _num_to_bytes(value: int) -> bytes:
...
+                "Wildcard EIP-712 signing failed; dropping batch this round: %s",
+                exc,
+            )
+            return
+
+        body = json.dumps(
+            {
+                "typed_data": typed_data,
+                "signature": signature_hex,
+                "events": events,
```
OjusWiZard: **[CRITICAL] `json.dumps` building the POST body is unguarded — can crash the FSM round.**

The `events` list comes from `synchronized_data.done_tasks` which is requester-influenced (via `raw_content`, tool params, etc.). A non-JSON-serializable value (leaked `bytes`, `float("inf")`, etc.) raises an unhandled `TypeError`/`ValueError` that escapes `async_act`. Same risk applies to `compute_batch_hash` at line 1928.

```python
try:
    body = json.dumps({"typed_data": typed_data, "signature": signature_hex, "events": events}).encode("utf-8")
except (TypeError, ValueError) as exc:
    self.context.logger.warning("Wildcard batch JSON serialisation failed; dropping: %s", exc)
    return
```

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1420,6 +1474,199 @@ def _finalize_done_task(self, cid: str) -> None:
...
+        )
+        # Requested_at: prefer the signed timestamp on the buffered request;
+        # the handler validates and stores it before enqueueing. Fall back to
+        # the executed_at on the response so the time-leading index entry is
+        # always populated for an offchain row.
+        requested_at_raw = request_data.get("datetime") or request_data.get(
+            "requested_at"
+        )
+        requested_at_iso = (
+            str(requested_at_raw) if requested_at_raw else executed_at_iso
```
OjusWiZard: **[BUG] `str()` on an int Unix timestamp yields `"1700000000"`, not an ISO 8601 string — server returns 422.**

The requester can store a Unix integer under `datetime` or `requested_at` in the IPFS blob. `str(1700000000)` is not an `AwareDatetime`-parseable string; the wildcard server will reject the batch with a 422.

```python
if isinstance(requested_at_raw, (int, float)):
    requested_at_iso = datetime.fromtimestamp(requested_at_raw, tz=timezone.utc).isoformat()
elif isinstance(requested_at_raw, str) and requested_at_raw:
    requested_at_iso = requested_at_raw
else:
    requested_at_iso = executed_at_iso
```

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1420,6 +1474,199 @@ def _finalize_done_task(self, cid: str) -> None:
...
+        )
+
+        delivery_rate_raw = executing_task.get(
+            "request_delivery_rate"
+        ) or executing_task.get("delivery_rate")
+        delivery_rate_str = (
+            str(delivery_rate_raw) if delivery_rate_raw is not None else None
+        )
+        nonce_raw = executing_task.get("request_id_nonce") or executing_task.get(
+            "nonce"
```
OjusWiZard: **[BUG] `or`-chain discards a legitimate `0` value for `nonce` and `delivery_rate`.**

When `executing_task["request_id_nonce"]` is `0` (a valid nonce), `0 or executing_task.get("nonce")` evaluates the second operand because `0` is falsy, returning `None` when `"nonce"` is also absent. Same for `delivery_rate_raw`. The nonce ends up `null` in the signed event — the server may use nonce for replay detection.

Use a sentinel-based first-found pattern:
```python
def _first_present(d: dict, *keys: str) -> object:
    _MISSING = object()
    for k in keys:
        v = d.get(k, _MISSING)
        if v is not _MISSING:
            return v
    return None

nonce_raw = _first_present(executing_task, "request_id_nonce", "nonce")
delivery_rate_raw = _first_present(executing_task, "request_delivery_rate", "delivery_rate")
```

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1420,6 +1474,199 @@ def _finalize_done_task(self, cid: str) -> None:
...
+        )
+        try:
+            nonce_int = int(nonce_raw) if nonce_raw is not None else None
+        except (TypeError, ValueError):
+            nonce_int = None
+        # The wildcard server requires `prompt` and `tool` to be non-empty
+        # strings; placeholder them rather than dropping the row.
+        prompt = str(request_data.get("prompt") or "[offchain request]")
+        if len(prompt) > MAX_PROMPT_BYTES:
```
OjusWiZard: **[BUG] Byte cap uses `len()` (code points), not UTF-8 byte length — attacker can exceed the cap by up to 3×.**

`len(prompt)` counts Unicode code points. A CJK prompt of 50 001 code points (well under `MAX_PROMPT_BYTES = 100_000`) encodes to ~150 KB. The variable name and the comment at the module level both say "bytes", but the enforcement is on characters.

```python
prompt_bytes = prompt.encode("utf-8")
if len(prompt_bytes) > MAX_PROMPT_BYTES:
    prompt = prompt_bytes[:MAX_PROMPT_BYTES].decode("utf-8", errors="ignore")
```
This matches how `IPFS_MAX_TASK_BYTES` is enforced elsewhere in the file.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1420,6 +1474,199 @@ def _finalize_done_task(self, cid: str) -> None:
...
+        result_text = response_data.get("result") if response_data else None
+        status = "failed" if invalid or result_text is None else "complete"
+        error_text = (
+            str(result_text)
+            if invalid and result_text
+            else (
+                None
+                if status == "complete"
+                else (str(result_text) if status == "failed" and result_text else None)
+            )
```
OjusWiZard: **[BUG] Dead branch + semantic gap: `invalid=True, result_text=None` → `error_text=None`, losing failure context.**

The inner-else arm (`status == "failed" and result_text`) is unreachable: when the outer condition `invalid and result_text` is false and status is `"failed"`, `result_text` must be falsy (that is exactly the only way to get here). So `str(result_text)` in the inner else never executes.

More importantly, when `invalid=True` and `result_text is None`, the failed row reaches the server with `error=null` and `status="failed"` — no diagnostic content at all.

The whole expression simplifies to:
```python
error_text = str(result_text) if status == "failed" and result_text else None
```

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1805,6 +1814,239 @@ def _num_to_bytes(value: int) -> bytes:
...
+            )
+            return
+
+        # The EIP-712 domain.chainId binds the signature to one chain. Every
+        # event in the batch was built by ``task_execution._build_wildcard_event``
+        # against the same ``mech_events_chain_id`` param, so we can read
+        # the int from the first event rather than duplicating the param
+        # across two skills (and the value-by-construction check below
+        # rejects a heterogeneous batch defensively).
+        chain_id = int(events[0].get("request", {}).get("chain_id", 0) or 0)
```
OjusWiZard, line 1917: **[IMPORTANT] No guard for `chain_id == 0` — misconfigured deployments spend a signing round-trip before the server rejects.**

When `mech_events_chain_id` is not set (default `0`), the domain separator is `chainId=0`. The behaviour proceeds through `compute_batch_hash → build_typed_data → compute_eip712_digest → get_signature → POST`, only to receive a server allowlist rejection. The URL and feature-flag checks both have early returns with clear log messages; a zero-chainId check should too.

```python
if chain_id == 0:
    self.context.logger.warning(
        "mech_events_chain_id is 0 (unconfigured); skipping wildcard write."
    )
    return
```
Add this immediately after the chain-homogeneity check.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1805,6 +1814,239 @@ def _num_to_bytes(value: int) -> bytes:
         return tx_list
 
 
+class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour, ABC):
```
OjusWiZard: **[QUALITY] `ABC` in the signature is misleading — no abstract methods defined, and the class is registered as concrete.**

`PostTxSettlementBehaviour` has no `@abstractmethod`s and appears directly in `TaskSubmissionRoundBehaviour.behaviours` (requiring a `# type: ignore` there because the type-checker flags it as abstract). The other concrete behaviours in this file inherit `ABC` because they are intermediate mixins; this one is not. Drop `ABC` from the class signature to remove the false signal and the suppressor.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1420,6 +1474,199 @@ def _finalize_done_task(self, cid: str) -> None:
...
+        )
+        # IPFS_DATA can land as either a parsed dict (normal handler path)
+        # or a JSON-encoded string (some test fixtures and the historical
+        # backfill path); normalise to dict here so downstream ``.get``
+        # calls don't crash on a string.
+        if isinstance(request_data_raw, (bytes, bytearray)):
+            try:
+                request_data_raw = request_data_raw.decode("utf-8")
+            except UnicodeDecodeError:
+                request_data_raw = ""
```
OjusWiZard, line 1563: **[QUALITY] `UnicodeDecodeError` swallowed with no log — prompt silently becomes `"[offchain request]"` placeholder.**

When the bytes request data is not valid UTF-8, the event emits the placeholder prompt with no log entry. Debugging why analytics rows show the placeholder for a specific request ID is impossible without this.

```python
except UnicodeDecodeError as exc:
    self.context.logger.warning(
        "request_data for req_id=%s is non-UTF-8 bytes; using empty payload: %s",
        req_id, exc,
    )
    request_data_raw = ""
```

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 01e0c0b6. Added `compute_eip712_digest` in `wildcard_write_client.py` that builds the full `keccak(0x19 || version || domainSep || hashStruct)`; the behaviour now signs that instead of `signable.body`. Pinned by `test_digest_matches_standard_eip712_formula` (asserts the formula with a negative control against the old `signable.body` path) and `test_sign_and_recover_round_trip` (signs with a known privkey, recovers via the server's exact `Account.recover_message` path). A regression that reintroduces the bug fails both.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Both tests added in 01e0c0b6 (`test_digest_matches_standard_eip712_formula` for the formula and `test_sign_and_recover_round_trip` for the end-to-end). See the reply on the sibling digest comment for details — they're the (a) and (b) tests you suggested.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Both fixed in 01e0c0b6. Replaced `get_http_response` with `_build_http_request_message` + `_do_request(timeout=...)` and added a `wildcard_events_timeout_seconds` skill param (default 5s) so a slow endpoint surfaces as a dropped batch rather than holding the round open. The success check is now `200 <= status < 300`, so a 3xx no longer falls through to the `POST OK` log; the non-success branch logs the body preview so a misconfigured proxy surfaces with a useful signal.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Good catch. Fixed in 01e0c0b6 by mirroring `mech_events_enabled` into `task_execution`'s `Params` + `skill.yaml`. `_finalize_done_task` now only builds the `wildcard_event` when the flag is on, so with the flag off (default) zero payload rides Tendermint consensus replication.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Capped in 01e0c0b6 via a new `_cap_raw_content` helper with a 256 KiB JSON-serialised cap (matches the wildcard server's `_MAX_TEXT` cap on TEXT columns). Over-cap payloads become `{truncated: True, size_bytes: N, cap_bytes: 262144}` so the row still lands and the analytics ETL can flag it rather than silently storing half. Applied to both request and response halves.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. Confirmed at behaviours.py:1000-1007 — the envelope shape is exactly as you said. ``_build_wildcard_event`` now drills into ``envelope.get('response', {})`` (with the rejection-path fallback to ``{}``). Without this, every completed task was tagged ``status='failed'`` with null result and null error.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Fixed in 330cb8ad. ``PostTxSettlementRound.end_block`` now clears ``done_tasks`` on DONE via ``synchronized_data.update(done_tasks=[])``, matching the ``TransactionPreparationRound`` pattern. Without this, NO_MAJORITY / ROUND_TIMEOUT re-entry would re-read the same events and re-fire the wildcard POST; the events also rode ``cross_period_persisted_keys`` into the next cycle.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. Wrapped the whole ``compute_batch_hash → build_typed_data → compute_eip712_digest`` block in ``except Exception`` (with a log) so a malformed typed-data dict — or any of the ``ValueError``/``KeyError``/``TypeError`` you called out — drops the batch instead of crashing ``async_act``. Matches the fail-soft contract for the round.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. ``json.dumps`` of the POST body is now wrapped in ``except (TypeError, ValueError)`` with a log + early return. ``compute_batch_hash`` falls inside the broader ``except Exception`` block from the digest-build path above, so the same content classes are caught there.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. ``requested_at`` coercion now branches: ``isinstance(value, (int, float))`` → ``datetime.fromtimestamp(value, tz=timezone.utc).isoformat()``; non-empty string → kept as-is; otherwise falls back to ``executed_at_iso``. ``str(int)`` on a Unix timestamp no longer leaks through.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. Replaced the ``or``-chain with explicit ``in``-key probes (``if 'request_id_nonce' in executing_task: ... else executing_task.get('nonce')``), so a legitimate ``0`` in the canonical key wins over a fallback ``None``. Same for ``delivery_rate``.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. Cap is now ``len(prompt.encode('utf-8')) > MAX_PROMPT_BYTES``; on overflow the bytes are sliced to the cap and decoded back with ``errors='ignore'`` to drop any half-character at the boundary. Matches the byte semantics the constant already documented and the ``IPFS_MAX_TASK_BYTES`` enforcement elsewhere in the file.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. Collapsed to the single expression you suggested: ``error_text = str(result_text) if status == 'failed' and result_text else None``. Added a comment noting that a ``failed`` row with no diagnostic content emits ``error=None`` (the wildcard server's shape CHECK allows that).

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. Added an early return right after the chain-homogeneity check: ``if chain_id == 0: logger.warning(...); return``. An unconfigured deployment no longer spends a signing dialogue + HTTP POST per round only to land a server-side allowlist rejection.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. Dropped ``ABC`` from the class signature and removed the ``# type: ignore`` on ``TaskSubmissionRoundBehaviour.behaviours``. Tests + mypy still green.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in 330cb8ad. The ``UnicodeDecodeError`` branch now logs at warning with the request_id and the underlying exception before falling back to the empty payload, so a non-UTF-8 request surfaces in the log instead of silently turning into the placeholder.

OjusWiZard (dismissed):
Second-pass review complete. All 16 first-pass findings are addressed. No new correctness issues found after commits 330cb8ad, d5a02b08, 0fb21637.

Remaining non-blocking items for follow-up:
- Test coverage for `_build_wildcard_event`, `_cap_raw_content`, `_do_wildcard_write_best_effort`, `_extract_offchain_events` (tracked in mech#455)
- Two `getattr` calls on always-set params (cosmetic)
- `-> Generator` missing type params on `_do_wildcard_write_best_effort`

LGTM — ready to merge.

bennyjo (dismissed):
Re-reviewed after the three fix commits. All 16 findings from the prior two passes are correctly resolved (EIP-712 digest + sign/recover test, dark-flag gating the build, raw_content cap, response-envelope drill-in, done_tasks cleared on DONE, fail-soft digest/JSON guards, requested_at int coercion, 0-preserving nonce/rate probes, UTF-8 byte prompt cap, chain_id==0 short-circuit, ABC drop). FSM wiring is consistent across composition.py, both fsm_specification.yaml files, and the AbciApp transition function; CI is fully green. LGTM.

One non-blocking item to keep an eye on (not gating, given the flag ships false): a requester-controlled `requested_at` *string* is forwarded to the server without format validation (task_execution/behaviours.py:376). If predict-api#162 rejects the batch as a single transaction, one malformed timestamp could drop analytics for every co-batched delivery in that round. Worth confirming the server's per-row vs per-batch rejection semantics; coercing to a parseable ISO string (falling back to executed_at) would close it.

bennyjo (approved):

OjusWiZard (approved):

PR #458, opened 2026-06-25
fix(task_execution): skip IPFS on offchain task fetch + canonical Z timestamps

DIvyaNautiyal07:
## Summary

Two follow-ups to #457 surfaced while running the offchain path end-to-end against staging (`mpp.staging.autonolas.tech`). With either gap in place, no offchain delivery ever reaches the wildcard data lake.

### 1. Bypass IPFS for the offchain task fetch

`_execute_task` was unconditionally fetching the request body via `_build_ipfs_get_file_req(ipfs_hash)` — even on the offchain path, where `in_memory_requests[request_id]` already has the prompt JSON in process memory. The handler short-circuits the IPFS *upload* (per #457's design intent) but the symmetric short-circuit on the *get* side was missing.

On a slow / unreachable IPFS endpoint the fetch ate the full 60s timeout and the offchain task never reached pebble. Branches on `is_offchain` now and feeds a synthetic `IpfsMessage` (the inline payload wrapped in `SimpleNamespace(files={\"metadata.json\": ...})`) straight into `_handle_get_task` so the rest of the validation + pebble dispatch stays unchanged.

### 2. Canonical timestamp form (Z, not +00:00)

`_build_wildcard_event` was emitting `requested_at` / `executed_at` / `delivered_at` with the `+00:00` UTC offset suffix:

```
\"requested_at\": \"2026-06-25T13:19:06.651013+00:00\"
```

The wildcard server reparses through Pydantic's `AwareDatetime` and re-emits via `model_dump(mode=\"json\")` which normalises to:

```
\"requested_at\": \"2026-06-25T13:19:06.651013Z\"
```

Same instant, different bytes. The mech's canonical-JSON `batch_hash` then disagreed with the server's recompute, dropping every offchain delivery with `400 BATCH_HASH_MISMATCH` at the very last auth-chain step (after `checkMech` + `getOwners` + owner verification had already passed). All three timestamps now `.replace(\"+00:00\", \"Z\")` to match the canonical form the server hashes.

### 3. Declare new params in `mech_abci`'s args allowlist

`mech_abci.Params` inherits from `TaskExecutionParams` (composition skill), so the three new fields (`mech_events_enabled`, `wildcard_events_url`, `wildcard_events_timeout_seconds`) are *read* via the inherited `kwargs.get`. But `mech_abci`'s own `skill.yaml` `args:` block didn't declare them, so AEA's strict-mode arg allowlist *rejected* any override block trying to flip them on. Result: the feature stayed dark even with the env vars set.

Adds the three fields with the same defaults as `task_submission_abci/skill.yaml` (off by default, empty URL, 5s timeout) so operators can override per deployment.

## E2E verification

Single-agent local run via `aea-helpers run-agent` against this PR's source + the predict-api staging server:

- mech-client sends offchain request → mech returns 200 with `Payment-Receipt` (#457 unchanged)
- Tool runs in pebble (`prediction-offline`)
- Done task with `wildcard_event` payload attached
- On-chain settlement via `deliverMarketplaceWithSignatures`
- `PostTxSettlementBehaviour` fires, signs the batch with the operator EOA
- POST to `https://mpp.staging.autonolas.tech/mech/events` returns **200**
- Mech log: `Wildcard batch POST OK (status=200 batch_size=1 signer-eoa=0x53B6c51B…)`

Before the timestamp fix, the same flow consistently returned `400 BATCH_HASH_MISMATCH` on the recompute step. After the fix the server accepts and inserts the `mech_requests` / `mech_responses` row pair.

## Test plan

- [ ] CI passes (lock + lint + tests)
- [ ] Existing unit tests for `_build_wildcard_event` still pass with the Z-normalised timestamps
- [ ] Add a unit test that asserts `_build_wildcard_event` emits `…Z` suffix on all three timestamp fields
- [ ] Confirm staging DB row landed via production team running the verification queries against `mech_requests` / `mech_responses`

### DIFF
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1680,8 +1680,18 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             # untouched and reintroduce the ``batch_hash`` mismatch the
             # canonicalisation is meant to close. Malformed strings fall
             # through to ``now_iso`` rather than crashing the FSM round.
+            #
+            # The trailing-``Z`` rewrite is needed because
+            # ``datetime.fromisoformat`` rejects ``…Z`` on Python 3.10;
+            # the canonical-``Z`` form is what the server emits and what
+            # JavaScript's ``Date.prototype.toISOString()`` produces, so
+            # a requester re-using the server's own timestamp shape
+            # would otherwise silently fall back to ``executed_at`` and
+            # lose the real request time.
             try:
-                requested_at_iso = _iso_z(datetime.fromisoformat(requested_at_raw))
+                requested_at_iso = _iso_z(
+                    datetime.fromisoformat(requested_at_raw.replace("Z", "+00:00"))
+                )
             except ValueError:
                 self.context.logger.warning(
                     "requested_at=%r for req_id=%s is not parsable ISO 8601; "
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1652,12 +1652,28 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # All three timestamps go through ``_iso_z`` so the canonical-JSON
         # bytes match the server's Pydantic round-trip (see helper docstring).
         now_iso = _iso_z(datetime.now(timezone.utc))
+        # ``datetime.fromtimestamp`` raises ``OverflowError`` (and on some
+        # libcs ``OSError``) for values past the platform's representable
+        # range, plus ``ValueError`` for NaN. The ``executed_at`` field
+        # comes from tool plumbing on the mech itself, but it still rides
+        # the FSM consensus channel and a buggy tool returning a wild
+        # value would otherwise crash the round; fall back to ``now_iso``.
         executed_at_unix = response_data.get("executed_at")
-        executed_at_iso = (
-            _iso_z(datetime.fromtimestamp(executed_at_unix, tz=timezone.utc))
-            if isinstance(executed_at_unix, (int, float))
-            else now_iso
-        )
+        if isinstance(executed_at_unix, (int, float)):
+            try:
+                executed_at_iso = _iso_z(
+                    datetime.fromtimestamp(executed_at_unix, tz=timezone.utc)
+                )
+            except (ValueError, OverflowError, OSError):
+                self.context.logger.warning(
+                    "executed_at=%r for req_id=%s is out of range; "
+                    "falling back to now_iso.",
+                    executed_at_unix,
+                    req_id,
+                )
+                executed_at_iso = now_iso
+        else:
+            executed_at_iso = now_iso
         # Requested_at: prefer the signed timestamp on the buffered request;
         # the handler validates and stores it before enqueueing. Fall back to
         # the executed_at on the response so the time-leading index entry is
@@ -1669,9 +1685,24 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             "requested_at"
         )
         if isinstance(requested_at_raw, (int, float)):
-            requested_at_iso = _iso_z(
-                datetime.fromtimestamp(requested_at_raw, tz=timezone.utc)
-            )
+            # ``datetime.fromtimestamp`` raises ``OverflowError`` /
+            # ``OSError`` for values past the platform's representable
+            # range and ``ValueError`` for NaN. ``requested_at`` comes
+            # from a requester-controlled field, so the fallback closes
+            # the round-crash path on a malformed integer the same way
+            # the string branch handles a malformed ISO 8601 string.
+            try:
+                requested_at_iso = _iso_z(
+                    datetime.fromtimestamp(requested_at_raw, tz=timezone.utc)
+                )
+            except (ValueError, OverflowError, OSError):
+                self.context.logger.warning(
+                    "requested_at=%r for req_id=%s is out of range; "
+                    "falling back to executed_at.",
+                    requested_at_raw,
+                    req_id,
+                )
+                requested_at_iso = executed_at_iso
         elif isinstance(requested_at_raw, str) and requested_at_raw:
             # Parse the requester-provided ISO 8601 string and re-emit via
             # ``_iso_z`` so a non-UTC offset (``…+05:30``, ``…-08:00``) gets
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -97,6 +97,36 @@ RESPONSE_SCHEMA_VERSION = "2.0"
 IPFS_MAX_TASK_BYTES = 1_048_576  # 1MB cap on attacker-controlled task payload
 MAX_PROMPT_BYTES = 100_000  # 100KB cap on the prompt field
 
+
+def _iso_z(dt: datetime) -> str:
+    """Render a timezone-aware datetime as ISO 8601 with the ``Z`` suffix.
+
+    Normalises through ``astimezone(timezone.utc)`` so a datetime with any
+    UTC offset (``+00:00``, ``+05:30``, ``-08:00``) emits the same canonical
+    string for the same wall-clock instant. The wildcard server's Pydantic
+    ``AwareDatetime`` round-trips through ``model_dump(mode="json")`` which
+    emits ``...Z`` for UTC; the mech's ``batch_hash`` recompute on the
+    server side compares byte-for-byte against ours, so leaving the offset
+    unconverted breaks the equality at the canonical-JSON layer (this was
+    the original ``+00:00`` → ``Z`` mismatch that dropped every offchain
+    delivery with ``400 BATCH_HASH_MISMATCH``).
+
+    Naive datetimes are rejected. Python 3.6+'s ``astimezone`` silently
+    coerces a naive value to the system's local timezone, which would
+    swap the canonical byte string between dev machines and CI without a
+    clear failure signal. The caller must attach a tz explicitly (always
+    ``timezone.utc`` in this module).
+
+    :param dt: a timezone-aware ``datetime``.
+    :return: ISO 8601 string of the same instant in UTC, suffixed with ``Z``.
+    :raises ValueError: if ``dt.tzinfo`` is None (naive datetime).
+    """
+    if dt.tzinfo is None:
+        raise ValueError(
+            "_iso_z requires a timezone-aware datetime; got a naive value"
+        )
+    return dt.astimezone(timezone.utc).isoformat().replace("+00:00", "Z")
+
 # Cap on the JSON-serialised size of each ``raw_content`` blob attached to
 # a wildcard event. The blob rides Tendermint consensus replication into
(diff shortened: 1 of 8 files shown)


packages/valory/skills/task_execution/behaviours.py
```diff
@@ -733,6 +734,29 @@ def _execute_task(self) -> None:
...
         self._request_handling_deadline = None
+
+        # Off-chain path: prompt JSON is already inline in ``task_data["ipfs_data"]``
+        # (set by the handler in ``_enqueue_offchain_request`` and mirrored to
+        # ``in_memory_requests``). Skip the IPFS GET round-trip entirely and feed
+        # a synthetic message straight to ``_handle_get_task`` so the rest of the
+        # pipeline (validation + pebble dispatch) is unchanged. Matches the design
+        # intent of #457 ("off-chain path skips the IPFS upload"); the GET half
+        # was left in place by oversight.
+        if task_data.get("is_offchain"):
```
OjusWiZard, line 774: Logic is sound — verified `_handle_get_task` only reads `message.files.values()` and never uses the `dialogue` arg, so the `SimpleNamespace` + `cast(Any, None)` is safe, and `_request_handling_deadline` is reset to `None` just above so the deadline guard no-ops correctly.

**Important (non-blocking): no test covers this branch.** Please add a unit test asserting that on `is_offchain` with inline `ipfs_data`, `_execute_task` dispatches through `_handle_get_task` with **no** IPFS GET, and that a missing/non-str `ipfs_data` sets `_invalid_request = True`.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1614,11 +1646,16 @@ def _build_wildcard_event(
...
+                datetime.fromtimestamp(requested_at_raw, tz=timezone.utc)
+                .isoformat()
+                .replace("+00:00", "Z")
+            )
         elif isinstance(requested_at_raw, str) and requested_at_raw:
-            requested_at_iso = requested_at_raw
+            # Normalise an explicit ``+00:00`` to ``Z`` so a requester-
+            # provided ISO 8601 string canonicalises the same way the
+            # server's Pydantic round-trip emits it.
+            requested_at_iso = requested_at_raw.replace("+00:00", "Z")
```
OjusWiZard: **Important (non-blocking): add the regression test the PR's own test plan flags.** `_build_wildcard_event` has zero existing coverage — assert all three of `requested_at`/`executed_at`/`delivered_at` end in `Z` (not `+00:00`), for both the unix-timestamp and string-`requested_at` inputs.

**Advisory:** this `.replace("+00:00", "Z")` only canonicalizes UTC. A requester-supplied `requested_at` with a non-UTC offset (e.g. `…+05:30`) passes through unchanged and could reintroduce a narrower `batch_hash` mismatch if the server normalizes to UTC. If requesters can set arbitrary offsets, consider `datetime.fromisoformat(...).astimezone(timezone.utc)` instead. Likely out of scope here — just flagging.

**Nit:** the `.isoformat().replace("+00:00", "Z")` idiom appears at three sites; a small `_iso_z(dt)` helper would DRY it up.

OjusWiZard (dismissed):
Approving. Both fixes are correct and well-reasoned, the comments are accurate, and the offchain synthetic-message contract holds (`_handle_get_task` only reads `message.files` and never touches the `dialogue` arg). No correctness, security, or data-loss issues.

One thing I'd genuinely like before/just-after merge: a regression test for the `Z`-suffix normalization. That bug was production-silent (dropped every offchain delivery at the final auth step with `400 BATCH_HASH_MISMATCH`), and `_build_wildcard_event` currently has no test coverage at all. This PR adds no test files. See inline notes.

DIvyaNautiyal07:
Thanks for the review. All four points addressed in 91da320 — full breakdown:

**1. Regression test for the offchain-skip branch (line 745).** Added `test_execute_task_offchain_dispatches_via_handle_get_task_no_ipfs_get` and `test_execute_task_offchain_missing_ipfs_data_sets_invalid_request`. The first asserts that on `is_offchain` + inline `ipfs_data`, `_handle_get_task` is called exactly once with the synthetic `SimpleNamespace(files={"metadata.json": <payload>})` and `send_message` is never called (no IPFS GET). The second asserts that a missing/non-str `ipfs_data` sets `_invalid_request = True` and neither dispatch path runs.

**2. Regression test for the Z-suffix fix (line 1658).** Added four tests covering all three timestamp fields (`requested_at` / `executed_at` / `delivered_at`) across the three input shapes: ISO 8601 string, Unix int, and malformed-string fallback. The happy-path test asserts every timestamp ends in `Z` and never contains `+00:00`.

**3. Advisory: non-UTC offset (e.g. `…+05:30`).** Good catch — applied. Replaced the `requested_at` string path with `_iso_z(datetime.fromisoformat(value))` so any offset rotates through UTC first. Added `test_build_wildcard_event_normalises_non_utc_requested_at` to pin it (uses `18:49:06+05:30` and asserts the output is `13:19:06Z`). Malformed ISO strings fall through to `executed_at` with a warning instead of crashing the FSM round.

**4. Nit: DRY the `.replace("+00:00", "Z")` idiom.** Introduced module-level `_iso_z(dt)` and routed all five sites through it (`_handle_done_task` plus the four in `_build_wildcard_event`). The helper rejects naive datetimes explicitly so a future caller that drops the explicit tz can't silently coerce to local system time (Python's `astimezone` would otherwise localise quietly — that's the trap behind the `pytest.raises` in `test_iso_z_naive_datetime_raises`).

129 / 129 tests pass (9 new + 120 existing).

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1614,11 +1669,27 @@ def _build_wildcard_event(
...
+            # ``_iso_z`` so a non-UTC offset (``…+05:30``, ``…-08:00``) gets
+            # converted to the canonical UTC ``Z`` form. A plain
+            # ``.replace("+00:00", "Z")`` would leave non-UTC offsets
+            # untouched and reintroduce the ``batch_hash`` mismatch the
+            # canonicalisation is meant to close. Malformed strings fall
+            # through to ``now_iso`` rather than crashing the FSM round.
+            try:
+                requested_at_iso = _iso_z(datetime.fromisoformat(requested_at_raw))
+            except ValueError:
+                self.context.logger.warning(
```
bennyjo, line 1727: `datetime.fromisoformat` rejects a trailing `Z` until Python 3.11, but `pyproject.toml` still allows 3.10. So on 3.10 a `requested_at` in the canonical `…Z` form (what the server emits, what JS `toISOString()` produces) raises `ValueError` and silently falls back to `executed_at` — no crash, but the real requested time is lost. Either confirm deployments are ≥3.11, or `replace("Z", "+00:00")` before parsing. The tests cover `+00:00`/`+05:30`/int/garbage but not the `Z` string — the one version-dependent case.

bennyjo (commented):
One brittleness + test-gap note on the requested_at timestamp parsing.

DIvyaNautiyal07:
Thanks @bennyjo, addressed in f24251c9. The string-parse path now does \`requested_at_raw.replace(\"Z\", \"+00:00\")\` before \`datetime.fromisoformat\` so 3.10 and 3.11+ behave identically; comment updated to call out the version-dependence so a future refactor doesn't drop the rewrite. Added \`test_build_wildcard_event_z_suffix_requested_at_is_parsed\` pinning the parse on the trailing-\`Z\` input.

OjusWiZard (dismissed):
Re-approving after reviewing the follow-up commits. All prior feedback is addressed with careful, well-tested code:

- **Non-UTC timestamps:** `requested_at` now routes through `datetime.fromisoformat(...) → _iso_z()` (`astimezone(utc)`), so non-UTC offsets like `+05:30` canonicalize to UTC `Z` — closes the `batch_hash` gap a bare `.replace` would leave open.
- **Python 3.10:** the `.replace("Z", "+00:00")` pre-step handles `fromisoformat` rejecting `…Z` on 3.10, still correct on 3.11+. Malformed strings are logged and fall back to `executed_at` — no silent swallow, no crash.
- **DRY:** the `_iso_z` helper consolidates all sites and rejects naive datetimes with a clear `ValueError`.
- **Tests:** new coverage is behavioral — the offchain test asserts both that `_handle_get_task` fires with the right synthetic payload AND that no IPFS GET is issued; the timestamp tests assert exact `Z` output across unix/offset/Z-suffix/malformed inputs.

No blocking issues. Remaining gate is CI `check-hash` on the fingerprint refresh.

bennyjo (commented):
One inline note on the requester-controlled timestamp parsing.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1614,11 +1669,37 @@ def _build_wildcard_event(
             "requested_at"
         )
         if isinstance(requested_at_raw, (int, float)):
-            requested_at_iso = datetime.fromtimestamp(
-                requested_at_raw, tz=timezone.utc
-            ).isoformat()
+            requested_at_iso = _iso_z(
+                datetime.fromtimestamp(requested_at_raw, tz=timezone.utc)
```
bennyjo: **Impact:** one offchain request can silently kill its own delivery. `requested_at` comes from the requester, and a wild value like `{"requested_at": 99999999999999}` makes `fromtimestamp` throw. Nothing catches it here or in `_handle_done_task`, so the task drops off the done list — no on-chain error, no signal.

The string branch just below already guards this and falls back to `executed_at`. Could you give this branch (and the `executed_at` one above) the same `try/except`? Catch `(ValueError, OverflowError, OSError)` — big values raise `OverflowError`, not `ValueError`. A test with an out-of-range value would pin the fallback.

DIvyaNautiyal07:
Addressed in 47502f4. Both Unix-int timestamp branches (\`requested_at\` requester-controlled, \`executed_at\` tool-controlled) now wrap \`fromtimestamp\` in \`try/except (ValueError, OverflowError, OSError)\` with the same fallback shape as the string branch. Added two regression tests using \`99_999_999_999_999\` as the wild value (year ~5138, well past the platform range) — \`test_build_wildcard_event_unix_requested_at_out_of_range_falls_back\` and \`test_build_wildcard_event_unix_executed_at_out_of_range_falls_back\`. Tests pin that the fallback timestamp still ends in \`Z\` so the canonical-JSON contract holds.

bennyjo (approved):
Approving. Both prior notes are addressed with tests: the Python-3.10 trailing-`Z` parse (`f24251c9`) and the Unix-int out-of-range guard (`47502f4`). Core fixes are correct and the offchain synthetic-message contract holds.

One non-blocking follow-up, same class as the bug `47502f4` just fixed: the string branch of `requested_at` (~behaviours.py:1701) still catches only `ValueError`, but it routes through `_iso_z` → `astimezone(utc)`, which raises `OverflowError` on a valid-but-extreme requester ISO string (e.g. `"9999-12-31T23:59:59-12:00"`, confirmed locally). That escapes the fallback and silently drops the delivery — exactly the failure mode the int branches now guard against. Suggest widening to `(ValueError, OverflowError, OSError)` for symmetry with the int branches, plus a regression test. Cheap to do as a follow-up; not gating the merge.

OjusWiZard (approved):

PR #460, opened 2026-06-30
feat(mech): on-chain write path — PostTxSettlement sweep + source field

DIvyaNautiyal07:
## Summary

- `_build_wildcard_event` sets the wire-level `source` to `mech_offchain` or `mech_onchain` based on the task's `is_offchain` flag, and the build gate relaxes to include on-chain marketplace tasks too.
- `filter_requests` stamps `enqueued_at_local` on every pending task; `PostTxSettlementBehaviour._sweep_pending_undelivered` walks the queue and emits request-only events (`response=None`, `source=mech_onchain`) for tasks past the configurable max age.
- `_extract_offchain_events` now keys on the presence of `wildcard_event`, so on-chain deliveries ride the same batch as off-chain.
- Two new default-off params: `mech_events_sweep_pending_enabled`, `mech_events_sweep_max_age_seconds`. Sweep is gated independently from the delivered-event write so operators can stage rollout.
- 597 tests pass locally (existing + new).

PRs land in order: predict-api#165 (the migration + model + writer) first, then this. Reversing the order would make request-only POSTs return 422 until the server catches up.

## Why local-clock sweep, not `getRequestStatus`

The contract's `RequestInfo.responseTimeout` is authoritative but reading it per sweep is one RPC per pending task. We explicitly avoid that: `filter_requests` stamps `enqueued_at_local` at enqueue time, and the sweep uses a configurable local max age (default 1 hour). If step-in delivers between the sweep and the predict-api write, the DB's `ON CONFLICT (request_id) DO UPDATE ... WHERE delivery_mech IS NULL` clause reconciles to the correct row. See `autonolas-marketplace/docs/onchain_write_path_scope.md` §3.2 and §3.4 for the trade-off.

## Notes for review

- Request-only events use placeholder `prompt='[onchain undelivered]'` and `tool='unknown'`. The IPFS body for the request isn't resolved at sweep time (would require a blocking fetch per task); the analytics-side undelivered signal is the absent `mech_responses` row, not the prompt content.
- The sweep operates in-place on `shared_state[PENDING_TASKS]` so other consumers (the task picker in `task_execution`) see the removal.
- Sweep is feature-flag gated separately (`mech_events_sweep_pending_enabled`); default off until v1 is validated against a live deployment. Operators can flip `mech_events_enabled` first to write delivered on-chain rows, then flip the sweep on once that's stable.
- PR targets `offchain-epic` per the team's branching request, not `main`.

## Test plan

- [x] `uv run pytest packages/valory/skills/task_submission_abci/tests/ packages/valory/skills/task_execution/tests/` — 597 passed locally.
- [ ] `autonomy packages lock --check` — verify the hash updates I generated round-trip cleanly.
- [ ] Cross-repo manual run of `autonolas-marketplace/docs/onchain_write_path_e2e.md` after predict-api#165 lands.

### DIFF
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1545,7 +1545,17 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # built and nothing rides Tendermint consensus replication. The flag
         # check lives here as well as in the post-settlement behaviour so
         # the consensus-state cost is gated, not just the HTTP write.
-        if is_offchain and getattr(self.params, "mech_events_enabled", False):
+        # Build the wildcard event for both off-chain HTTP deliveries
+        # (is_offchain=True) and on-chain marketplace deliveries
+        # (is_offchain=False AND is_marketplace_mech). The ``source`` field
+        # on the event (set by ``_build_wildcard_event``) is what
+        # disambiguates the two paths on the wildcard side: ``mech_offchain``
+        # for the paid HTTP path, ``mech_onchain`` for the on-chain rails.
+        # See ``autonolas-marketplace/docs/onchain_write_path_scope.md`` for
+        # the agreed shape.
+        wildcard_mode_enabled = getattr(self.params, "mech_events_enabled", False)
+        is_marketplace_delivery = bool(done_task.get("is_marketplace_mech"))
+        if wildcard_mode_enabled and (is_offchain or is_marketplace_delivery):
             done_task["wildcard_event"] = self._build_wildcard_event(
                 done_task=done_task,
                 cid=cid,
@@ -1643,6 +1653,17 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         tool = str(done_task.get("tool") or "unknown")
         delivery_mech = str(
             done_task.get("mech_address") or self.params.mech_marketplace_address or ""
+        )
+        # Wildcard ``source`` derives from the per-task ``is_offchain``
+        # flag: True → ``mech_offchain`` (paid HTTP path); False →
+        # ``mech_onchain`` (the on-chain marketplace rails). This is the
+        # only field that disambiguates the two paths on the wildcard
+        # side; the per-row ``is_offchain`` boolean stays on the
+        # response payload as well (the lake keeps both because the
+        # ipfs_historical backfill describes either path).
+        is_offchain_task = bool(executing_task.get("is_offchain", False))
+        wildcard_source = (
+            "mech_offchain" if is_offchain_task else "mech_onchain"
         )
 
         # `start_time` is a perf_counter() value (monotonic, not wall-clock),
@@ -1829,7 +1850,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 "error": error_text,
                 "executed_at": executed_at_iso,
                 "cost_dict": cost_dict,
-                "is_offchain": True,
+                "is_offchain": is_offchain_task,
                 "tool_hash": tool_hash,
                 "execution_latency_ms": None,
                 "params_used": (
@@ -1845,6 +1866,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 "response_cid": None,
                 "delivered_at": now_iso,
             },
+            "source": wildcard_source,
         }
 
     def _reset_executing_task(self) -> None:
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1662,9 +1662,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # response payload as well (the lake keeps both because the
         # ipfs_historical backfill describes either path).
         is_offchain_task = bool(executing_task.get("is_offchain", False))
-        wildcard_source = (
-            "mech_offchain" if is_offchain_task else "mech_onchain"
-        )
+        wildcard_source = "mech_offchain" if is_offchain_task else "mech_onchain"
 
         # `start_time` is a perf_counter() value (monotonic, not wall-clock),
         # not suitable as a delivered_at timestamp. Use the current wall-clock
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1813,7 +1813,21 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 "marketplace_address": str(
                     getattr(self.params, "mech_marketplace_address", "") or ""
                 ),
-                "requester": str(executing_task.get("sender") or ""),
+                # ``sender`` is the off-chain HTTP path's requester key
+                # (set from the signed request body in
+                # ``task_execution.handlers._enqueue_offchain_request``).
+                # Marketplace pending tasks come from
+                # ``MechMarketplaceContract.get_marketplace_undelivered_reqs``
+                # which keys the requester as ``requester`` and carries no
+                # ``sender`` field. This block now runs for on-chain
+                # marketplace deliveries too, so falling back to
+                # ``requester`` keeps the wildcard row correctly populated
+                # instead of writing an empty ``requester``.
+                "requester": str(
+                    executing_task.get("sender")
+                    or executing_task.get("requester")
+                    or ""
+                ),
                 "priority_mech": str(
                     executing_task.get("priority_mech") or delivery_mech
                 ),
--- a/packages/valory/skills/task_execution/tests/test_behaviours.py
+++ b/packages/valory/skills/task_execution/tests/test_behaviours.py
@@ -3380,3 +3380,71 @@ def test_execute_task_offchain_missing_ipfs_data_sets_invalid_request(
     assert behaviour._invalid_request is True
     assert handle_calls == []
     assert send_calls == []
+
+
+# ---------------------------------------------------------------------------
+# On-chain write path: source field on _build_wildcard_event
+# ---------------------------------------------------------------------------
+#
+# After the on-chain write path lands, _build_wildcard_event sets a
+# top-level ``source`` field on the returned event dict based on the
+# task's ``is_offchain`` flag:
+#   - is_offchain=True  → source="mech_offchain" (the paid HTTP path).
+#   - is_offchain=False → source="mech_onchain"  (the on-chain rails).
+# The wildcard server uses this field to tag both mech_requests and
+# mech_responses rows for the analytics ETL. See
+# ``autonolas-marketplace/docs/onchain_write_path_scope.md`` §3.
+
+
+def test_build_wildcard_event_offchain_task_carries_mech_offchain_source(
+    behaviour: Any,
+    params_stub: Any,
+    shared_state: Dict[str, Any],
+) -> None:
+    """A task with ``is_offchain=True`` produces ``source='mech_offchain'``."""
+    request_data = {
+        "prompt": "p",
+        "tool": "prediction-offline",
+        "requested_at": "2026-06-25T13:19:06.651013Z",
+    }
+    done_task, executing_task = _wildcard_event_setup(
+        behaviour, shared_state, request_data
+    )
+    event = behaviour._build_wildcard_event(
+        done_task=done_task, cid="bafy", executing_task=executing_task
+    )
+    assert event["source"] == "mech_offchain"
+    # Response payload's is_offchain field still reflects the underlying
+    # delivery path (kept in lock-step with the source for clarity).
+    assert event["response"]["is_offchain"] is True
+
+
+def test_build_wildcard_event_onchain_task_carries_mech_onchain_source(
+    behaviour: Any,
+    params_stub: Any,
+    shared_state: Dict[str, Any],
+) -> None:
+    """A task with ``is_offchain=False`` produces ``source='mech_onchain'``.
+
+    The on-chain marketplace path lands here when the mech delivers an
+    on-chain request via ``deliverMarketplaceWithSignatures``. The
+    per-row response.is_offchain flag also flips false to keep the
+    analytics lake's two signals (which path + which source) consistent.
+    """
+    request_data = {
+        "prompt": "p",
+        "tool": "prediction-offline",
+        "requested_at": "2026-06-25T13:19:06.651013Z",
+    }
+    done_task, executing_task = _wildcard_event_setup(
+        behaviour, shared_state, request_data
+    )
+    # Flip the flags on both halves: the source decision keys on the
+    # executing_task; the response payload's is_offchain mirrors it.
+    executing_task["is_offchain"] = False
+    done_task["is_offchain"] = False
+    event = behaviour._build_wildcard_event(
+        done_task=done_task, cid="bafy", executing_task=executing_task
+    )
+    assert event["source"] == "mech_onchain"
+    assert event["response"]["is_offchain"] is False
--- a/packages/valory/skills/task_execution/tests/test_behaviours.py
+++ b/packages/valory/skills/task_execution/tests/test_behaviours.py
@@ -3454,3 +3454,49 @@ def test_build_wildcard_event_onchain_task_carries_mech_onchain_source(
     )
     assert event["source"] == "mech_onchain"
     assert event["response"]["is_offchain"] is False
+
+
+def test_build_wildcard_event_marketplace_task_falls_back_to_requester_key(
+    behaviour: Any,
+    params_stub: Any,
+    shared_state: Dict[str, Any],
+) -> None:
+    """Marketplace-delivered tasks carry `requester` (not `sender`); requester still populates.
+
+    Pending marketplace tasks come from
+    ``MechMarketplaceContract.get_marketplace_undelivered_reqs``, which
+    keys the requester as ``requester`` and carries no ``sender`` field.
+    The delivered wildcard row must still populate ``request.requester``
+    or the analytics-side join breaks.
+
+    The shared ``_wildcard_event_setup`` fixture always stamps
+    ``sender`` on ``executing_task`` (that's the off-chain HTTP shape).
+    This test deletes it and injects ``requester`` alone, exercising the
+    ``sender or requester`` fallback added in ``_build_wildcard_event``.
+
+    :param behaviour: task_execution behaviour under test.
+    :param params_stub: params fixture — consumed by
+        ``_wildcard_event_setup`` to populate ``self.params`` bits.
+    :param shared_state: shared_state fixture used by
+        ``_wildcard_event_setup`` for ``IN_MEMORY_REQUESTS`` /
+        ``OFFCHAIN_REQUEST_RESPONSES`` seeding.
+    """
+    request_data = {
+        "prompt": "on-chain marketplace delivery",
+        "tool": "prediction-offline",
+        "requested_at": "2026-06-25T13:19:06.651013Z",
+    }
+    done_task, executing_task = _wildcard_event_setup(
+        behaviour, shared_state, request_data
+    )
+    # Marketplace shape: only ``requester``, no ``sender``. Also flip the
+    # per-row provenance so this is the on-chain path.
+    del executing_task["sender"]
+    executing_task["requester"] = "0xMARKETPLACE_REQUESTER"
+    executing_task["is_offchain"] = False
+    done_task["is_offchain"] = False
+    event = behaviour._build_wildcard_event(
+        done_task=done_task, cid="bafy", executing_task=executing_task
+    )
+    assert event["request"]["requester"] == "0xMARKETPLACE_REQUESTER"
+    assert event["source"] == "mech_onchain"
--- a/packages/valory/skills/task_execution/tests/test_behaviours.py
+++ b/packages/valory/skills/task_execution/tests/test_behaviours.py
@@ -3430,6 +3430,12 @@ def test_build_wildcard_event_onchain_task_carries_mech_onchain_source(
     on-chain request via ``deliverMarketplaceWithSignatures``. The
     per-row response.is_offchain flag also flips false to keep the
     analytics lake's two signals (which path + which source) consistent.
+
+    :param behaviour: the wired ``task_execution`` behaviour under test.
+    :param params_stub: params fixture — indirectly consumed by
+        ``_wildcard_event_setup`` to populate ``self.params`` bits.
+    :param shared_state: shared_state fixture used by
+        ``_wildcard_event_setup`` to build ``done_task`` / ``executing_task``.
     """
     request_data = {
         "prompt": "p",
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -1866,7 +1866,8 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
             1. The feature flag is off (the default during Phase 1).
             2. The skill has no wildcard URL configured.
             3. The round's done_tasks carry no offchain wildcard_event
-               entries (the on-chain path doesn't populate these).
+               entries and the pending-task sweep produced no
+               request-only events (a truly quiet round).
 
         :yield: AEA protocol messages (signing dialogue, HTTP request) via
             the underlying generator-based helpers; this method does not
@@ -1887,7 +1888,12 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
         # have sat in the queue past the operator's max age. Appended to
         # the same batch as delivered events so the signing round-trip
         # costs one signature per settlement, not one per event class.
-        request_only_events = self._sweep_pending_undelivered()
+        # ``_sweep_pending_undelivered`` returns the swept ``request_id``s
+        # alongside the events but does NOT mutate the pending queue: any
+        # guard-skip or POST failure below would otherwise drop the tasks
+        # with no re-queue and no written row. The queue is only mutated
+        # once the wildcard POST has confirmed a 2xx below.
+        request_only_events, swept_request_ids = self._sweep_pending_undelivered()
         if request_only_events:
             events = events + request_only_events
         if not events:
@@ -2064,6 +2070,15 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
             len(events),
             self.context.agent_address,
         )
+        # Wildcard confirmed the batch; drop the swept undelivered tasks
+        # from the pending queue now. Deferring to this point (rather than
+        # inside ``_sweep_pending_undelivered``) ensures that any early
+        # return above — missing mech address, missing marketplace, mixed
+        # chains, unconfigured chain_id, digest build, signing, JSON
+        # serialisation, POST timeout, or non-2xx status — leaves the
+        # sweep at-least-once: the tasks stay on the queue and the next
+        # FSM cycle re-attempts the write.
+        self._drop_swept_from_pending(swept_request_ids)
 
     def _extract_offchain_events(self) -> List[Dict[str, Any]]:
         """Return the ``wildcard_event`` payload from every done_task that
@@ -2090,11 +2105,26 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
             events.append(event)
         return events
 
-    def _sweep_pending_undelivered(self) -> List[Dict[str, Any]]:
-        """Walk the on-chain pending-tasks queue, emit request-only
+    def _sweep_pending_undelivered(
+        self,
+    ) -> Tuple[List[Dict[str, Any]], List[str]]:
+        """Walk the on-chain pending-tasks queue and return request-only
         events for any task that has sat in the queue past the
-        operator-configured sweep window, and remove those tasks from
-        the queue.
+        operator-configured sweep window, alongside the ``request_id``s
+        of the swept tasks so the caller can remove them from the queue
+        AFTER the wildcard POST confirms success.
+
+        The sweep does NOT mutate ``PENDING_TASKS``. Dropping tasks here
+        (as an earlier revision of this PR did) would lose the
+        undelivered signal on any of the six early-return paths below
+        the caller — missing mech address, missing marketplace, mixed
+        chains, unconfigured chain id, digest-build failure, or the POST
+        itself failing — because ``PENDING_TASKS`` is the only record
+        the mech has for a task that never got delivered (unlike a
+        delivered event, which is anchored on-chain). Deferring the
+        queue mutation to a confirmed 2xx keeps the sweep at-least-once
+        rather than at-most-once and lets a transient POST failure
+        recover on the next FSM cycle.
 
         The sweep is the mech side of the on-chain write path's
         undelivered-request capture. Local-clock comparison only: the
@@ -2116,63 +2146,83 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
         request-only event always carries ``source='mech_onchain'``;
         the off-chain path doesn't produce them by construction.
 
-        :return: an ordered list of request-only event dicts ready to
-            append to the batch.
+        :return: a ``(events, swept_request_ids)`` tuple. ``events`` is
+            an ordered list of request-only event dicts ready to append
+            to the batch; ``swept_request_ids`` is the ordered list of
+            ``request_id`` strings the caller should drop from
+            ``PENDING_TASKS`` if the wildcard POST returns 2xx. Empty
+            lists on any short-circuit or when nothing timed out.
         """
         if not getattr(self.params, "mech_events_enabled", False):
-            return []
+            return [], []
         if not getattr(self.params, "mech_events_sweep_pending_enabled", False):
-            return []
+            return [], []
         max_age = float(
             getattr(self.params, "mech_events_sweep_max_age_seconds", 3600.0)
         )
         if max_age <= 0:
-            return []
+            return [], []
         pending = self.context.shared_state.get(PENDING_TASKS)
         if not isinstance(pending, list) or not pending:
-            return []
+            return [], []
 
         now_ts = time.time()
-        kept: List[Dict[str, Any]] = []
         request_only_events: List[Dict[str, Any]] = []
+        swept_request_ids: List[str] = []
         for task in pending:
             if not isinstance(task, dict):
                 # Anything that isn't a dict shouldn't be in
-                # pending_tasks. Keep it as-is rather than dropping —
-                # whatever bug put it there is not this sweep's
-                # problem.
-                kept.append(task)
+                # pending_tasks. Leave it alone — whatever bug put it
+                # there is not this sweep's problem.
                 continue
             enqueued_at = task.get("enqueued_at_local")
             if not isinstance(enqueued_at, (int, float)):
                 # Older entries (pre-sweep deployment) won't carry the
                 # stamp. Leave them alone; they'll be picked up on the
                 # next enqueue or when the handler stamps fresh.
-                kept.append(task)
                 continue
             if now_ts - float(enqueued_at) < max_age:
-                kept.append(task)
                 continue
             event = self._build_request_only_event(task, now_ts)
             if event is None:
                 # If we can't build a valid event from a stale task,
-                # keep it on the queue rather than silently dropping —
+                # leave it on the queue rather than silently dropping —
                 # better a re-attempted sweep than a lost write.
-                kept.append(task)
                 continue
             request_only_events.append(event)
+            swept_request_ids.append(str(event["request"].get("request_id")))
             self.context.logger.info(
-                "Wildcard sweep: emitting request-only event for "
-                "timed-out task request_id=%s, age=%.0fs",
+                "Wildcard sweep: staging request-only event for "
+                "timed-out task request_id=%s, age=%.0fs "
+                "(queue not mutated until POST succeeds)",
                 event["request"].get("request_id"),
                 now_ts - float(enqueued_at),
             )
 
-        # Mutate the shared list in place: other consumers (the task
-        # picker in task_execution) hold the same reference and would
-        # not see a rebinding.
-        pending[:] = kept
-        return request_only_events
+        return request_only_events, swept_request_ids
+
+    def _drop_swept_from_pending(self, swept_request_ids: List[str]) -> None:
+        """Remove the swept tasks from ``PENDING_TASKS`` after a
+        confirmed wildcard write.
+
+        Mutates the shared list in place: other consumers (the task
+        picker in task_execution) hold the same reference and would not
+        see a rebinding.
+        """
+        if not swept_request_ids:
+            return
+        pending = self.context.shared_state.get(PENDING_TASKS)
+        if not isinstance(pending, list) or not pending:
+            return
+        drop_set = set(swept_request_ids)
+        pending[:] = [
+            task
+            for task in pending
+            if not (
+                isinstance(task, dict)
+                and str(task.get("requestId") or "") in drop_set
+            )
+        ]
 
     def _build_request_only_event(
         self, task: Dict[str, Any], now_ts: float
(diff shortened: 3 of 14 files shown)


packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2051,27 +2066,240 @@ def _do_wildcard_write_best_effort(self) -> Generator:
...
+        )
+
+        # The wildcard server requires ``prompt`` to be a non-empty
+        # string and ``tool`` to be a non-empty string. Placeholders
+        # match the off-chain build's pattern ("[offchain request]")
+        # so a reader can spot a sweep-emitted row from the value.
+        prompt = "[onchain undelivered]"
+        tool = "unknown"
+        chain_id = int(
+            getattr(self.params, "mech_events_chain_id", 0) or 0
```
OjusWiZard: 🔴 **Critical — sweep events get `chain_id=0`, so every sweep batch is dropped (data loss).**

`mech_events_chain_id` is declared only on `task_execution`'s `Params`, **not** on `task_submission_abci.models.Params` (this PR added the two sweep params there but not this one). So this `getattr` falls back to `0` and every request-only event carries `chain_id=0`. Then `_sign_and_post_events` drops it:
- pure-sweep round → `chain_id == 0` guard (~L1949) → batch skipped;
- mixed round → delivered events carry the real chain id, request-only events carry `0` → the heterogeneous-chain guard (~L1933) drops the **whole batch, including the valid delivered events**.

And `_sweep_pending_undelivered` has already removed the tasks from `PENDING_TASKS` (`pending[:] = kept`) before the post runs, so the undelivered signal is lost with no re-queue. This is masked by tests that stub `Params` as a `SimpleNamespace` with `mech_events_chain_id` injected.

**Fix:** add `self.mech_events_chain_id = int(kwargs.get("mech_events_chain_id", 0) or 0)` to `task_submission_abci/models.py`, and mirror `mech_events_chain_id: 0` into that `skill.yaml` (matching `task_execution`).

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1644,6 +1654,17 @@ def _build_wildcard_event(
...
         )
+        # Wildcard ``source`` derives from the per-task ``is_offchain``
+        # flag: True → ``mech_offchain`` (paid HTTP path); False →
+        # ``mech_onchain`` (the on-chain marketplace rails). This is the
+        # only field that disambiguates the two paths on the wildcard
+        # side; the per-row ``is_offchain`` boolean stays on the
+        # response payload as well (the lake keeps both because the
+        # ipfs_historical backfill describes either path).
+        is_offchain_task = bool(executing_task.get("is_offchain", False))
+        wildcard_source = (
```
OjusWiZard: 🟠 **Important — delivered marketplace events get an empty `requester`.**

This block now runs for on-chain marketplace deliveries too. But a few lines down (L1816) the requester is read as `executing_task.get("sender")`. Marketplace pending tasks come from `MechMarketplaceContract.get_marketplace_undelivered_reqs`, which keys the requester as **`requester`** and carries **no `sender`** — so every delivered marketplace wildcard row is written with `requester=""`. Off-chain is unaffected (its dict comes from the HTTP body, which does have `sender`).

Note the sweep's `_build_request_only_event` correctly uses `requester` one method over — only this delivered path has the mismatch.

**Fix at L1816:** `str(executing_task.get("sender") or executing_task.get("requester") or "")`. No delivered-marketplace test currently covers this path.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2051,27 +2066,240 @@ def _do_wildcard_write_best_effort(self) -> Generator:
...
+                "Wildcard sweep: emitting request-only event for "
+                "timed-out task request_id=%s, age=%.0fs",
+                event["request"].get("request_id"),
+                now_ts - float(enqueued_at),
+            )
+
+        # Mutate the shared list in place: other consumers (the task
+        # picker in task_execution) hold the same reference and would
+        # not see a rebinding.
+        pending[:] = kept
```
OjusWiZard: 🟡 **Optimistic removal before durable write.**

`pending[:] = kept` drops the swept tasks here, before signing + the HTTP POST. Any guard-skip (see the `chain_id=0` issue) or POST failure then loses the undelivered signal permanently — there's no re-queue/retry, and `PENDING_TASKS` is the only record for these tasks (unlike delivered events, whose source of truth is on-chain).

Consider removing tasks from the queue only after a confirmed write, or re-queueing them on a failed/skipped post.

OjusWiZard (commented):
Multi-perspective review (automated, then verified against the pinned `mech_marketplace` contract). Two genuine pre-merge defects — both silent (no crash; wrong/no data into the lake) and both masked by stubbed tests. Details inline.

**Severity map**
- 🔴 Critical: sweep events built with `chain_id=0` → batches dropped by the chain-id guards, and the swept tasks are already removed from `PENDING_TASKS` → permanent loss of the undelivered signal.
- 🟠 Important: delivered marketplace events written with empty `requester` (`sender` vs `requester` key).
- 🟡 Optimistic removal before durable write in the sweep.

**Test gaps that let these through**
- No test exercises the marketplace **delivered** path (`is_offchain=False, is_marketplace_mech=True`) through `_build_wildcard_event` — all delivered fixtures are off-chain (`sender`). Would catch the empty-requester bug.
- Sweep tests stub `Params` as a `SimpleNamespace` with `mech_events_chain_id` injected, never constructing the real `task_submission_abci.Params` — masks the `chain_id=0` bug. (For the record, the fixtures' `requester` key is **correct** for marketplace tasks.)

**Advisory**
- `_build_request_only_event` returning `None` re-queues + re-logs a WARNING every settlement round with no attempt cap (log-spam / queue growth for malformed tasks).
- The sweep avoids mislabeling off-chain tasks as `mech_onchain` only because off-chain enqueue happens not to stamp `enqueued_at_local`; an explicit `if task.get("is_offchain"): continue` would be safer defense-in-depth.

The sweep's field keys, the in-place `pending[:] = kept` mutation, and the gating logic are otherwise correct.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2051,27 +2066,240 @@ def _do_wildcard_write_best_effort(self) -> Generator:
...
+        """
+        request_id = task.get("requestId")
+        if not request_id:
+            self.context.logger.warning(
+                "Pending sweep: skipping task with no requestId: %s", task
+            )
+            return None
+        request_id_str = str(request_id)
+        priority_mech = str(task.get("priorityMech") or "")
+        requester = str(task.get("requester") or "")
```
bennyjo, line 2458: Wrong key: requesters are stored under `sender`, not `requester`. The working delivered path proves it (`task_execution/behaviours.py:1816` reads `sender`), and the executing task is the pending task unchanged. So `task.get("requester")` is always empty, the `return None` guard below fires, and the sweep emits nothing.

**Impact:** the whole feature is dead in production — no request-only events ever get written.
**Likelihood:** 100% — every real on-chain task.

Fix: `str(task.get("sender") or "")`. The tests miss it because `_make_pending_task` invents a `requester` key real events don't have.

packages/valory/skills/task_submission_abci/behaviours.py
bennyjo: `mech_events_chain_id` doesn't exist on this skill — it's only on task_execution (`models.py:126`, `skill.yaml:116`). So this always reads 0.

**Impact:** request-only-only batches get skipped at line 1947; mixed batches trip the multi-chain guard at line 1933 and drop the valid delivered events too.
**Likelihood:** 100% once the requester-key issue is fixed.

Fix: add the param to this skill, or take chain_id from the delivered events instead of params.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1876,10 +1883,18 @@ def _do_wildcard_write_best_effort(self) -> Generator:
             return
 
         events = self._extract_offchain_events()
+        # On-chain pending-task sweep: request-only events for tasks that
+        # have sat in the queue past the operator's max age. Appended to
+        # the same batch as delivered events so the signing round-trip
+        # costs one signature per settlement, not one per event class.
+        request_only_events = self._sweep_pending_undelivered()
```
bennyjo: The sweep deletes tasks from the queue (`shared_state["pending_tasks"]`, via `pending[:] = kept` at line 2174) before the write is confirmed. Six paths below return early without writing: missing mech address (1913), missing marketplace (1920), multi-chain (1933), chain_id 0 (1947), digest failure (1971), and the POST itself failing.

**Impact:** on any of those, the task is gone from the delivery queue (the task picker pops from the same list) and no row is written — lost for both delivery and analytics.
**Likelihood:** high during rollout — chain_id 0 and missing config are the normal half-configured states.

Suggest removing from the queue only after a successful write. Fine if dropping timed-out tasks is intended — flagging because a failed best-effort write shouldn't also discard a deliverable task.

bennyjo (commented):
On-chain write path: three blocking issues — requester key, chain_id on this skill, and queue mutation before the write. Details inline.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 14d706ae. Added `mech_events_chain_id` (default 0) to `task_submission_abci.models.Params` and mirrored the field into both `task_submission_abci/skill.yaml` and `mech_abci/skill.yaml`. Pure-sweep rounds now carry the real chain id; the batch-level guard only fires when the operator hasn't configured it, which is the intended behaviour. Test fixture already stubbed the field on `params`; that ambush surfaced this exact bug in production, so I left the fixture as-is and added the missing real-code side. Package hashes may need regenerating via `autonomy packages lock` after this lands; CI will surface it.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in 14d706ae. Changed `str(executing_task.get("sender") or "")` to `str(executing_task.get("sender") or executing_task.get("requester") or "")`. Off-chain path still finds `sender` first (the HTTP handler stamps that key on the enqueued dict); marketplace-delivered path falls back to `requester` (the key `get_marketplace_undelivered_reqs` sets on the pending-task dict at contract.py:297). No delivered-marketplace test currently covers this — flagged as a gap; the off-chain unit test doesn't stress it because it starts from an HTTP body that includes `sender`.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 14d706ae. `_sweep_pending_undelivered` now returns `(events, swept_request_ids)` and does NOT mutate `PENDING_TASKS`. A new `_drop_swept_from_pending` helper mutates the queue in place, and `_do_wildcard_write_best_effort` only calls it after a confirmed wildcard 2xx. All six early-return paths above the POST (missing mech address, missing marketplace, mixed chains, unconfigured chain_id, digest failure, POST failure) now leave the swept tasks on the queue and the next FSM cycle retries. Added `test_drop_swept_from_pending_*` cases; updated every existing sweep test to unpack the new tuple and assert the queue is left untouched by the sweep itself.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Checked this against the actual contract wire — pending marketplace tasks DO have a `requester` key (`contract.py:292-304` in `get_marketplace_undelivered_reqs` sets `"requester": req.get("requester")` from the on-chain `RequestForMarketplace` event). The off-chain HTTP handler stamps `sender` (from the signed body), and its enqueued dict flows through `filter_requests` → `pending_tasks`, but the sweep only runs against marketplace tasks by construction (an off-chain task without `priorityMech` and `requester` returns `None` from `_build_request_only_event` at the sanity check just above this line).\n\nSo the sweep's `task.get("requester")` on line 2213 is correct as-is. What was NOT correct — and thanks for pushing on this because it surfaced the delivered-side bug — was the DELIVERED path in `task_execution/behaviours.py:1816` which read `sender` and got empty for marketplace deliveries. Fixed that separately in 14d706ae with a `sender or requester` fallback (see OjusWiZard's thread above). Feature isn't dead in production once the requester fallback + `mech_events_chain_id` param are both landed.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 14d706ae. Added `mech_events_chain_id` (default 0) to `task_submission_abci.models.Params` and to both `task_submission_abci/skill.yaml` and `mech_abci/skill.yaml`. Went with the params-fix route (rather than reading chain_id from delivered events) because a pure-sweep round has no delivered event to fall back to, and the existing behaviour already assumes a single chain_id per Safe by construction.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 14d706ae. Queue mutation is now deferred to a confirmed POST 2xx via the new `_drop_swept_from_pending` helper. All six early-return paths you listed (missing mech address, missing marketplace, multi-chain, chain_id 0, digest failure, POST failure) now leave the swept tasks on the queue for retry on the next FSM cycle. The sweep itself is now pure — returns `(events, swept_request_ids)` and doesn't touch `PENDING_TASKS`. Fine to drop tasks that would fail POST every time is NOT the intent — the design (§3.2 of the scope doc) treats a failed best-effort write as "try again next cycle", not as "forget".

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -3380,3 +3380,77 @@ def test_execute_task_offchain_missing_ipfs_data_sets_invalid_request(
...
+        "prompt": "p",
+        "tool": "prediction-offline",
+        "requested_at": "2026-06-25T13:19:06.651013Z",
+    }
+    done_task, executing_task = _wildcard_event_setup(
+        behaviour, shared_state, request_data
+    )
+    # Flip the flags on both halves: the source decision keys on the
+    # executing_task; the response payload's is_offchain mirrors it.
+    executing_task["is_offchain"] = False
```
bennyjo, line 3450: This test reuses a fixture that always sets `sender`, so the new `sender → requester` fallback never runs and the empty-`requester` bug it fixes would go undetected — could you add a case with `requester` set and `sender` absent that asserts `event["response"]["requester"]` is populated?

bennyjo (commented):
Re-review: three prior blockers resolved and verified against source; one residual — the requester fallback fix lacks a discriminating test.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1876,10 +1884,23 @@ def _do_wildcard_write_best_effort(self) -> Generator:
...
+        # On-chain pending-task sweep: request-only events for tasks that
+        # have sat in the queue past the operator's max age. Appended to
+        # the same batch as delivered events so the signing round-trip
+        # costs one signature per settlement, not one per event class.
+        # ``_sweep_pending_undelivered`` returns the swept ``request_id``s
+        # alongside the events but does NOT mutate the pending queue: any
+        # guard-skip or POST failure below would otherwise drop the tasks
+        # with no re-queue and no written row. The queue is only mutated
+        # once the wildcard POST has confirmed a 2xx below.
+        request_only_events, swept_request_ids = self._sweep_pending_undelivered()
```
OjusWiZard, line 1929: 🟠 **Important — a poison sweep event drops legitimate delivered events from the lake, every round.**

Here the swept request-only events are merged into the **same atomic batch** as the delivered events (`events = events + request_only_events`), which is hashed + signed + POSTed all-or-nothing. If one swept event is permanently rejected (server 4xx, or a digest/JSON-build failure on requester-influenced content), the whole batch fails → `_drop_swept_from_pending` never runs → the task stays on `PENDING_TASKS` and re-poisons the batch on the next round.

Because delivered events are **not** retried across rounds (they come from this round's `synchronized_data.done_tasks`), every round the poison task co-occurs with real deliveries, those already-settled deliveries are silently and permanently lost. A single malformed on-chain request becomes a self-reinforcing drain on the analytics write path.

**Fix:** split delivered vs. sweep into separate batches/signatures, or on a batch failure retry delivered-only so real deliveries aren't held hostage by a request-only event.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2049,29 +2070,285 @@ def _do_wildcard_write_best_effort(self) -> Generator:
...
         )
+        # Wildcard confirmed the batch; drop the swept undelivered tasks
+        # from the pending queue now. Deferring to this point (rather than
+        # inside ``_sweep_pending_undelivered``) ensures that any early
+        # return above — missing mech address, missing marketplace, mixed
+        # chains, unconfigured chain_id, digest build, signing, JSON
+        # serialisation, POST timeout, or non-2xx status — leaves the
+        # sweep at-least-once: the tasks stay on the queue and the next
+        # FSM cycle re-attempts the write.
+        self._drop_swept_from_pending(swept_request_ids)
```
OjusWiZard: 🟠 **Important — 4xx is handled identically to 5xx, so terminally-rejected tasks retry forever (livelock).**

This drop only runs on a 2xx. The non-2xx branch above (~L2051) logs `4xx is terminal (bad payload, allowlist miss, signer-not-owner)` but then does the same thing for 4xx and 5xx: log a warning and return without dropping. So a hard 4xx (e.g. signer EOA not on the server allowlist) is re-swept and re-POSTed **every settlement round forever**, growing `PENDING_TASKS` unbounded (also iterated by the task_execution picker) and spamming identical warnings — while looking like normal fail-soft operation.

**Fix:** distinguish terminal (4xx) from retryable (5xx/network); on terminal, drop or move to a quarantine/dead-letter list with escalated (error-level) logging.

🟡 **Advisory (same line):** on 2xx the response body is never parsed, so if the server does a partial upsert (rejects some rows, still returns 200) `_drop_swept_from_pending` drops the whole `swept_request_ids` set anyway — the rejected rows' undelivered signal is lost with a "POST OK" log. Consider parsing a per-event ack and dropping only confirmed ids.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2049,29 +2070,285 @@ def _do_wildcard_write_best_effort(self) -> Generator:
...
+            str(delivery_rate_raw) if delivery_rate_raw is not None else None
+        )
+
+        # The wildcard server requires ``prompt`` to be a non-empty
+        # string and ``tool`` to be a non-empty string. Placeholders
+        # match the off-chain build's pattern ("[offchain request]")
+        # so a reader can spot a sweep-emitted row from the value.
+        prompt = "[onchain undelivered]"
+        tool = "unknown"
+        chain_id = int(getattr(self.params, "mech_events_chain_id", 0) or 0)
```
OjusWiZard, line 2511: 🟠 **Important — `mech_events_chain_id` is now two independently-configured params → silent full-batch drop on divergence.**

Request-only events read `mech_events_chain_id` off the **task_submission_abci** Params here, while delivered events read it off the **task_execution** Params. Both default to `0` and are declared in three skill.yaml files (`mech_abci:179`, `task_execution:116`, `task_submission_abci:178`) with **no override wired** in `aea-config.yaml`/`service.yaml`.

If an operator sets task_execution's copy (e.g. `100`) but forgets this new task_submission copy, a mixed round has delivered `chain_id=100` + swept `chain_id=0` → the heterogeneous-chain guard (~L1938) drops the **entire batch, including the valid delivered events**; a pure-sweep round gets `chain_id=0` and is skipped. The fix moved the bug from "always broken" to "silently broken on config drift."

Two stale artifacts reinforce it: the comment at ~L1934 ("rather than duplicating the param across two skills") is now false, and the trip-warning says "investigate task_execution" when the culprit is likely this newer copy.

**Fix:** single source of truth (or a startup consistency check that logs distinctly when the two values disagree), and update the comment + warning text.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1792,7 +1811,21 @@ def _build_wildcard_event(
...
+                # ``task_execution.handlers._enqueue_offchain_request``).
+                # Marketplace pending tasks come from
+                # ``MechMarketplaceContract.get_marketplace_undelivered_reqs``
+                # which keys the requester as ``requester`` and carries no
+                # ``sender`` field. This block now runs for on-chain
+                # marketplace deliveries too, so falling back to
+                # ``requester`` keeps the wildcard row correctly populated
+                # instead of writing an empty ``requester``.
+                "requester": str(
+                    executing_task.get("sender")
```
OjusWiZard, line 1825: 🧪 **Test gap — the `sender or requester` fallback added here has no test.**

The shared `_wildcard_event_setup` helper in `task_execution/tests/test_behaviours.py` always sets `executing_task["sender"]`, so the `or executing_task.get("requester")` branch — the marketplace-delivered path this fix restores — is never exercised. Add a case where the task carries only `requester` (no `sender`) and assert `event["request"]["requester"]` is populated.

OjusWiZard (commented):
Follow-up review of the fix commit `14d706ae` itself (the three earlier findings are confirmed fixed). New surface introduced by the at-least-once rework — all **latent behind the default-off flags**, so none blocks merge, but worth addressing before enabling in a live deployment (the `kv_store` replay-buffer follow-up would soften #1/#3). Substantive items inline.

**Also worth doing (not anchored inline):**
- **Advisory — silent skips have no logging.** In `_sweep_pending_undelivered`, the non-dict and missing/invalid `enqueued_at_local` branches `continue` with zero logging, and nothing ever re-stamps `enqueued_at_local` (`handlers.py` only `setdefault`s at first `filter_requests`). A genuinely unstamped task sits unswept and invisible forever — add a debug log so queue growth is diagnosable.
- **Test gap — no end-to-end `_do_wildcard_write_best_effort` test.** Only the two building blocks (`_sweep_pending_undelivered`, `_drop_swept_from_pending`) are tested in isolation. A refactor moving the drop *above* the POST (exactly the bug just fixed) would pass the whole suite. Add: POST fails/early-returns → task stays on `PENDING_TASKS`; POST 2xx → task dropped.
- **Test gap — `Params` still stubbed as `SimpleNamespace`.** No test instantiates the real `task_submission_abci.Params` from skill.yaml-shaped kwargs, so the `mech_events_chain_id` wiring that was just added could silently regress.

**Checked and cleared (not bugs):** `request_id` matching between sweep and drop (both `str()` the same value; int/bytes round-trips, no rebind hazard); duplicate-POST on retry (idempotent via the server's `ON CONFLICT … WHERE delivery_mech IS NULL` upsert).

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 24585e9a. `_do_wildcard_write_best_effort` now hands delivered and sweep events off to `_post_wildcard_batch` as TWO independent batches. A poison sweep event (server 4xx, or a digest/JSON-build failure on requester-influenced content) only kills the sweep batch; delivered events land unaffected. Since delivered events come from this round's `synchronized_data.done_tasks` and aren't re-queued, this isolates the failure so a single malformed on-chain request can't self-reinforce drain the analytics write path.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 24585e9a. `_post_wildcard_batch` now classifies non-2xx status: 4xx / 3xx are terminal (drop swept tasks anyway to break the livelock; error-level log so operators notice), 5xx / network / undefined are retryable (keep swept tasks on the queue, warn-level log). The JSON-serialisation failure branch also drops swept tasks — the same content would fail every subsequent round. The advisory on partial-upsert is noted in the code: today we drop the whole swept set on any 2xx because the server is all-or-nothing per batch; if we ever adopt partial upsert, we would need to parse a per-event ack from `response.body` and only drop confirmed ids. Called out inline so the follow-up is easy to find.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 24585e9a. Two changes: (1) split-batch design (see the sibling comment on the poison sweep) means a chain_id drift no longer drops the whole batch — the sweep batch short-circuits at the `chain_id=0` guard on its own while delivered lands. (2) Added `_check_mech_events_chain_id_consistency` on the behaviour: it compares `task_execution.params.mech_events_chain_id` and `task_submission_abci.params.mech_events_chain_id` and logs a single WARNING per behaviour instance on divergence (self-resets on mech restart, so operators aren't spammed but they do see the drift when it happens). Updated the stale L1934 comment ("rather than duplicating the param across two skills") and the heterogeneous-chain trip-warning to name both skills, per your callout.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in 24585e9a. Added `test_build_wildcard_event_marketplace_task_falls_back_to_requester_key` at the end of `test_behaviours.py`: it deletes `sender` on `executing_task`, injects `requester="0xMARKETPLACE_REQUESTER"`, and asserts the built event's `request.requester` is populated with that value. Also flips `is_offchain=False` so the on-chain source branch is exercised at the same time.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Same fix as OjusWiZard's parallel thread — 24585e9a adds `test_build_wildcard_event_marketplace_task_falls_back_to_requester_key` which does exactly this: strips `sender`, sets `requester`, asserts the delivered event carries the marketplace requester. The `_wildcard_event_setup` fixture stays as-is (the off-chain paths still need `sender`); the new test builds on top of it with a targeted delete + inject.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1851,6 +1864,61 @@ def async_act(self) -> Generator:
...
+        if self._chain_id_divergence_logged:
+            return
+        submission_chain_id = int(getattr(self.params, "mech_events_chain_id", 0) or 0)
+        # ``task_execution.Params`` is the same object graph — the
+        # PostTxSettlement behaviour extends TaskExecutionBaseBehaviour
+        # which composes both skills' params under ``self.params``. Pull
+        # the task_execution copy directly from the skill context so the
+        # comparison isn't tautological.
+        task_exec_params = getattr(
+            self.context.skills.task_execution.models, "params", None
```
bennyjo: `self.context.skills` doesn't exist on an AEA SkillContext, so this line throws on every settlement round — and since it runs before the feature-flag check, it crashes PostTxSettlement on every deployment, even with the new features off. A skill can't read another skill's params this way; drop the check or pass the value through `shared_state` instead.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2015,63 +2196,355 @@ def _do_wildcard_write_best_effort(self) -> Generator:
...
+            doesn't have the minimum fields the wildcard server
+            requires.
+        """
+        request_id = task.get("requestId")
+        if not request_id:
+            self.context.logger.warning(
+                "Pending sweep: skipping task with no requestId: %s", task
+            )
+            return None
+        request_id_str = str(request_id)
```
bennyjo: Undelivered tasks still hold `requestId` as raw bytes (the int conversion only happens when a task runs), so this produces `"b'\x12...'"` instead of the decimal id the delivered event uses — the two never match, so predict-api's ON CONFLICT reconciliation fails and a delivered request stays marked undelivered forever. Convert bytes→int here like the `content_cid` block just below does.

bennyjo (commented):
Two blocking issues from the recent fix commits — details inline.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2015,63 +2196,355 @@ def _do_wildcard_write_best_effort(self) -> Generator:
...
+                # nothing per batch, so this is exact.
+                self._drop_swept_from_pending(swept_request_ids)
+            return
+
+        # Non-2xx: distinguish terminal (4xx, 3xx) from retryable
+        # (5xx / undefined status / network-classified elsewhere).
+        # ``None`` here is a defensive undefined-status case, which the
+        # framework rarely surfaces; treat it as retryable so we don't
+        # silently poison-drop a healthy sweep on a bookkeeping edge.
+        is_terminal = status is not None and (
```
OjusWiZard: 🔴 **Critical — terminal-4xx drop silently and permanently discards both the analytics signal AND real queued work.**

This classifies **all** 4xx/3xx as terminal and then calls `_drop_swept_from_pending` (below) even though nothing was written. But the causes the comment names — allowlist miss (401/403), signer-not-owner, **429 rate-limit** (the class docstring notes a per-EOA limiter), **BATCH_HASH_MISMATCH** — are *systemic, whole-batch* failures, not single poison events. During the ordered "predict-api-first" rollout, or on any future hash drift, **every** sweep batch 4xxes → **every** swept task is dropped **every round**. Two altitudes:

1. **Analytics:** the entire undelivered-request signal is silently, irreversibly lost (error log only — no metric/dead-letter).
2. **Execution:** `PENDING_TASKS` is task_execution's **live work queue** — same `shared_state` key (all three modules define `PENDING_TASKS = "pending_tasks"`), and the executor does `self.context.shared_state[PENDING_TASKS].pop(0)`. Eviction removes entries the executor would run; because `_handle_get_undelivered_reqs` advances `from_block` past them (`handlers.py:417`), they're never re-scanned. Since `mech_events_sweep_max_age_seconds` (default 3600s) is a local proxy the docstring says "may differ" from the contract's real `responseTimeout`, if it's set below the on-chain timeout this **abandons still-deliverable, already-paid requests**.

Also: the JSON-serialization branch (~L2176) drops the *whole* swept set for a single non-serializable event (no per-event granularity).

This is strictly worse than the livelock it replaced — livelock is visible and self-healing; this is silent and irreversible.

**Fix:** don't evict on terminal 4xx (evict only on a confirmed 2xx); narrow "terminal" to genuine per-event rejections (e.g. 422) and treat 401/403/429/hash-mismatch as retryable; ideally decouple sweep eviction from the best-effort analytics POST outcome, or give the sweep its own `shared_state` key so analytics can't mutate the execution queue.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2015,63 +2196,355 @@ def _do_wildcard_write_best_effort(self) -> Generator:
...
+            if swept_request_ids:
+                self._drop_swept_from_pending(swept_request_ids)
             return
 
-        self.context.logger.info(
-            "Wildcard batch POST OK (status=%s batch_size=%d signer-eoa=%s)",
+        # Retryable (5xx / undefined): the server may recover on the next
+        # cycle. Leave sweep tasks on the queue.
+        self.context.logger.warning(
+            "Wildcard %s POST returned retryable non-success (status=%s); "
```
OjusWiZard: 🟠 **Important — the delivered batch has no parking spot, so a transient 5xx/network blip is silent permanent analytics loss.**

Delivered events come from this round's `done_tasks`, aren't re-queued across rounds, and (unlike the sweep batch) have no `PENDING_TASKS` fallback. A transient 5xx or network exception here drops those rows forever with only this WARNING — the replay buffer is a future PR. Funds/execution are safe (on-chain delivery already happened via the Safe), so this is a lake-completeness gap, not a correctness bug — but there's no counter/metric, so an operator has no signal the lake is missing rows for a window.

**Fix:** at minimum emit a metric (the codebase already has a `mech_metrics` wrapper) on retryable/terminal wildcard failures so the gap is observable before the replay buffer lands.

packages/valory/skills/task_submission_abci/wildcard_write_client.py
```diff
@@ -63,16 +63,36 @@ def canonical_json_bytes(value: Any) -> bytes:
...
+    the mech's own EOA already signed for; the mech owner is trusted
+    for the origin of their own rows.
+
     :param events: the ordered list of settled-delivery event dicts.
+        ``source`` may or may not be present on each dict; the hash
+        excludes it either way.
     :return: the 0x-prefixed 32-byte hex digest of the canonical batch.
     """
-    return "0x" + keccak(canonical_json_bytes(events)).hex()
+    hash_input = [{k: v for k, v in event.items() if k != "source"} for event in events]
```
OjusWiZard, line 94: 🧪 **Test gap — this `source`-exclusion has no test.**

This is the fix for the `BATCH_HASH_MISMATCH` rollout risk described in the commit, but `test_wildcard_write_client.py::TestComputeBatchHash` was left untouched and none of its fixtures ever include a `source` key — so all four existing hash tests pass whether or not this line exists. A revert (or an accidental refactor) of the `if k != "source"` filter would silently reintroduce `BATCH_HASH_MISMATCH` for every not-yet-upgraded off-chain mech during the rollout window, with no red test.

**Add:** `compute_batch_hash([{"a": 1, "source": "mech_onchain"}]) == compute_batch_hash([{"a": 1, "source": "mech_offchain"}]) == compute_batch_hash([{"a": 1}])`.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1851,6 +1864,61 @@ def async_act(self) -> Generator:
...
+                "round will use the task_execution value. Under the "
+                "split-batch flow the sweep batch may short-circuit at "
+                "the chain_id=0 guard while the delivered batch still "
+                "lands, so operators can miss this silently — check the "
+                "``mech_events_chain_id`` field on both skill.yaml / "
+                "aea-config overrides.",
+                task_exec_chain_id,
+                submission_chain_id,
+            )
+            self._chain_id_divergence_logged = True
```
OjusWiZard: ⚪ **Advisory — this latch goes blind after the first log, not just deduped.**

`_chain_id_divergence_logged` is a plain bool checked at the top of the method, so once any divergence is logged the *entire* comparison is skipped for the rest of the process lifetime. If the drift is later fixed and then a *different* divergence appears (e.g. a subsequent deploy sets a third value), it's invisible until a restart — stronger than the intended log-once-per dedup.

**Fix:** key the latch on the `(task_exec_chain_id, submission_chain_id)` pair (e.g. `_logged_divergence_pair`) so a new/different divergence is still surfaced. (Related: a non-zero-but-different drift doesn't hit the clean `chain_id==0` short-circuit — it POSTs with the wrong chain and lands in the terminal-4xx trap above.)

OjusWiZard (commented):
Fourth-pass review of commit `24585e9a` (the fixes for the third-pass findings). The batch split, hash-exclude-source symmetry, and requester-fallback test are all correct — but the terminal-4xx drop overcorrected the livelock into a silent data-loss path. Still gated behind the default-off flags (`mech_events_enabled` + `mech_events_sweep_pending_enabled`), so not a main-breaker, but #1 is worth fixing before the sweep is enabled. Substantive items inline.

**Test gaps (the whole new surface is untested — only the requester-fallback test was added, which is good):**
- `compute_batch_hash` excluding `source` — no test asserts two events differing only in `source` hash identically; a revert of the exclusion silently reintroduces `BATCH_HASH_MISMATCH` during rollout with no red test.
- Terminal (4xx/3xx → drop) vs retryable (5xx → keep) — no test drives `_post_wildcard_batch` with a mocked status and asserts drop/keep.
- **Two-batch isolation** (the headline fix) — nothing proves a failing sweep batch still lets the delivered batch POST.
- `_check_mech_events_chain_id_consistency` divergence-warning path — untested.

**Checked & cleared (not bugs):** hash shallow-strip is correct (only top-level `source`; POST body still sends full events); two-batch split is genuinely isolated; `source` outside the signature is a documented, reasonable trade-off.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 73c10d35. Dropped `_check_mech_events_chain_id_consistency` entirely — you're right, `self.context.skills` doesn't exist on AEA's SkillContext and my check crashed PostTxSettlement on every settlement round before the feature-flag gate. Replaced with a class-level comment on the two-params trade-off: the split-batch design already handles a chain_id drift acceptably (sweep short-circuits at `chain_id=0` with a visible log, delivered lands unaffected).

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 73c10d35. Now coerces bytes → int → str via `int.from_bytes(request_id, "big")` before stringifying, mirroring the content_cid block's structure. Added `test_request_only_event_coerces_bytes_request_id_to_decimal_string` that seeds the pending task with `(42).to_bytes(4, "big")` and asserts the built event's `request.request_id == "42"` — the same decimal shape the delivered path would emit, so predict-api's ON CONFLICT can now reconcile the two rows.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: You're right — the terminal-4xx drop was worse than the livelock. Reverted in 73c10d35. Only a confirmed 2xx now mutates `PENDING_TASKS`; every non-2xx path (4xx, 3xx, JSON serialisation failure, network error) leaves the swept tasks on the queue for retry. The comment block on the non-2xx branch explicitly names the four reasons your review flagged (live work queue, non-re-scanning `_handle_get_undelivered_reqs`, sweep-timeout-vs-responseTimeout mismatch, systemic-vs-per-event failure classes) so a future revisit doesn't re-introduce the same bug.\n\nOn the JSON-serialisation branch: also reverted to leave tasks on the queue. A single non-serialisable value shouldn't evict the whole swept set silently.\n\nOn the true per-event 422 case (server rejects a subset within an all-or-nothing batch): left as future work, needs a per-event ack from the server first. Noted inline.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Added `mech_wildcard_events_total` in 73c10d35, labelled `(batch_label, outcome)` where batch_label ∈ {delivered, sweep} and outcome ∈ {ok, http_error, pre_flight_failed, network_error}. Incremented on every terminal path in `_post_wildcard_batch`. Sustained non-`ok` on `batch_label="delivered"` will be visible on Prometheus before the kv_store replay-buffer PR lands. Followed the module-level `Counter(...)` pattern that `SharedState` uses for its existing Gauges rather than reaching into task_execution's `MechMetrics` (kept the analytics-write counter local to this skill's ownership).

packages/valory/skills/task_submission_abci/wildcard_write_client.py
DIvyaNautiyal07: Added in 73c10d35. `TestComputeBatchHash.test_hash_excludes_source_field` asserts `compute_batch_hash([{"a": 1, "source": "mech_onchain"}]) == compute_batch_hash([{"a": 1, "source": "mech_offchain"}]) == compute_batch_hash([{"a": 1}])`. Also added `test_hash_still_changes_on_non_source_fields` to guard against an accidental widening of the strip that would hide tamper on real fields — asserts that flipping `request.prompt` still moves the hash even when a `source` key sits alongside it.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: The whole check is gone now (see bennyjo's parallel thread — my `self.context.skills` path crashed on every round). Since the split-batch design already surfaces a drift via the sweep batch's `chain_id=0` short-circuit warning, the paired startup check isn't earning its cost. If we want it back later, the per-value-pair latch you suggested is the right shape — noted for whoever picks it up.

bennyjo (approved):
Both blockers from the earlier passes are resolved: the crashing `_check_mech_events_chain_id_consistency` (self.context.skills doesn't exist) is removed, and the swept `request_id` now coerces bytes→int→str (byteorder="big") to match the delivered path so predict-api's ON CONFLICT reconciles. The terminal-4xx eviction is correctly reverted — only a confirmed 2xx mutates PENDING_TASKS (the executor's live work queue), every other path leaves it for retry. New tests are discriminating and the outcome counter is a good add. One non-blocking FYI: the sweep uses `isinstance(bytes)` vs the delivered path's `type()==bytes` — agree for plain bytes today, latent divergence only if web3 ever returns HexBytes. Good to merge pending CI.

PR #462, opened 2026-07-02
fix(task_submission): preserve done_tasks across PostTxSettlementRound

DIvyaNautiyal07:
## Summary

- `PostTxSettlementRound.end_block` was clearing `synchronized_data.done_tasks` on `Event.DONE`. That field is what `TaskPoolingBehaviour.handle_submitted_tasks` reads at the start of the next cycle to know which `request_id`s to prune from `shared_state[DONE_TASKS]`.
- With the clear in place, `submitted_tasks` reaches `remove_tasks(...)` as `[]` on every cycle, so delivered tasks are never pruned. The same batch is re-pooled, re-prepared, and re-submitted every settlement round. Safe `execTransaction` succeeds each time, but `mech.deliverMulti` reverts on the already-delivered ids — burning gas, emitting no new `Deliver` events, degrading on-chain delivery rate.
- Restores `PostTxSettlementRound.end_block` to a straight `(synchronized_data, DONE)` return. `done_tasks` is already in `cross_period_persisted_keys` and now flows correctly into the next cycle's `handle_submitted_tasks`. Adds a regression test that pins the invariant.

## Root cause

Introduced by commit `330cb8ad` (PR #457 second-pass review). That commit added the clear to prevent the round from re-firing the wildcard POST on its `NO_MAJORITY` / `ROUND_TIMEOUT` self-loops. Correct concern, wrong layer — it guarded a shared consensus field that a downstream behaviour depends on. The re-fire concern is not observable while `mech_events_enabled=False` (behaviour short-circuits before the POST) and is safe when the flag is on (wildcard server is idempotent on `request_id`). If the round-level guard is wanted later, the right place for it is an idempotence flag inside `PostTxSettlementBehaviour`, not mutation of `synchronized_data.done_tasks`.

## Production impact observed

Single-agent Polygon mech (agent 21, offchain flag off): batch grew monotonically 10 → 19 over a 71-minute window; 150 finalized Safe txs across 150 cycles, all Success on Polygonscan; only ~4 actual `Deliver` events emitted (the first-time deliveries). Roughly ~13 POL/hour wasted at 275 gwei on redundant `execTransaction` calls.

Both `mech_events_enabled=False` and `mech_events_enabled=True` hit the bug — the round-level clear runs regardless of the behaviour-level flag.

## Test plan

- [x] `pytest packages/valory/skills/task_submission_abci/tests/` — 238/238 pass
- [x] `tomte tox -e black-check isort-check flake8 darglint mypy`
- [x] `autonomy packages lock --check`
- [x] `autonomy push-all`
- [ ] Deploy to single-agent Polygon mech; watch `Marketplace Tasks Found: N` — pre-fix it grows monotonically, post-fix it should track new task arrivals only (grows then drains on settlement)
- [ ] Cross-check Polygonscan: `ExecutionFailure` on Safe internal txs (currently visible) should stop appearing on the delivery Safe calls
- [ ] Multi-agent smoke: run on a 4-agent test service, confirm `TaskPoolingRound` consensus still lands cleanly

## Files

- `packages/valory/skills/task_submission_abci/rounds.py` — drop the clear, expand the class docstring so the invariant is explicit
- `packages/valory/skills/task_submission_abci/tests/test_rounds.py` — `test_done_tasks_preserved_on_done` regression pin
- `packages/packages.json`, `packages/valory/agents/mech/aea-config.yaml`, `packages/valory/services/mech/service.yaml`, `packages/valory/skills/mech_abci/skill.yaml`, `packages/valory/skills/task_submission_abci/skill.yaml` — IPFS hash bumps

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -8,12 +8,12 @@
         "connection/valory/websocket_client/0.1.0": "bafybeid4cnbjpmzcsw3usg7umufgqyod2u3xfw5uso3vukt3n5hfuxvdea",
         "skill/valory/contract_subscription/0.1.0": "bafybeicqlkmfpttw4hbbaakr6xos62rd6ytjwkyzlfli4xpvl5yqcycf6q",
         "skill/valory/delivery_rate_abci/0.1.0": "bafybeiaefr3eiajklnb6m4jtl7rpxr5t2lmrpizfy7gqkjem3qiewo7cz4",
-        "skill/valory/mech_abci/0.1.0": "bafybeidhjps2ctsggi3dpexdmuftfnvsgm2lgrotss3azsjufufpy2yfqu",
+        "skill/valory/mech_abci/0.1.0": "bafybeifqmjqfdhs3beaswozr6zugsst43tq6ias7chg5tjayekcr7e45ai",
         "skill/valory/task_execution/0.1.0": "bafybeifc56wu5xwjpjhfstgm6wbf6yptshhfbpqji7dsrtrdukkumiwwpq",
-        "skill/valory/task_submission_abci/0.1.0": "bafybeidzqkqghem72wyluktwe3qhtdruphrksx4qayoengjmprnkrh4dk4",
+        "skill/valory/task_submission_abci/0.1.0": "bafybeict3okafrtmy3hjnz2rlrcjjf5e2y7goqy5zr4svrvic2mlt54q2u",
         "skill/valory/websocket_client/0.1.0": "bafybeid6oyycsrvk4duftffs5mrfgmcepibt3qe4om3kfzhwhilfq3ynay",
-        "agent/valory/mech/0.1.0": "bafybeiet4mcpr36cm6wuprswfrxgibr5ob5rii277ynx65gl7a27qpekcm",
-        "service/valory/mech/0.1.0": "bafybeifpdygfygghydm7i3xfsnjbshdoyluq4dtu2izhit6utc43herg2u"
+        "agent/valory/mech/0.1.0": "bafybeiexugogtn7crjto2yxfwqj76y7w66cv7pabzdgjr3x5obztvjlemu",
+        "service/valory/mech/0.1.0": "bafybeiazqc66bhsw6gaw5eg2kng3utmlh654l6ymuumfdwnl65cwlplk2e"
     },
     "third_party": {
         "protocol/open_aea/signing/1.0.0": "bafybeifsjmldwyki3beqyvdt5lzenrg6wyrqaar5plc5rpnvtc4zlentye",
--- a/packages/valory/agents/mech/aea-config.yaml
+++ b/packages/valory/agents/mech/aea-config.yaml
@@ -40,12 +40,12 @@ skills:
 - valory/abstract_abci:0.1.0:bafybeie5et43th2nta76cd3kpm3fmurd7oyzk3qxim76octvt56m6r2poi
 - valory/abstract_round_abci:0.1.0:bafybeicioxhbhnsoqyqvkeevd7df7r2vubbiglyf7ucps2el73objazjsq
 - valory/contract_subscription:0.1.0:bafybeicqlkmfpttw4hbbaakr6xos62rd6ytjwkyzlfli4xpvl5yqcycf6q
-- valory/mech_abci:0.1.0:bafybeidhjps2ctsggi3dpexdmuftfnvsgm2lgrotss3azsjufufpy2yfqu
+- valory/mech_abci:0.1.0:bafybeifqmjqfdhs3beaswozr6zugsst43tq6ias7chg5tjayekcr7e45ai
 - valory/registration_abci:0.1.0:bafybeihjhbn5seffyu5cwugvm26fkamp3svhggmtzh577daijgl2ksxs3e
 - valory/reset_pause_abci:0.1.0:bafybeihc4ruh25s7wuunfy2b3k2n4eypyhvtry3usgom44wsymagbcr3ki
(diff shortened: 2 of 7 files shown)


dagacha:
Reviewed the change end-to-end against `main`. This is a correct fix for a real, high-severity bug.

## Root cause confirmed

Traced the exact mechanism:

1. `PostTxSettlementRound.end_block` clears `done_tasks` to `[]` on `Event.DONE`.
2. `done_tasks` is in `cross_period_persisted_keys`, so the cleared `[]` flows into the next cycle.
3. Next cycle, `TaskPoolingBehaviour.async_act` runs `handle_submitted_tasks()` first, which reads the synchronized copy: `submitted_tasks = self.synchronized_data.done_tasks`.
4. Because it's `[]`, the `len(submitted_tasks) > 0` guard is false, so `remove_tasks(...)` never runs and the delivered ids are never pruned from `shared_state[DONE_TASKS]`.
5. `get_done_tasks` then re-reads the still-populated `shared_state[DONE_TASKS]` → the same batch is re-pooled, re-prepared, and re-settled; `deliverMulti` reverts on already-delivered ids.

This matches the reported monotonic batch growth and the redundant Safe txs.

## The removed justification was genuinely flawed

The original comment claimed the clear prevented re-firing the wildcard POST on the `NO_MAJORITY` / `ROUND_TIMEOUT` self-loops. But the clear only ran on the `DONE` branch, and `DONE` routes to `FinishedPostTxSettlementRound` (a final state), never a self-loop. The self-loop arms never touch `done_tasks` in either version, so the clear could not have addressed that concern. "Correct concern, wrong layer" is accurate.

## No regression to the wildcard path

`PostTxSettlementBehaviour._extract_offchain_events` reads `synchronized_data.done_tasks`, but `TaskPoolingRound.end_block` overwrites `done_tasks` every cycle, so by the time `PostTxSettlement` runs again it only carries the current cycle's events. Combined with `mech_events_enabled=False` short-circuiting the POST and the server being idempotent on `request_id`, preserving `done_tasks` introduces no duplicate-POST issue. No unbounded growth in `synchronized_data` either, since the field is rewritten each cycle.

## Test & housekeeping

- `test_done_tasks_preserved_on_done` pins the invariant correctly (would fail on `main`, passes with the fix).
- No unused imports left behind (`get_name` / `SynchronizedData` still used by the other rounds).
- IPFS hash bumps are consistent across `skill.yaml`, `mech_abci`, agent, and service.

## Minor, non-blocking

- The new test is a unit-level pin on `end_block` only; the pruning it protects lives in `handle_submitted_tasks`. A small integration-style test asserting `remove_tasks` receives the delivered ids after a settlement cycle would guard the invariant end-to-end.
- Suggest completing the remaining deploy-verification checklist items (Polygon watch, Polygonscan cross-check, multi-agent smoke) before merge, since the failure mode is on-chain and gas-costly.

LGTM once the deploy verification is done.

dagacha (approved):

bennyjo (approved):
Approving — verified the fix end-to-end. Removing the `done_tasks` clear in `PostTxSettlementRound.end_block` restores the cross-period flow into `TaskPoolingBehaviour.handle_submitted_tasks` → `remove_tasks`, so delivered batches are pruned from `shared_state[DONE_TASKS]` again and the re-pool/re-deliver loop (reverting `deliverMulti`, wasted gas) stops. FSM audit is clean: transition function complete, `db_post_conditions[FinishedPostTxSettlementRound]` is empty so the unchanged return is fine, and `done_tasks` stays in `cross_period_persisted_keys`. Regression test is discriminating (fails against the old `done_tasks: []` return). No security/attacker surface touched.

Non-blocking follow-up: the PR #457 guard against re-firing the wildcard POST on the `NO_MAJORITY`/`ROUND_TIMEOUT` self-loops is now moot only because `mech_events_enabled=False`. The docstring says the concern should move to a behaviour-level idempotence flag, but that isn't in this PR — worth tracking so it lands before `mech_events_enabled` is flipped on (harmless meanwhile: the POST is idempotent server-side).

jmoreira-valory (approved):
Independent verification at HEAD 4809218 (post-merge, CI 22/22 green across all OS/Python combos):

- Causal chain confirmed: `done_tasks` is included in `cross_period_persisted_keys` (rounds.py:322-327) alongside `final_tx_hash`; `handle_submitted_tasks` calling `remove_tasks([])` is a no-op that preserves the full task set (behaviours.py:285-322, :195-215); `_do_wildcard_write_best_effort` short-circuits when `mech_events_enabled=False`, which is the default (skill.yaml:172) and not overridden in aea-config.
- Asymmetry fix verified: the erroneous clear from commit 330cb8ad is now removed from BOTH the DONE and NO_MAJORITY arms of `PostTxSettlementRound.end_block`. 330cb8ad's own rationale ("both self-loop") never matched its diff — it only added the clear on the DONE arm, so NO_MAJORITY was already silently dropping tasks before this fix.
- Regression test `test_done_tasks_preserved_on_done` (tests/test_rounds.py:521-538) is falsifiable — verified it fails against pre-fix code.
- FSM untouched: only `end_block` logic and docstring changed, no Event/transition_function/fsm_specification.yaml drift, no re-sync needed.

Two non-blocking follow-ups for a fast-follow PR:

1. NO_MAJORITY arm has no direct regression test — extend `test_no_majority_when_impossible` (~tests/test_rounds.py:526) with `assert new_sync_data.done_tasks == tasks` to lock in the symmetric fix the same way DONE is covered.
2. @dagacha's LGTM was conditioned on deploy verification; per the PR test plan, 3 of 6 checkboxes are still unchecked (single-agent Polygon deploy watch, Polygonscan ExecutionFailure stop-check, 4-agent consensus smoke test). Worth confirming those post-merge since this already shipped.

LGTM.

PR #464, opened 2026-07-02
feat(balance_tracker): add build_deposit_for_data + native variant package

DIvyaNautiyal07:
## Summary

Two additive changes to unblock mech-interact removing its slim BalanceTracker duplicates and consuming the canonical packages as `third_party`.

- **`balance_tracker` (existing token variant)**: adds `build_deposit_for_data(ledger_api, contract_address, account, amount)` that encodes `depositFor(account, amount)` for a Safe multisend batch (paired with an ERC20 `approve`). Return shape matches the file convention (`{"data": bytes.fromhex(...)}`); no `0x` prefix guard because sibling `get_process_payment_tx` doesn't guard either.
- **`balance_tracker_fixed_price_native` (new package)**: sibling variant for the fixed-price native (xDAI / ETH) payment model. Exposes `get_requester_balance(requester)` and `build_deposit_for_data(account)` — no `amount` parameter because `depositFor(address account) payable` takes `msg.value` on chain, not calldata. ABI JSON adapted from mech-interact's existing slim wrapper.

## Why

mech-interact currently ships slim `balance_tracker_fixed_price_token` and `balance_tracker_fixed_price_native` packages that duplicate reads already canonical in mech's `balance_tracker` (token variant) plus a `build_deposit_for_data` that only lives in mech-interact. Landing the extension + native sibling here lets mech-interact delete both slim packages and consume mech's canonicals as `third_party`, which also stops the duplication fan-out into downstream consumers (optimus, trader).

## Invariant

Additive only. No existing method is renamed, resignatured, or altered in return-key shape. Consumers pick this up via a routine `third_party` hash bump in `mech-tools-dev`, `mech-predict`, `mech-agents-fun` at their next re-lock — no source changes required.

## Verification

- `autonomy packages lock --check` green
- `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` all pass
- `tomte tox -p -e check-hash -e check-packages -e check-dependencies` all pass
- New method style matches sibling `get_process_payment_tx` (encode_abi + `bytes.fromhex(data[2:])`)

## Test plan

- [ ] CI green on the branch
- [ ] mech-interact follow-up PR that swaps its slim wrappers for these canonicals exercises both methods end-to-end against a live marketplace (offchain-request e2e already covers this shape)

## Notes on base

Targeting `offchain-epic` (not `main`) because mech-interact's slim wrappers currently live on that epic. Merging this into `main` while the epic is still open would create a hash race between the two lines.


packages/valory/contracts/balance_tracker/contract.py
```diff
@@ -134,3 +134,19 @@ def get_process_payment_tx(
...
+        cls,
+        ledger_api: EthereumApi,
+        contract_address: str,
+        account: str,
+        amount: int,
+    ) -> JSONLike:
+        """Encode depositFor(account, amount) for a Safe multisend batch."""
+        contract_instance = cls.get_instance(ledger_api, contract_address)
+        data = contract_instance.encode_abi(
+            abi_element_identifier="depositFor",
```
OjusWiZard, line 149: **Critical — `depositFor` is not in the bundled ABI; this method fails on every call.**

`get_instance` builds the contract from `build/BalanceTracker.json`, but that artifact is the abstract `BalanceTrackerBase` (its `contractName` field says so): it has a `Deposit` *event* but no `depositFor` *function*. Reproduced against the packaged ABI: `encode_abi("depositFor", args=[account, amount])` raises `web3.exceptions.MismatchedABI: ABI Not Found! No element named depositFor with 2 argument(s)`.

The concrete token contract does expose it — `DEPOSIT_FN_ABI = "depositFor(address requester, uint256 amount)"` in `task_execution/handlers.py:156` — the packaged artifact is just stale/wrong. The PR bumps only the `contract.py` fingerprint; `build/BalanceTracker.json` was never regenerated.

**Fix:** regenerate `build/BalanceTracker.json` from the concrete `BalanceTrackerFixedPriceToken` source (or add the `depositFor(address,uint256)` ABI entry), update its fingerprint in `contract.yaml`, and re-run `autonomy packages lock`. The native package in this PR shows the correct pattern (its ABI is the concrete contract and encodes fine).

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
```diff
@@ -0,0 +1,66 @@
...
+        return {"requester_balance": requester_balance}
+
+    @classmethod
+    def build_deposit_for_data(
+        cls,
+        ledger_api: EthereumApi,
+        contract_address: str,
+        account: str,
+    ) -> JSONLike:
+        """Encode depositFor(account) calldata."""
```
OjusWiZard: **Advisory — the `tx.value` requirement is invisible to callers.**

On-chain `depositFor(address)` is `payable`: the deposit amount travels as the transaction's native value, not calldata. Nothing in the signature, return shape, or docstring says so — unlike the token sibling, whose explicit `amount: int` captures that invariant. A future caller could easily build a zero-value deposit.

Suggestion: accept `amount: int` and return `{"data": ..., "value": amount}` so the contract API is self-documenting; or at minimum note in the docstring that the caller must set the tx value to the deposit amount.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
```diff
@@ -0,0 +1,66 @@
...
+
+from aea.common import JSONLike
+from aea.configurations.base import PublicId
+from aea.contracts.base import Contract
+from aea.crypto.base import LedgerApi
+from aea_ledger_ethereum import EthereumApi
+
+PUBLIC_ID = PublicId.from_str("valory/balance_tracker_fixed_price_native:0.1.0")
+
+_logger = logging.getLogger(
```
OjusWiZard, line 32: **Advisory (nit)** — `_logger` is never used in this file (there's no `simulate_tx`-style method here, unlike the token sibling). Same scaffold boilerplate exists in other packages (`multisend`, `gnosis_safe_proxy_factory`), so fine to keep for consistency — just flagging that it's dead as of this PR.

OjusWiZard (commented):
Multi-perspective review (/ecc:review-pr): 1 Critical, 3 Advisory. Inline comments below cover the Critical ABI mismatch and two advisories.

**Advisory (no anchor):** neither new encode path has a test. That's consistent with repo convention (dev-authored contract packages have no `tests/` dirs and CI's `pytest_targets` only covers the three skill test dirs), but a small unit test asserting the returned bytes start with the expected `depositFor` selector would have caught the Critical below, and would guard against a future web3 upgrade changing `encode_abi`'s output format. Worth adding either here or in the follow-up PR that wires these into behaviours.

Verified clean: all package hashes and `contract.yaml` fingerprints independently recomputed and match; native ABI matches both methods (live `encode_abi` succeeds); the unguarded `data[2:]` is safe under pinned `open-aea-ledger-ethereum==2.2.7` (always `0x`-prefixed); types, docstring style, and formatting match repo convention.

packages/valory/contracts/balance_tracker/contract.py
DIvyaNautiyal07: Good catch. Confirmed the artifact was BalanceTrackerBase not the concrete token. Added the depositFor(address,uint256) entry to build/BalanceTracker.json in 19e8e2ac, kept tokenCreditRatio intact so the other reads on this class still work. Verified locally with encode_abi: selector 0x2f4f21e2, 138-byte payload.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
DIvyaNautiyal07: Good suggestion, took the API-change option. build_deposit_for_data now accepts amount: int and returns {"data": ..., "value": amount} so the payable requirement is explicit in the response shape. mech-interact's caller will pass amount when the follow-up PR wires this in.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
DIvyaNautiyal07: Ack, keeping _logger for scaffold consistency with the sibling packages.

packages/valory/contracts/balance_tracker/contract.py
DIvyaNautiyal07: Sourced from autonolas-marketplace/abis/0.8.28. dc103e85 rebuilds build/BalanceTracker.json from canonical: base is BalanceTrackerFixedPriceToken.json (concrete contract behind mech's balance_tracker), plus the tokenCreditRatio entry copied from BalanceTrackerNvmSubscriptionToken.json (needed because _adjust_mech_balance in task_submission_abci uses the same package for NVM mechs). Every function/event/error in the resulting ABI now traces back to a canonical marketplace artifact.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
DIvyaNautiyal07: Actually reverted this in 34bab1d7. Went with your docstring alternative instead. Reasoning: (1) other build_*_for_data methods in the framework return {"data": ...} only, so adding value breaks convention; (2) the amount is redundant, the caller already has it; (3) the token variant's amount param is functional (it's encoded into the calldata), replicating it on the native side purely for documentation invites confusion about which value the framework actually cares about. Docstring now says the caller must set tx.value to the deposit amount.

OjusWiZard (approved):

packages/valory/contracts/balance_tracker/contract.py
```diff
@@ -134,3 +134,19 @@ def get_process_payment_tx(
             ledger_api, contract_address, sender_address, tx_data
         ).pop("data")
         return {"data": bytes.fromhex(tx_data[2:]), "simulation_ok": simulation_ok}  # type: ignore
+
+    @classmethod
+    def build_deposit_for_data(
```
bennyjo, line 139: Would you consider adding a minimal encode-assertion test (selector + payload length) for `build_deposit_for_data` in both variants? This method already shipped once against the wrong ABI (`MismatchedABI` on every call) and was caught in review rather than by a test — you validated the fix by hand (`selector 0x2f4f21e2`), so capturing exactly that would make the ABI-mismatch regression deterministic to catch instead of relying on the deferred mech-interact e2e.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
```diff
@@ -0,0 +1,66 @@
...
+        return {"requester_balance": requester_balance}
+
+    @classmethod
+    def build_deposit_for_data(
+        cls,
+        ledger_api: EthereumApi,
+        contract_address: str,
+        account: str,
+    ) -> JSONLike:
+        """Encode depositFor(account) calldata; caller must set tx.value to the deposit amount (depositFor is payable)."""
```
bennyjo: Confirmed against the on-chain source (autonolas-marketplace `BalanceTrackerFixedPriceNative.sol`): `depositFor(address)` is just `mapRequesterBalances[account] += msg.value` with **no zero-value guard** — a value-less call doesn't revert, it silently credits zero and emits a `Deposit(account, 0)` that looks like a real deposit. So this docstring is the *only* thing guaranteeing the amount gets attached; there's no on-chain backstop. Since `build_deposit_for_data(account)` never receives the amount it can't validate here — so could we either (a) re-add `amount`, `require > 0`, and return `{"data": ..., "value": amount}` so a caller spreading the dict can't omit value, or (b) confirm the mech-interact caller asserts `value > 0` before submit? Flagging because the failure is a silent no-op deposit that reports success.

bennyjo (commented):
Additive change looks correct and the token ABI is a safe superset — two non-blocking items: an encode test for the new method, and a verified silent-zero-deposit gap on the native variant.

packages/valory/contracts/balance_tracker/contract.py
DIvyaNautiyal07: Done in 2aff76fb. Added packages/valory/contracts/balance_tracker/tests/test_contract.py and the sibling for the native variant, registered both under [tool.tomte].pytest_targets so CI picks them up. Assertions cover: (1) 4-byte selector matches keccak of the Solidity signature (2f4f21e2 for token, aa67c919 for native), (2) calldata length matches (68 / 36 bytes), (3) amount decodes correctly from the tail 32 bytes on token, value == amount on native, (4) ValueError on amount <= 0 for native (see reply to your other comment for the guard). Would have caught the wrong-ABI regression from the earlier review deterministically.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
DIvyaNautiyal07: You're right, this is a real hazard — silent no-op deposit that reports success. Reversed my earlier flip in 2aff76fb: re-added amount: int, added a ValueError guard for amount <= 0, and return {"data": ..., "value": amount} so a caller spreading the dict into the tx builder cannot silently drop value. The value key isn't documentation echo anymore, it's a functional wiring that closes the silent-no-op path.

bennyjo (approved):
Approving — genuinely additive, ABI change verified purely additive (no function entries removed), and both encoders check out against the on-chain selectors: token `depositFor(address,uint256)` → `2f4f21e2`, native `depositFor(address)` → `aa67c919`, matching the test constants. The encode-assertion tests are well-constructed and would catch the wrong-ABI regression the commit history describes.

One non-blocking item worth a reply: guard asymmetry. The native builder raises on `amount <= 0` to prevent a silent zero-value deposit, but the token `build_deposit_for_data` has the same failure mode (`amount=0` credits zero, no-op `transferFrom(0)`, emits a bogus `Deposit(0)`) with no guard. Worth either adding the same `amount <= 0` guard on the token side for parity, or noting why the two differ. Happy to leave as-is if the token path validates amount upstream.

PR #468, opened 2026-07-08
feat(mech): dual-purpose use_offchain, drop mech_events_enabled, rename wildcard→predict_api

DIvyaNautiyal07:
## Summary

Companion to [valory-xyz/predict-api#166](https://github.com/valory-xyz/predict-api/pull/166) (which drops the server-side kill switch and seeds the production operator allowlist). This PR simplifies the client-side config surface to one flag with clear semantics and renames the wildcard-branded params.

### 1. Dual-purpose `use_offchain`, drop `mech_events_enabled`

The two flags now collapse into one:

- **`use_offchain=true`** — mech accepts offchain requests via the offchain HTTP handler AND writes settlement events (both offchain-settled and on-chain-settled) to the predict-api events endpoint from PostTxSettlement.
- **`use_offchain=false`** — offchain endpoint returns 503 AND no settlement events are written at all. Legacy on-chain + IPFS behaviour unchanged.

Default stays `false`. Writes require the mech to be in predict-api's `mech_operators_by_chain` allowlist. A mech is either part of the predict-api analytics pipeline or it isn't; there's no coherent state where ingress and egress diverge, so a single flag is sufficient.

`mech_events_enabled` is deleted from both `task_execution` and `task_submission_abci` `models.py` and `skill.yaml`. All gate points in `behaviours.py` now read `use_offchain`.

### 2. Rename `wildcard_*` → `predict_api_*`

Config params (defaults in the composed skill):

| Old | New | Default |
|---|---|---|
| `wildcard_events_url` | `predict_api_events_url` | `https://mpp.autonolas.tech/mech/events` |
| `wildcard_events_timeout_seconds` | `predict_api_events_timeout_seconds` | `5.0` |

Files (renamed via `git mv` — history preserved):

- `packages/valory/skills/task_submission_abci/wildcard_write_client.py` → `predict_api_events_client.py`
- `packages/valory/skills/task_submission_abci/tests/test_wildcard_write_client.py` → `tests/test_predict_api_events_client.py`

Internal helpers: `_post_wildcard_batch` → `_post_predict_api_batch`, `_build_wildcard_event` → `_build_predict_api_event`, `wildcard_source` → `predict_api_source`, `done_task[\"wildcard_event\"]` → `done_task[\"predict_api_event\"]`, `_do_wildcard_write_best_effort` → `_do_predict_api_write_best_effort`.

**On-wire EIP-712 domain constants (`Olas Mech Events`, `MechEventBatch`, `1`) stay exactly as-is — the server binds to them, do NOT rename.**

### 3. Behaviour-change note: `predict_api_events_url` default

The rename in §2 also flipped the default value for this param, so calling it out explicitly rather than burying under "rename":

| | Old default | New default |
|---|---|---|
| `wildcard_events_url` / `predict_api_events_url` | `""` (short-circuit — egress helper returned before any HTTP work) | `https://mpp.autonolas.tech/mech/events` |

Why this is safe under the flag semantics:
- The opt-in stays **`use_offchain`**, not the URL. A legacy mech upgrading with `use_offchain=false` sees zero behaviour change; the egress helper still short-circuits at the flag gate above the URL check.
- When `use_offchain=true`, the URL now auto-targets production. Deployments that want to point at a staging predict-api set `predict_api_events_url` in the composed skill's `params.args` block just as before.
- The empty-string sentinel path (URL-missing DEBUG log) still exists — flip the flag on with an empty URL and you get a quiet DEBUG line, not a WARNING, for staged rollouts where the endpoint isn't wired yet.

## Test plan

- [x] `autonomy packages lock` clean — skill → agent → service hashes cascade
- [x] 112 tests pass across `task_execution/tests/test_handlers.py`, `task_submission_abci/tests/test_predict_api_events_client.py`, `task_submission_abci/tests/test_onchain_write_path.py`
- [x] Ingress-gate coverage on `_finalize_done_task` (`use_offchain=True` off-chain HTTP + on-chain marketplace shapes, and the `use_offchain=False` dark-ship guarantee).
- [x] Egress-gate coverage on `_do_predict_api_write_best_effort` (`use_offchain=False` short-circuit, empty-URL DEBUG short-circuit, silent-divergence WARNING when the two flag copies drift).
- [ ] Full CI green
- [ ] Downstream (mech-predict / trader / optimus / basius / meme-ooorr) pick up the rename cleanly on the next `upstream_pins` bump

### Prometheus metric rename

`mech_wildcard_events_total` → `mech_predict_api_events_total`. No Grafana dashboard or alert rule keyed on the old name at time of merge (checked before shipping), so no dual-emit needed. If a consumer surfaces later, the counter is a straight rename with no label-shape change; migrating the query is a one-line find-and-replace on the metric name.

### Deployment coordination for the 4 mechs currently running `use_offchain=true`

The `predict_api_events_url` default changed from `""` (short-circuit) to `https://mpp.autonolas.tech/mech/events`. Any mech that ships with `use_offchain=true` and no explicit URL override will start POSTing settlement events to prod on its next upgrade. Handled operationally rather than in this PR: the 4 mechs already running `use_offchain=true` will have the flag set back to `false` in agent-deployments before the redeploy that picks up this change, so they explicitly re-enrol via config rather than silently upgrade into POSTing.

## Downstream migration

Any deployment currently setting `MECH_EVENTS_ENABLED=true` or `WILDCARD_EVENTS_URL=...` at the env-var layer will emit \"unknown param\" warnings from the aea loader after upgrade but will otherwise boot. In practice these envs are unset today (predict-api endpoint was never live in prod), so the rename is cost-free.

After merge: cut a new mech tag (v0.34.3 or similar) so downstreams have a pin.

### DIFF
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1553,7 +1553,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # for the paid HTTP path, ``mech_onchain`` for the on-chain rails.
         # See ``autonolas-marketplace/docs/onchain_write_path_scope.md`` for
         # the agreed shape.
-        predict_api_mode_enabled = getattr(self.params, "use_offchain", False)
+        predict_api_mode_enabled = self.params.use_offchain
         is_marketplace_delivery = bool(done_task.get("is_marketplace_mech"))
         if predict_api_mode_enabled and (is_offchain or is_marketplace_delivery):
             done_task["predict_api_event"] = self._build_predict_api_event(
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -103,7 +103,7 @@ def _iso_z(dt: datetime) -> str:
 
     Normalises through ``astimezone(timezone.utc)`` so a datetime with any
     UTC offset (``+00:00``, ``+05:30``, ``-08:00``) emits the same canonical
-    string for the same wall-clock instant. The wildcard server's Pydantic
+    string for the same wall-clock instant. The predict-api server's Pydantic
     ``AwareDatetime`` round-trips through ``model_dump(mode="json")`` which
     emits ``...Z`` for UTC; the mech's ``batch_hash`` recompute on the
     server side compares byte-for-byte against ours, so leaving the offset
@@ -127,10 +127,10 @@ def _iso_z(dt: datetime) -> str:
 
 
 # Cap on the JSON-serialised size of each ``raw_content`` blob attached to
-# a wildcard event. The blob rides Tendermint consensus replication into
+# a predict-api event. The blob rides Tendermint consensus replication into
 # ``synchronized_data`` on every agent and the field is requester-influenced
 # (``params``, ``model``, response body), so an uncapped large payload would
-# inflate per-node state. 256 KiB matches the wildcard server's ``_MAX_TEXT``
+# inflate per-node state. 256 KiB matches the predict-api server's ``_MAX_TEXT``
 # cap on individual TEXT columns; the analytics ETL drops the row to a
 # truncated sentinel when the cap is hit so the rest of the event still
 # lands.
@@ -1533,30 +1533,30 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # pop the data key value as it's bytes which causes issues
         # with json dumps and not required anywhere
         done_task.pop("data", None)
-        # Attach the offchain wildcard-event payload before the
+        # Attach the offchain predict-api event payload before the
         # IN_MEMORY_REQUESTS pop below; the post-settlement behaviour in
         # task_submission_abci reads it off ``synchronized_data.done_tasks``
-        # (i.e. after consensus replication) and posts it to the wildcard
+        # (i.e. after consensus replication) and posts it to the predict-api
         # data lake, but the buffered request metadata it needs is local
         # to this agent's shared_state and goes away at the pop.
         #
-        # Gated on BOTH ``is_offchain`` AND the ``mech_events_enabled`` flag
+        # Gated on BOTH ``is_offchain`` AND the ``use_offchain`` flag
         # so Phase 1 is genuinely dark: when the flag is off, no payload is
         # built and nothing rides Tendermint consensus replication. The flag
         # check lives here as well as in the post-settlement behaviour so
         # the consensus-state cost is gated, not just the HTTP write.
-        # Build the wildcard event for both off-chain HTTP deliveries
+        # Build the predict-api event for both off-chain HTTP deliveries
         # (is_offchain=True) and on-chain marketplace deliveries
         # (is_offchain=False AND is_marketplace_mech). The ``source`` field
-        # on the event (set by ``_build_wildcard_event``) is what
-        # disambiguates the two paths on the wildcard side: ``mech_offchain``
+        # on the event (set by ``_build_predict_api_event``) is what
+        # disambiguates the two paths on the predict-api side: ``mech_offchain``
         # for the paid HTTP path, ``mech_onchain`` for the on-chain rails.
         # See ``autonolas-marketplace/docs/onchain_write_path_scope.md`` for
         # the agreed shape.
-        wildcard_mode_enabled = getattr(self.params, "mech_events_enabled", False)
+        predict_api_mode_enabled = getattr(self.params, "use_offchain", False)
         is_marketplace_delivery = bool(done_task.get("is_marketplace_mech"))
-        if wildcard_mode_enabled and (is_offchain or is_marketplace_delivery):
-            done_task["wildcard_event"] = self._build_wildcard_event(
+        if predict_api_mode_enabled and (is_offchain or is_marketplace_delivery):
+            done_task["predict_api_event"] = self._build_predict_api_event(
                 done_task=done_task,
                 cid=cid,
                 executing_task=cast(Dict[str, Any], executing_task),
@@ -1571,29 +1571,29 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         self.context.shared_state.get(IN_MEMORY_REQUESTS, {}).pop(req_id, None)
         self._reset_executing_task()
 
-    def _build_wildcard_event(
+    def _build_predict_api_event(
         self,
         *,
         done_task: Dict[str, Any],
         cid: str,
         executing_task: Dict[str, Any],
     ) -> Dict[str, Any]:
-        """Build the structured wildcard-event payload from on-hand data.
+        """Build the structured predict-api event payload from on-hand data.
 
         The post-settlement behaviour batches every event from one FSM round
-        into a single signed POST to the wildcard data lake. Fields are
+        into a single signed POST to the predict-api data lake. Fields are
         populated best-effort from the buffered request metadata
         (``IN_MEMORY_REQUESTS``), the offchain response
         (``OFFCHAIN_REQUEST_RESPONSES``), the in-flight ``executing_task``,
         and skill params. Optional fields that aren't readily available are
-        left ``None``; the wildcard server accepts a NULL there. Required
+        left ``None``; the predict-api server accepts a NULL there. Required
         fields default to safe placeholders rather than blocking the row: a
-        malformed event surfaces as a 422 when the wildcard write fires (and
+        malformed event surfaces as a 422 when the predict-api write fires (and
         lands in the local replay buffer so it's recoverable), which is
         preferable to silently dropping analytics data.
 
         Datetimes are emitted as ISO 8601 strings with UTC offset so the
-        wildcard ``AwareDatetime`` parser accepts them; mech ints (Unix
+        predict-api ``AwareDatetime`` parser accepts them; mech ints (Unix
         seconds) are converted here so the structured event can ride the
         FSM consensus channel without round-tripping through a custom
         codec.
@@ -1654,15 +1654,15 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         delivery_mech = str(
             done_task.get("mech_address") or self.params.mech_marketplace_address or ""
         )
-        # Wildcard ``source`` derives from the per-task ``is_offchain``
+        # The predict-api ``source`` derives from the per-task ``is_offchain``
         # flag: True → ``mech_offchain`` (paid HTTP path); False →
         # ``mech_onchain`` (the on-chain marketplace rails). This is the
-        # only field that disambiguates the two paths on the wildcard
+        # only field that disambiguates the two paths on the predict-api
         # side; the per-row ``is_offchain`` boolean stays on the
         # response payload as well (the lake keeps both because the
         # ipfs_historical backfill describes either path).
         is_offchain_task = bool(executing_task.get("is_offchain", False))
-        wildcard_source = "mech_offchain" if is_offchain_task else "mech_onchain"
+        predict_api_source = "mech_offchain" if is_offchain_task else "mech_onchain"
 
         # `start_time` is a perf_counter() value (monotonic, not wall-clock),
         # not suitable as a delivered_at timestamp. Use the current wall-clock
@@ -1698,7 +1698,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # the executed_at on the response so the time-leading index entry is
         # always populated for an offchain row. The requester can store the
         # value as a Unix int or as an ISO 8601 string — ``str(int)`` would
-        # produce ``"1700000000"`` which the wildcard's ``AwareDatetime``
+        # produce ``"1700000000"`` which the predict-api's ``AwareDatetime``
         # parser rejects with a 422, so coerce int / float explicitly.
         requested_at_raw = request_data.get("datetime") or request_data.get(
             "requested_at"
@@ -1772,7 +1772,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             nonce_int = int(nonce_raw) if nonce_raw is not None else None
         except (TypeError, ValueError):
             nonce_int = None
-        # The wildcard server requires ``prompt`` to be a non-empty string;
+        # The predict-api server requires ``prompt`` to be a non-empty string;
         # placeholder rather than dropping the row. Enforce the cap on the
         # UTF-8 BYTE length, not on ``len()`` (code points) — a CJK prompt
         # of 50k code points encodes to ~150 KB and would otherwise sail
@@ -1788,11 +1788,11 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # Failed-row ``error`` carries the available diagnostic, regardless
         # of which branch put us on the failed arm. When ``invalid`` is set
         # but no ``result_text`` is available (the tool produced no
-        # response object), the field is ``None``; the wildcard server's
+        # response object), the field is ``None``; the predict-api server's
         # CHECK allows a ``failed`` row with empty error so the row still
         # lands as a settled failure.
         error_text = str(result_text) if status == "failed" and result_text else None
-        # Strip the result from the failed arm so the wildcard's
+        # Strip the result from the failed arm so the predict-api's
         # status/error/result shape CHECK accepts the row.
         result_value: Optional[str] = (
             None if status == "failed" else (str(result_text) if result_text else "")
@@ -1819,7 +1819,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 # which keys the requester as ``requester`` and carries no
                 # ``sender`` field. This block now runs for on-chain
                 # marketplace deliveries too, so falling back to
-                # ``requester`` keeps the wildcard row correctly populated
+                # ``requester`` keeps the predict-api row correctly populated
                 # instead of writing an empty ``requester``.
                 "requester": str(
                     executing_task.get("sender")
@@ -1878,7 +1878,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 "response_cid": None,
                 "delivered_at": now_iso,
             },
-            "source": wildcard_source,
+            "source": predict_api_source,
         }
 
     def _reset_executing_task(self) -> None:
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1540,18 +1540,23 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # data lake, but the buffered request metadata it needs is local
         # to this agent's shared_state and goes away at the pop.
         #
-        # Gated on BOTH ``is_offchain`` AND the ``use_offchain`` flag
-        # so Phase 1 is genuinely dark: when the flag is off, no payload is
-        # built and nothing rides Tendermint consensus replication. The flag
-        # check lives here as well as in the post-settlement behaviour so
-        # the consensus-state cost is gated, not just the HTTP write.
-        # Build the predict-api event for both off-chain HTTP deliveries
-        # (is_offchain=True) and on-chain marketplace deliveries
-        # (is_offchain=False AND is_marketplace_mech). The ``source`` field
-        # on the event (set by ``_build_predict_api_event``) is what
-        # disambiguates the two paths on the predict-api side: ``mech_offchain``
-        # for the paid HTTP path, ``mech_onchain`` for the on-chain rails.
-        # See ``autonolas-marketplace/docs/onchain_write_path_scope.md`` for
+        # Gate the whole event build on ``use_offchain``. When the flag
+        # is off, no payload is built and nothing rides Tendermint
+        # consensus replication — so the ingress-side consensus cost
+        # matches the egress-side HTTP write cost (nothing on either
+        # side). The paired gate in
+        # :py:meth:`task_submission_abci.behaviours.PostTxSettlementBehaviour._do_predict_api_write_best_effort`
+        # additionally warns if it receives events under a locally-off
+        # flag, so a config drift between the two skill copies of
+        # ``use_offchain`` is alertable rather than silent.
+        # Build the event for both off-chain HTTP deliveries
+        # (``is_offchain=True``) and on-chain marketplace deliveries
+        # (``is_offchain=False`` AND ``is_marketplace_delivery=True``).
+        # The ``source`` field on the event (set by
+        # ``_build_predict_api_event``) disambiguates the two paths on
+        # the predict-api side: ``mech_offchain`` for the paid HTTP path,
+        # ``mech_onchain`` for the on-chain rails. See
+        # ``autonolas-marketplace/docs/onchain_write_path_scope.md`` for
         # the agreed shape.
         predict_api_mode_enabled = self.params.use_offchain
         is_marketplace_delivery = bool(done_task.get("is_marketplace_mech"))
--- a/packages/valory/skills/task_execution/models.py
+++ b/packages/valory/skills/task_execution/models.py
@@ -114,23 +114,16 @@ class Params(Model):
         )
         self.offchain_tx_list: List = list()
         self.default_chain_id: str = self._ensure_get("default_chain_id", kwargs, str)
-        # EIP-155 integer chain id for the wildcard data-lake write and the
+        # EIP-155 integer chain id for the predict-api data lake write and the
         # EIP-712 domain. Separate from ``default_chain_id`` (the AEA alias
-        # ``"gnosis"``) because the wildcard schema and the EIP-712 domain
+        # ``"gnosis"``) because the predict-api schema and the EIP-712 domain
         # bind to the integer, and converting at the boundary via a
         # hardcoded alias→int table couples task_execution to a chain
         # registry it has no other reason to know about. Defaults to 0,
-        # which fails the wildcard server's per-chain marketplace
+        # which fails the predict-api server's per-chain marketplace
         # allowlist on first POST — operator-friendly: the row is rejected
         # at a known boundary rather than silently mis-tagged.
         self.mech_events_chain_id: int = int(kwargs.get("mech_events_chain_id", 0) or 0)
-        # Feature flag for the wildcard analytics write path, mirrored from
-        # task_submission_abci so the build cost at finalize is also gated:
-        # when False (Phase 1 default) ``_finalize_done_task`` skips the
-        # ``_build_wildcard_event`` call entirely so nothing rides Tendermint
-        # consensus replication. Must agree with the task_submission_abci
-        # value across the deployment.
-        self.mech_events_enabled: bool = bool(kwargs.get("mech_events_enabled", False))
         self.gnosis_ledger_rpc: str = kwargs.get("gnosis_ledger_rpc", "")
         self.polygon_ledger_rpc: str = kwargs.get("polygon_ledger_rpc", "")
         self.base_ledger_rpc: str = kwargs.get("base_ledger_rpc", "")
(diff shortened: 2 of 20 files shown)


OjusWiZard (commented):
Multi-perspective review (correctness, tests, silent failures, type design, comments, simplification). No blocking issues — the rename and flag collapse are mechanically complete and consistent (zero stray `wildcard`/`mech_events_enabled` references, EIP-712 constants untouched, writer/reader dict keys match, config cascade coherent). Inline comments below: 3 important (test coverage of the gate, cross-skill flag divergence, stale docstring) and 3 advisory.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1884,36 +1884,36 @@ def async_act(self) -> Generator:
...
+            3. The round's done_tasks carry no offchain predict_api_event
                entries and the pending-task sweep produced no
                request-only events (a truly quiet round).
 
         :yield: AEA protocol messages (signing dialogue, HTTP request) via
             the underlying generator-based helpers; this method does not
             ``return`` anything.
         """
-        if not getattr(self.params, "mech_events_enabled", False):
+        if not getattr(self.params, "use_offchain", False):
```
OjusWiZard: **[Important — test coverage]** This gate — the PR's core semantic on the egress side — has no test in either direction. Nothing invokes `_do_predict_api_write_best_effort` or `_post_predict_api_batch` at all: there's no regression guard proving `use_offchain=False` prevents the external POST (the riskiest failure mode: silently sending settlement data to an external endpoint), no test that `use_offchain=True` + URL set actually attempts the POST, and no test of the empty-URL short-circuit two lines below. The `mech_predict_api_events_total` outcome labels are likewise unexercised. Suggest a mocked-HTTP pair: flag off → `_build_http_request_message` never called; flag on + URL + a `predict_api_event` in done_tasks → POST attempted and `outcome="ok"` incremented.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1533,30 +1533,30 @@ def _finalize_done_task(self, cid: str) -> None:
...
+        # disambiguates the two paths on the predict-api side: ``mech_offchain``
         # for the paid HTTP path, ``mech_onchain`` for the on-chain rails.
         # See ``autonolas-marketplace/docs/onchain_write_path_scope.md`` for
         # the agreed shape.
-        wildcard_mode_enabled = getattr(self.params, "mech_events_enabled", False)
+        predict_api_mode_enabled = getattr(self.params, "use_offchain", False)
         is_marketplace_delivery = bool(done_task.get("is_marketplace_mech"))
-        if wildcard_mode_enabled and (is_offchain or is_marketplace_delivery):
-            done_task["wildcard_event"] = self._build_wildcard_event(
+        if predict_api_mode_enabled and (is_offchain or is_marketplace_delivery):
```
OjusWiZard, line 1563: **[Important — test coverage]** The ingress-side gate is also untested in both directions: no test calls `_finalize_done_task` and asserts `done_task["predict_api_event"]` is present with `use_offchain=True` and **absent** with `False`. All existing `_build_predict_api_event` tests call the builder directly, bypassing this gate. Since the flag collapse is exactly the change that could flip this condition, a gate-on/gate-off pair (covering both an offchain-settled and an on-chain marketplace-settled task) would close the gap.

packages/valory/skills/task_submission_abci/models.py
```diff
@@ -123,33 +123,40 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        # Dual-purpose gate: controls BOTH offchain-request ingress
+        # (handled in task_execution) AND egress to the predict-api events
+        # endpoint from PostTxSettlement (settlement writes for both
+        # offchain-settled AND on-chain-settled requests). A mech is
+        # either part of the predict-api analytics pipeline or it isn't;
+        # there's no coherent state where ingress and egress diverge.
+        # False by default so an out-of-the-box mech does no off-mech
+        # HTTP work at settlement time and does not accept offchain
+        # requests.
+        self.use_offchain: bool = bool(kwargs.get("use_offchain", False))
```
OjusWiZard, line 145: **[Important — silent divergence]** "There's no coherent state where ingress and egress diverge" holds only by deployment discipline: `use_offchain` is two independent params in two skills' config blocks. If an operator overrides `task_execution.use_offchain=true` but not this one, events get built at consensus-replication cost and then dropped with **zero log output** — the early return in `_do_predict_api_write_best_effort` is silent. The analogous duplicated param `mech_events_chain_id` got a runtime warning safety net in the same file; suggest at least a one-time `warning` log when `predict_api_event` entries arrive in done_tasks while this flag is off.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2259,20 +2259,20 @@ def _post_wildcard_batch(  # noqa: C901
...
-        ride the same wildcard write path; the per-event ``source`` field
-        set inside :py:meth:`task_execution.behaviours._build_wildcard_event`
-        is what disambiguates the two on the wildcard side. The filter
-        here keys on the presence of a ``wildcard_event`` field — built
-        only when both ``mech_events_enabled`` is set AND the task is a
+        ride the same predict-api write path; the per-event ``source`` field
+        set inside :py:meth:`task_execution.behaviours._build_predict_api_event`
+        is what disambiguates the two on the predict-api side. The filter
+        here keys on the presence of a ``predict_api_event`` field — built
+        only when both ``use_offchain`` is set AND the task is a
```
OjusWiZard: **[Important — stale docstring]** "built only when both `use_offchain` is set AND the task is a marketplace task" is inaccurate: the actual condition in `_finalize_done_task` is `use_offchain AND (is_offchain OR is_marketplace_delivery)` — plain offchain HTTP tasks also produce events, not only marketplace ones. Pre-existing wording, but this PR edited the docstring and carried it forward; worth fixing while here.

packages/valory/skills/task_submission_abci/skill.yaml
```diff
@@ -170,9 +170,9 @@ models:
       serious_slash_unit_amount: 8000000000000000
       service_endpoint_base: https://dummy_service.autonolas.tech/
       default_chain_id: gnosis
-      mech_events_enabled: false
-      wildcard_events_url: ''
-      wildcard_events_timeout_seconds: 5.0
+      use_offchain: false
+      predict_api_events_url: https://mpp.autonolas.tech/mech/events
```
OjusWiZard, line 174: **[Advisory — default value change]** This isn't just a key rename: the default changes from `''` (no-op) to the live production endpoint. Safe today because `use_offchain` defaults to `false`, but flipping the bool alone now auto-targets `mpp.autonolas.tech` where the old default would short-circuit. If intentional (opt-in is the flag, not the URL), worth one line in the PR description so it doesn't read as a pure rename.

packages/valory/skills/task_submission_abci/predict_api_events_client.py
```diff
@@ -17,9 +17,9 @@
 #
 # ------------------------------------------------------------------------------
 
-r"""EIP-712 typed-data builder and batch-hash helper for the wildcard write client.
+r"""EIP-712 typed-data builder and batch-hash helper for the predict-api events client.
 
-The server side at ``wildcard/server/src/mech_sig.py`` (predict-api#162)
+The server side at ``server/src/mech_sig.py`` (predict-api#162)
```
OjusWiZard, line 22: **[Advisory — guard comment]** The EIP-712 constants below (`EVENTS_DOMAIN_NAME = "Olas Mech Events"`, `EVENTS_PRIMARY_TYPE = "MechEventBatch"`) deliberately keep the old branding, but the comment above them only explains the versioning discipline — not *why* they weren't renamed in this sweep. A future wildcard→predict_api-style cleanup is likely to "fix" them and silently break server signature verification. Suggest one explicit sentence: "Deliberately NOT renamed — the server's EIP-712 domain allowlist binds to these exact strings."

packages/valory/skills/task_execution/models.py
```diff
@@ -143,9 +136,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
-        # enabled per deployment in the Phase 2 rollout. False = today's
-        # on-chain + IPFS behaviour, unchanged.
+        # Dual-purpose gate: controls BOTH offchain-request ingress AND
+        # egress to the predict-api events endpoint (settlement writes for
+        # both offchain-settled AND on-chain-settled requests). A mech is
+        # either part of the predict-api analytics pipeline or it isn't;
+        # there's no coherent state where ingress and egress diverge.
+        # Default False = on-chain + IPFS only, no analytics writes,
+        # unchanged for legacy deployments. Set True to enrol the mech.
         self.use_offchain: bool = kwargs.get("use_offchain", False)
```
OjusWiZard: **[Advisory — consistency]** Two minor inconsistencies in the flag's own plumbing: (1) this line lacks the `bool(...)` coercion its twin in `task_submission_abci/models.py:135` has, so a non-boolean YAML override is silently truthy here; (2) the gate sites read the flag via `getattr(self.params, "use_offchain", False)` while `handlers.py:824` accesses the attribute directly — the string literal defeats mypy's rename checking, so a future typo/rename silently defaults to `False`. Both pre-existing style, but the PR touched every one of these lines and normalizing was in easy reach.

packages/valory/skills/task_execution/models.py
```diff
@@ -143,9 +136,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
             for key, value in kwargs.get("payment_type_to_asset_address", {}).items()
         }
-        # Phase 1 ships dark: the offchain HTTP path is disabled by default and
-        # enabled per deployment in the Phase 2 rollout. False = today's
-        # on-chain + IPFS behaviour, unchanged.
+        # Dual-purpose gate: controls BOTH offchain-request ingress AND
+        # egress to the predict-api events endpoint (settlement writes for
+        # both offchain-settled AND on-chain-settled requests). A mech is
+        # either part of the predict-api analytics pipeline or it isn't;
+        # there's no coherent state where ingress and egress diverge.
```
bennyjo: Is the plan to open offchain ingress on all 12 mechs seeded in predict-api#166? 8 of them have no `MECH_OFFCHAIN_URL` today, and with the single flag an on-chain-only mech can't be enrolled in analytics (`mech_onchain` writes) without also opening the offchain HTTP endpoint — so "no coherent state where ingress and egress diverge" doesn't hold for the on-chain write path. Happy to leave as-is if opening ingress everywhere is deliberate; flagging because otherwise most of the companion PR's seed list can never write.

packages/valory/skills/task_submission_abci/skill.yaml
bennyjo, line 174: Flipping this default from `''` to the prod URL removes the empty-URL short-circuit, so any deployment already running `use_offchain=true` (at least 4 mechs) silently starts POSTing settlement events to prod on its next upgrade — and a `use_offchain=true` mech not on the server allowlist will sign+POST+403 every settlement round. If upgrade-as-enrolment is the intended rollout, worth saying so in the PR body; otherwise consider keeping the default empty.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -125,14 +125,14 @@
...
-# rounds), so a transient wildcard failure silently discards the row —
+# rounds), so a transient predict-api failure silently discards the row —
 # the replay buffer that would recover it is a future PR. This counter
 # gives operators a signal before the buffer lands: sustained non-``ok``
 # on ``batch_label="delivered"`` means the analytics lake is missing
 # rows for that window.
-mech_wildcard_events_total = Counter(
-    "mech_wildcard_events_total",
-    "Wildcard analytics-write batch outcomes",
+mech_predict_api_events_total = Counter(
```
bennyjo, line 133: Renaming `mech_wildcard_events_total` → `mech_predict_api_events_total` will make any dashboard/alert keyed on the old counter go silent rather than error — worth checking Grafana for consumers before merge.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1533,30 +1533,30 @@ def _finalize_done_task(self, cid: str) -> None:
...
         # IN_MEMORY_REQUESTS pop below; the post-settlement behaviour in
         # task_submission_abci reads it off ``synchronized_data.done_tasks``
-        # (i.e. after consensus replication) and posts it to the wildcard
+        # (i.e. after consensus replication) and posts it to the predict-api
         # data lake, but the buffered request metadata it needs is local
         # to this agent's shared_state and goes away at the pop.
         #
-        # Gated on BOTH ``is_offchain`` AND the ``mech_events_enabled`` flag
+        # Gated on BOTH ``is_offchain`` AND the ``use_offchain`` flag
         # so Phase 1 is genuinely dark: when the flag is off, no payload is
```
bennyjo: This comment block still describes the two-flag world ("Phase 1 is genuinely dark", flag check mirrored from task_submission_abci) — worth updating to the new single-flag semantics since this is exactly the gate the PR changes.

bennyjo (commented):
Rename verified clean (no stale refs, EIP-712 constants untouched, hashes re-locked). Main question is the flag-collapse semantics — see inline.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: You're right, the biggest silent-failure surface in the PR had no test. Added in 60e302ee: `test_egress_gate_off_short_circuits_without_http_call` (flag off + otherwise-quiet done_tasks → `_build_http_request_message` never called), `test_egress_gate_on_with_empty_url_short_circuits_at_debug_level` (opt-in flag on, URL missing → DEBUG log, no HTTP call), and the divergence-warning pair below. Uses a `_build_http_request_message` spy on a SimpleNamespace so the assertion is direct rather than mock-count arithmetic.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fair, the ingress side had the same shape gap. Added three tests in 60e302ee mirroring the actual gate condition (`use_offchain AND (is_offchain OR is_marketplace_delivery)`): `test_finalize_done_task_attaches_predict_api_event_when_use_offchain_on` for the off-chain HTTP shape (source=mech_offchain), `test_finalize_done_task_attaches_predict_api_event_on_marketplace_delivery` for the on-chain marketplace shape (source=mech_onchain), and `test_finalize_done_task_omits_predict_api_event_when_use_offchain_off` for the Phase-1 dark-ship guarantee. All three call `behaviour._finalize_done_task(cid)` end-to-end rather than the direct-call pattern the other tests used, so a future flag flip changes the assertion outcome instead of the tests staying quiet.

packages/valory/skills/task_submission_abci/models.py
DIvyaNautiyal07: Good catch on the silent-drop side. Added a WARNING at the egress gate in 60e302ee: when `use_offchain=False` fires but any `done_task` in the round carries a `predict_api_event`, we log `predict_api_event entries present in done_tasks but task_submission_abci.use_offchain is False — the ingress-side (task_execution) and egress-side (task_submission_abci) use_offchain flags are diverging. Dropping the batch. Align both skill configs to eliminate the ingress work.` `test_egress_gate_off_with_diverged_ingress_emits_divergence_warning` pins it, and `test_egress_gate_off_with_no_events_stays_silent` pairs it so the warning doesn't itself become boot noise on production mechs that ship dark.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Stale, agreed. Fixed in 60e302ee. Docstring now describes the actual condition: `use_offchain` AND (`is_offchain` OR `is_marketplace_delivery`), so on-chain non-marketplace tasks stay out of the lake by construction while off-chain HTTP + on-chain marketplace both feed it via the per-event `source` field.

packages/valory/skills/task_submission_abci/skill.yaml
DIvyaNautiyal07: Fair, that isn't a pure rename. Updated the PR body — added a Behaviour-change note calling out that (a) the default flipped from `''` (short-circuit) to `https://mpp.autonolas.tech/mech/events`, (b) the opt-in is still `use_offchain`, so the URL only takes effect when the flag is flipped, (c) legacy deployments upgrading with `use_offchain=false` see zero change. If we ever want to flip that pairing, we can gate on the URL being non-empty as the opt-in signal separately, but that would need its own review.

packages/valory/skills/task_submission_abci/predict_api_events_client.py
DIvyaNautiyal07: Real risk — a future cleanup sweep is exactly the shape that would silently break server-side signature verification. Added an explicit comment in 60e302ee above the constants: `Deliberately NOT renamed in the wildcard→predict_api sweep — the server's EIP-712 domain allowlist binds to these exact strings, and a rename here silently breaks every server-side signature verification with no test failure. If a future rename is genuinely wanted, ship the string change alongside a coordinated server-side allowlist bump.` Reads as intent, not a lecture, so a future reader gets the guard without feeling condescended to.

packages/valory/skills/task_execution/models.py
DIvyaNautiyal07: Both fixed in 60e302ee. (1) Added `bool(...)` coercion on `task_execution/models.py:146` so it matches its twin in `task_submission_abci/models.py:135` — a non-bool YAML override no longer silently truthies here. (2) The three `getattr(self.params, "use_offchain", False)` sites in `task_submission_abci/behaviours.py` and `task_execution/behaviours.py` switched to direct `self.params.use_offchain` access, matching what `handlers.py:824` already does; a future rename now fails mypy loudly instead of silently defaulting to `False`.

packages/valory/skills/task_execution/models.py
DIvyaNautiyal07: Deliberate, and documented explicitly in 2160d0cb. Params docstring on `task_execution/models.py` now names the coupling: the predict-api pipeline is a set — a mech is either enrolled (ingress + egress) or not. An on-chain-only mech that wants `mech_onchain` analytics writes still flips `use_offchain=true` and accepts that its offchain HTTP handler is also live. The cost is a wire nothing rides (no client's `MECH_OFFCHAIN_URL` targets an on-chain-only mech), not a traffic-shape change. Splitting egress back out is a genuine architectural question — happy to have that in a follow-up if the seed-list operators actually push back on the coupling, but that's a real re-design rather than an env-override on the current shape.

packages/valory/skills/task_submission_abci/skill.yaml
DIvyaNautiyal07: Handling this operationally rather than reverting the default. The 4 mechs currently running `use_offchain=true` will have the flag set back to `false` in agent-deployments before the redeploy that picks up this change, so they explicitly re-enrol via config rather than silently upgrade into POSTing. Called that out in the PR body in the new deployment-coordination note under Test plan. If the operational side slips and someone redeploys the old set without the flag flip, we get 403s per settlement round rather than data-integrity issues — noisy, but self-diagnosing.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Checked and there's no Grafana dashboard or alert rule keyed on `mech_wildcard_events_total` at time of merge, so no dual-emit needed here. Called out in the PR body's new Prometheus-metric-rename note — straight rename with no label-shape change, so if a consumer surfaces later, migrating the query is a one-line find-and-replace on the metric name.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in 2160d0cb. Refreshed the comment block: no more `Phase 1 is genuinely dark` or `flag check mirrored from task_submission_abci` framing. New wording describes the single-flag semantics and cross-references the paired silent-divergence WARNING gate in `_do_predict_api_write_best_effort`, so a future reader sees the two sites are intentionally coupled and knows where to look if they change one.

packages/valory/skills/task_submission_abci/models.py
```diff
@@ -123,33 +123,40 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
-        self.mech_events_enabled: bool = bool(kwargs.get("mech_events_enabled", False))
-        # Wildcard ``POST /mech/events`` endpoint. Empty string = no URL
-        # configured; the PostTxSettlement behaviour short-circuits when
-        # either this is empty or ``mech_events_enabled`` is False. No
+        # Dual-purpose gate: controls BOTH offchain-request ingress
+        # (handled in task_execution) AND egress to the predict-api events
+        # endpoint from PostTxSettlement (settlement writes for both
+        # offchain-settled AND on-chain-settled requests). A mech is
+        # either part of the predict-api analytics pipeline or it isn't;
+        # there's no coherent state where ingress and egress diverge.
```
OjusWiZard: **[nit — contradictory comment]** This copy still claims "there's no coherent state where ingress and egress diverge" — which now contradicts the divergence safety net added in `_do_predict_api_write_best_effort` (it exists precisely because the two flags *can* diverge). The sibling comment in `task_execution/models.py` was rewritten to the nuanced wording in the same commit; only this copy was missed. Suggest aligning it with the task_execution version.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -3495,8 +3495,184 @@ def test_build_wildcard_event_marketplace_task_falls_back_to_requester_key(
...
+    """Wire the minimum ``behaviour`` state for a ``_finalize_done_task`` call.
+
+    Seeds ``_executing_task`` / ``_done_task`` in the same shape the
+    happy-path ``_handle_store_response`` test does, plus
+    ``IN_MEMORY_REQUESTS`` / ``OFFCHAIN_REQUEST_RESPONSES`` so
+    ``_build_predict_api_event`` can produce a real event when the gate
+    lets it. ``to_multihash`` is monkeypatched to a cheap prefix so
+    ``_finalize_done_task`` doesn't try to run the real multihash logic.
+
+    :param behaviour: FetchDataBehaviour fixture.
```
OjusWiZard: **[nit — phantom class in docstrings]** `FetchDataBehaviour` doesn't exist anywhere in this codebase — the `behaviour` fixture (conftest.py) is a `TaskExecutionBehaviour`. Copy-paste artifact repeated in all four new docstrings: here, 3594, 3628, and 3661. s/FetchDataBehaviour/TaskExecutionBehaviour/.

OjusWiZard (dismissed):
Second-pass review of the fix commits (60e302ee, b01cea44, 2160d0cb): all first-round findings verified addressed — gate tests land on both sides, the divergence warning is exception-safe (`done_tasks` defaults to `[]`, non-dict entries skipped, unconditional early return), the direct-attribute refactor can't `AttributeError` since both `Params.__init__` set the flag unconditionally, and all five re-locked package hashes match `packages/packages.json`. Approving with two doc-level nits inline; the remaining known gap (no happy-path test for `_post_predict_api_batch` / `outcome="ok"`) is fine as a follow-up.

packages/valory/skills/task_submission_abci/models.py
DIvyaNautiyal07: Fixed in bf023754. The comment on `task_submission_abci/models.py:126-144` now matches the rewritten sibling in `task_execution/models.py` — coupling framed as deliberate, but with an explicit note that env-overrides *can* make the two skill copies drift and a cross-reference to the `_do_predict_api_write_best_effort` WARNING that surfaces that drift. Thanks for the catch.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Fixed in bf023754. Renamed `FetchDataBehaviour` → `TaskExecutionBehaviour` in the four docstrings at lines 3534, 3594, 3628, 3661. Copy-paste artifact you nailed — the `behaviour` fixture from `conftest.py` is a `TaskExecutionBehaviour`.

packages/valory/skills/task_submission_abci/models.py
```diff
@@ -123,33 +123,50 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
-        # either this is empty or ``mech_events_enabled`` is False. No
+        # Dual-purpose gate: controls BOTH offchain-request ingress
+        # (handled in task_execution) AND egress to the predict-api events
+        # endpoint from PostTxSettlement (settlement writes for both
+        # offchain-settled AND on-chain-settled requests). A mech is
+        # either part of the predict-api analytics pipeline or it isn't;
+        # the coupling is deliberate. On-chain-only mechs that want
+        # analytics writes (``mech_onchain`` source) still flip this on
+        # and accept that their offchain HTTP handler is also live — the
+        # handler is inert if no client's ``MECH_OFFCHAIN_URL`` points
```
OjusWiZard: **[nit — unverifiable name]** `MECH_OFFCHAIN_URL` doesn't exist anywhere in this repo outside this comment and its twin in `task_execution/models.py:146` — it's not an env var, client setting, or anything a reader of this codebase can confirm. The underlying claim is right (the enabled-but-unused handler is a no-op), so either drop the name ("inert unless a client actually sends offchain requests to it") or, if it's a real variable in a downstream repo (mech-client / trader?), point at where it lives.

OjusWiZard (dismissed):
Third pass over bf023754: hash cascade verified consistent via AEA's package manager (both skills → mech_abci → agent → service → packages.json, zero stale references), doc edits introduce no black/darglint regressions, and the rewritten comment's behavioural claims all check out against the code. One nit inline about an unverifiable config name. Approval stands.

packages/valory/skills/task_submission_abci/models.py
DIvyaNautiyal07: Good catch. Reworded to "the handler is inert unless a client actually sends offchain requests to it" in both places (9b09e65b), so a reader of this codebase can verify the claim without a cross-repo lookup. `MECH_OFFCHAIN_URL` does exist — it's the env var operators set in `mech-deployments/deployments/*.env` that becomes the `url` field in the mech's on-chain metadata JSON — but since none of that surfaces in this repo, referencing it here was more confusing than helpful.

OjusWiZard (approved):

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2053,7 +2074,7 @@ def _post_wildcard_batch(  # noqa: C901
             for e in events
         ):
             self.context.logger.warning(
-                "Wildcard %s batch contains events from multiple chains; "
+                "predict-api %s batch contains events from multiple chains; "
```
bennyjo, line 2077: The chain-id parsing just above runs outside the `try` below, so one malformed event in `done_tasks` crashes the agent instead of being logged and dropped. Pre-existing, but this PR makes the path far more reachable — suggest moving the parse inside the `try` (fine as a follow-up).

bennyjo (commented):
Re-review: all first-round findings verified fixed. One remaining item inline.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fair point on reachability. Not blocking here because the `chain_id` value we parse is one we set ourselves in `_build_predict_api_event` / `_build_request_only_event` from the `mech_events_chain_id` param (already an int at param-load time), not something a requester threads in. `int(...)` would only raise on a non-numeric string, and we never write one there. Happy to fold the parse into the `try` in a follow-up so it's defence-in-depth rather than relying on the invariant.

bennyjo (approved):

PR #475, opened 2026-08-11
feat(offchain): verify signatures at accept time (fail fast, cache Safe owners)

DIvyaNautiyal07:
## Summary

The offchain request handler in `task_execution` currently accepts a signed request, checks the requester's prepaid balance on the BalanceTracker, and enqueues the task. It never verifies the signature. Bad signatures only surface later at settlement time when the mech itself pays gas to submit `deliverMarketplaceWithSignatures` and Gnosis Safe reverts with `GS026 Invalid owner`. By then the tool has already run, compute + API costs are eaten, and the mech has no way to recover them.

Real production impact: on one Autonolas Mech III (id 2340) log window of 12 hours, a single trader Safe with a client-side signing bug posted 1626 offchain requests. The mech served every one of them, tried to settle 188 of them on chain in batches, all 188 reverted with GS026, `mapRequestCounts[safe]` stayed pinned at 264, and the mech ate the cost of ~1600 unpaid jobs. This PR closes that gap so the very first bad-sig request gets rejected with a 401 before the tool ever runs.

## Design

Verify the signature in `_handle_signed_requests` before the balance check, with a hard constraint of "no new per-request latency on the accept path". Achieved with process-level caching:

* **Domain separator**, cached per `(chain_id, marketplace_address)`. Never rotates unless the marketplace redeploys.
* **Mech `paymentType`**, cached per mech. Constant for a deployment.
* **`Safe.getOwners()` + `getThreshold()`**, cached per requester in an LRU with TTL. Rare enough that a fresh sender pays one round trip to fetch owners; every subsequent request from that sender is pure CPU.

Verification path per request:
1. Locally re-derive the marketplace `request_id` from the posted fields using the same formula as `MechMarketplace.getRequestId` (all inputs known, no RPC needed on the hot path).
2. If the sender is a contract (Safe), locally wrap in `SafeMessage(bytes message)` EIP-712 struct. If EOA, skip the wrap.
3. Local `ecrecover` → recovered signer.
4. For EOA senders, compare against the sender address. For Safe threshold=1, compare against the cached owner set. For Safe threshold>1, fall back to `Safe.isValidSignature` RPC (rare for agent Safes).
5. On mismatch, return `401 Unauthorized` with a clear reason and skip enqueue entirely.

Cache-hit `_verify_offchain_signature` cost measured on this diff:
- With `coincurve` (libsecp256k1): ~154µs EOA, ~238µs Safe threshold=1
- Pure-Python `eth_keys`: ~5.6ms
- Recommend `coincurve` in the deployment image if not already present

Wire format is unchanged. The trader already POSTs `signature` and `nonce` for the settlement path — the server just wasn't reading them.

## Test plan

- [x] 23 new tests in `test_signature_verifier.py`:
  - Hash-math regression against a hand-computed EIP-712 digest
  - EOA valid / invalid
  - Safe threshold=1 valid / non-owner
  - Safe threshold>1 RPC fallback (valid / invalid / missing)
  - Cache TTL expiry and LRU eviction
  - Cache-hit RPC accounting (100 requests from same sender → 1 code_fetcher call, 0 safe_meta_fetcher calls)
  - Handler-level 401 short-circuit before balance check
  - Handler-level 200 fall-through when verifier ok
  - Missing-signature 401
  - `paymentType` cached once per mech
- [x] Pre-existing handler tests untouched behaviourally: `_install_verify_ok` helper threaded into pre-existing setups so the balance/enqueue path stays exercised
- [x] Full `pytest packages/valory/skills/task_execution/tests/` → **389 passed** (previous suite + 23 new)
- [x] `autonomy packages lock --check` will pass (fingerprints refreshed in a follow-up commit on this branch)

## Notes for reviewer

- Files edited under `packages/valory/skills/task_execution/...`, which is the versioned source. `agent/vendor/...` is regenerated build output from `packages/` and is not touched here.
- No change to how the trader talks to the mech (POST body, form encoding, response shape). Only accept-time validation was added.
- If the deployment image doesn't have `coincurve`, verify cost drops to ~5ms per request, which is still an order of magnitude better than a per-request RPC and still small compared to tool execution time. Adding `coincurve` to the deployment image is a separate, optional ops change.

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -9,12 +9,12 @@
         "connection/valory/websocket_client/0.1.0": "bafybeid4cnbjpmzcsw3usg7umufgqyod2u3xfw5uso3vukt3n5hfuxvdea",
         "skill/valory/contract_subscription/0.1.0": "bafybeicqlkmfpttw4hbbaakr6xos62rd6ytjwkyzlfli4xpvl5yqcycf6q",
         "skill/valory/delivery_rate_abci/0.1.0": "bafybeiathwdfpcfqbipqbkslv6j22q5cvov4n3czmehzij3czrhdnlg42e",
-        "skill/valory/mech_abci/0.1.0": "bafybeiatph3obd4vomjsft7civq7iukue5c3q63t44cghvrfghca65aoma",
-        "skill/valory/task_execution/0.1.0": "bafybeiblzdymhr6gps5be7eemrgaqwu45hzmd5rsugpkldutjumwl2gkqm",
-        "skill/valory/task_submission_abci/0.1.0": "bafybeicm4kt4zaznkb3vtdl3wz574j5xpqpo4gn2zwpaezpt4tqx2cj36a",
+        "skill/valory/mech_abci/0.1.0": "bafybeie53bioafwoqmtxmjknrg47v6sk5n74hnulny5pmuxxjfjuqi6use",
+        "skill/valory/task_execution/0.1.0": "bafybeidfpmrhs2lsw5hczub3opi4wnlw4zy3ekm6qd5fzxpk63266aesdu",
+        "skill/valory/task_submission_abci/0.1.0": "bafybeiaczm2mlojowbfsf5nqdxebif5az4mytwhfa5qtpyujlj5p6ramta",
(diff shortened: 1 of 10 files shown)


DIvyaNautiyal07:
Redoing this with the correct design (sequential verify+balance on accept path, no caching, contract wrapper pattern). New PR incoming.

PR #477, opened 2026-08-12
fix(offchain): 402 challenge reads asset from balance tracker's token()

DIvyaNautiyal07:
## Summary

- The 402 challenge's `asset` field is now read from the balance tracker's on-chain `token()` getter, cached per tracker address for the process lifetime.
- Removes the `payment_type_to_asset_address` map primitive that duplicated the same information as an operator-managed env config.

## Background

While bench-testing Pearl 1.9.2-rc4 with Polystrat on Polygon, mech-interact repeatedly failed offchain requests with:

```
Offchain 402 asset mismatch: mech reported '0x0000000000000000000000000000000000000000'
but BalanceTracker 0x5c50ebc17d002a4484585c8fbf62f51953493c0b accepts
'0x3c499c542cef5e3811e1192ce70d8cc03d5c3359'. Refusing to deposit and failing over.
```

The trader's asset-match guard (from mech-interact PR #95) fired correctly and prevented a wrong-asset deposit. Root cause: the mech's 402 builder looked up `paymentType → asset_address` in `Params.payment_type_to_asset_address`, a map that:

1. Had no env-override wiring in the mech agent / service yaml or in any consumer repo (mech-predict, mech-agents-fun). `PAYMENT_TYPE_TO_ASSET_ADDRESS='...'` in a `.env` was a no-op.
2. Fell back to `ZERO_ADDRESS` (native) when the paymentType was missing, which is only correct for native trackers.
3. Would have required 6 yaml edits across 3 repos + per-chain `.env` entries per mech to make usable.

## The fix

`_check_offchain_requester_balance` now calls `_resolve_balance_tracker_asset(ledger_api, tracker_addr)` and adds the result to the response dict under `ResponseKey.ASSET_ADDRESS`. The resolver:

- Reads `BalanceTrackerContract.get_token_address()` (the `token()` view).
- Caches in `_balance_tracker_asset_cache: Dict[str, str]` keyed by tracker address.
- Returns `ZERO_ADDRESS` on `ContractLogicError` (native trackers revert on `token()` — that's their identity).
- Returns `ZERO_ADDRESS` with a WARNING log on any other error so the 402 still emits and mech-interact's asset-match guard catches the mismatch downstream.

`_build_402_challenge` reads `asset` from `balance_check[ASSET_ADDRESS]`, no more Params lookup.

The `payment_type_to_asset_address` field is deleted from `models.Params`, from `skill.yaml`'s default block, and from the test `params_stub` fixture.

## Why auto-derive over env config

`BalanceTrackerFixedPriceToken.token` is declared `immutable` in Solidity — the value baked into runtime bytecode never changes. A re-registered tracker (rare) has a different address, which is a fresh cache key. So one RPC per unique balance tracker per process lifetime, on a code path that already makes three sequential balance-check RPCs. Effectively free. Removes an entire config-drift class of bug.

## Test plan

- [x] `pytest packages/valory/skills/task_execution/tests packages/valory/skills/mech_abci/tests` — 563 pass (+3 new tests: `test_resolve_balance_tracker_asset_caches_per_address`, `test_resolve_balance_tracker_asset_native_tracker_returns_zero_address`, `test_build_402_challenge_forwards_asset_address_from_balance_check`; renamed one)
- [x] `tomte tox -e black-check -e isort-check -e flake8 -e mypy -e darglint -e check-hash -e check-packages -e check-dependencies -e check-abciapp-specs` — all green
- [x] Third-party package hashes unchanged
- [ ] Re-test Polystrat offchain flow on Polygon post-merge (requires trader with insufficient prepaid balance so the 402 branch fires end-to-end)

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -9,12 +9,12 @@
         "connection/valory/websocket_client/0.1.0": "bafybeid4cnbjpmzcsw3usg7umufgqyod2u3xfw5uso3vukt3n5hfuxvdea",
         "skill/valory/contract_subscription/0.1.0": "bafybeiey3zrdef6qcspojfbi2qmzc7cdxrwljmu6cnfxqr7okwfimyjrey",
         "skill/valory/delivery_rate_abci/0.1.0": "bafybeifjiazvf3qhr2264i55u4tlvmyf6ztgak2ndqolujt4kw6jiy7l3m",
-        "skill/valory/mech_abci/0.1.0": "bafybeicaohpjhxmliqepo7zxysz2odhek2imvpmsdgp76sl6tltexui4ly",
-        "skill/valory/task_execution/0.1.0": "bafybeifgai2d7qkysxjnein3meyjdsklgt57rfjei3v5p2acxfqxuhavli",
-        "skill/valory/task_submission_abci/0.1.0": "bafybeia65wzj3vjrtdqkq55lsgvmju7hq4rz4jhau3b65wmg7tnj6fr6oy",
+        "skill/valory/mech_abci/0.1.0": "bafybeicdhperwghzemntlzc3evrpnlvbuzjw3bmrsstmpfiwynwyxn7bz4",
+        "skill/valory/task_execution/0.1.0": "bafybeiduw74ltjr6utxsb3ojfyxmgtght2x7yxovvcm7cmuaj7dlrgsxfy",
+        "skill/valory/task_submission_abci/0.1.0": "bafybeie4pmlwxrvhj3icrmodoiucjx3cqizdeo7v7wnrc56c2v7cs5tk6a",
         "skill/valory/websocket_client/0.1.0": "bafybeiht2ify5vkseeulutzjwkw5vcmhcwtun2o5ehn6hya6uz4eanua2m",
-        "agent/valory/mech/0.1.0": "bafybeig4kfzh7cfdta245ayb3hc62tstu5d2y3rtpxhvibhzkhmj3vbk6y",
-        "service/valory/mech/0.1.0": "bafybeifnswe2nfcpvnelrgfbnfpokhb2f7eeu5v4dcnskj32uuvmafvycq"
+        "agent/valory/mech/0.1.0": "bafybeibt7dq5ux6h4j55jp25r7fuy7tbx4c3z6mwcjxqdjbiegijtzpyey",
+        "service/valory/mech/0.1.0": "bafybeick5dfyj3irgdbkovikudfg3aedoetlmnh3g6pnzz4riffxbeswra"
     },
     "third_party": {
         "protocol/open_aea/signing/1.0.0": "bafybeifsjmldwyki3beqyvdt5lzenrg6wyrqaar5plc5rpnvtc4zlentye",
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -9,12 +9,12 @@
         "connection/valory/websocket_client/0.1.0": "bafybeid4cnbjpmzcsw3usg7umufgqyod2u3xfw5uso3vukt3n5hfuxvdea",
         "skill/valory/contract_subscription/0.1.0": "bafybeiey3zrdef6qcspojfbi2qmzc7cdxrwljmu6cnfxqr7okwfimyjrey",
         "skill/valory/delivery_rate_abci/0.1.0": "bafybeifjiazvf3qhr2264i55u4tlvmyf6ztgak2ndqolujt4kw6jiy7l3m",
-        "skill/valory/mech_abci/0.1.0": "bafybeigfubl4qqto3oimy4tsxclg3wpgn3ixee7veietoiifhb5sgkzidm",
-        "skill/valory/task_execution/0.1.0": "bafybeie63xevn4p7phvpswadq7lbsr6qvtxkhdmcmoyqtdlzvjvcpfjdza",
-        "skill/valory/task_submission_abci/0.1.0": "bafybeiesn6mgvjl53lbqigqoq3z2n2f5f6siyv6g445x6c2muu4dgsfdve",
+        "skill/valory/mech_abci/0.1.0": "bafybeicaohpjhxmliqepo7zxysz2odhek2imvpmsdgp76sl6tltexui4ly",
+        "skill/valory/task_execution/0.1.0": "bafybeifgai2d7qkysxjnein3meyjdsklgt57rfjei3v5p2acxfqxuhavli",
+        "skill/valory/task_submission_abci/0.1.0": "bafybeia65wzj3vjrtdqkq55lsgvmju7hq4rz4jhau3b65wmg7tnj6fr6oy",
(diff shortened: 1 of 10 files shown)


OjusWiZard (dismissed):
Approving — the design is right. Deriving `asset` from the tracker's `token()` instead of an unwired config map removes a real config-drift class, and the `immutable` argument for caching per tracker address is sound.

One critical issue to fix before this ships, though.

### Transient RPC failures permanently poison the asset cache

`packages/valory/skills/task_execution/handlers.py:3005-3026`

```python
except ContractLogicError:
    asset = ZERO_ADDRESS          # deterministic: native tracker identity
except Exception:
    self.context.logger.warning(...)
    asset = ZERO_ADDRESS          # transient: timeout, saturation, transport
self._balance_tracker_asset_cache[balance_tracker_address] = asset   # <-- unconditional
return asset
```

The cache write sits outside the `try`, so both branches persist. `_run_with_wall_clock_deadline` raises `concurrent.futures.TimeoutError` when the RPC exceeds `_BALANCE_RPC_DEADLINE_SECONDS = 2.0` (handlers.py:265) and `RuntimeError(RPC_QUEUE_SATURATED)` under executor queue pressure (handlers.py:1430) — a load condition this file explicitly designs for. Neither is a `ContractLogicError`; both land in the broad branch.

So a single 2-second RPC blip during the first resolution for an ERC-20 tracker writes `ZERO_ADDRESS` into a cache with no TTL and no invalidation path. Every subsequent 402 for that tracker then reports `asset: 0x000…0` — "native, skip the ERC20 approve" — until the process restarts. That's the exact Polystrat/Polygon symptom quoted in the PR description, now reachable from a transient blip rather than a config gap. And because the 402 body is persisted and re-served via the polling endpoint (`record_response=True`, handlers.py:2066), the wrong asset outlives the response itself.

This is also the only RPC-backed helper in the file that persists a fallback on transient failure. `_get_requester_balance` (handlers.py:2963) catches nothing locally and lets infra errors reach the outer handler at 2909, which returns a non-cached "unavailable". The `mapNonces` and EIP-1271 paths both separate infra from deterministic and never persist the infra verdict. Line 1652 states the principle directly: *"leave the constants loaded so a transient RPC blip does not permanently disable the endpoint."*

**Suggested fix:** cache only on the deterministic branches; return the fallback uncached on transient errors so the next request retries.

```python
try:
    token_res = self._run_with_wall_clock_deadline(...)
    asset = cast(str, token_res.get("token_address") or ZERO_ADDRESS)
except (ContractLogicError, BadFunctionCallOutput):
    # No ``token()`` on this contract — a native tracker. Deterministic,
    # so it is safe to cache.
    asset = ZERO_ADDRESS
except Exception:  # pylint: disable=broad-exception-caught
    # Timeout / queue saturation / transport. We do not know the asset;
    # fall back to native for this reply but do NOT cache the guess.
    self.context.logger.warning(
        "Failed to resolve balance tracker asset for %s; falling back to "
        "native for this reply without caching.",
        balance_tracker_address,
        exc_info=True,
    )
    return ZERO_ADDRESS
self._balance_tracker_asset_cache[balance_tracker_address] = asset
return asset
```

Note the coupling: `BadFunctionCallOutput` needs to be in the deterministic set alongside `ContractLogicError`. The ABI here is `BalanceTrackerFixedPriceToken`, so against a native tracker there is no `token()` selector, and whether web3 surfaces that as `ContractLogicError` or `BadFunctionCallOutput` depends on whether the provider returns an error object or a bare `0x`. If only `ContractLogicError` counts as deterministic, native trackers on the other kind of provider would re-RPC on every 402 forever. This file already pairs the two everywhere it handles this class (lines 1663 and 3665).

### Two smaller notes

- The wiring line `ResponseKey.ASSET_ADDRESS.value: asset_address` (handlers.py:2906) — arguably the crux of the PR — has no test coverage. Deleting it leaves all 185 tests in `test_handlers.py` green. `test_check_offchain_requester_balance_success` (test_handlers.py:1781-1811) already exercises the real path; adding an assertion on `result[ResponseKey.ASSET_ADDRESS.value]` closes the gap in one line.
- The generic-exception branch has no test. One that raises `TimeoutError` from `get_token_address` and asserts the value is *not* cached would pin the fix above.

Everything else checks out: the cache key safely omits `chain_id` (`_get_ledger_settings` reads the fixed `params.default_chain_id`), `balance_tracker_address` comes from the marketplace rather than the request so the cache is not attacker-growable, no `payment_type_to_asset_address` references remain anywhere in the repo, and the package hashes are consistent across all five yamls.

DIvyaNautiyal07:
Addressed in 6b2a05ef.

**Critical (cache-poison on transient RPC failure).** Moved the cache write inside the deterministic-only branches: `ContractLogicError` and `BadFunctionCallOutput` (native-tracker identity) cache as `ZERO_ADDRESS`; the broad `except Exception` now returns `ZERO_ADDRESS` for this reply but does **not** write the cache, so the next 402 retries. Pinned by `test_resolve_balance_tracker_asset_transient_failure_is_not_cached` which raises `concurrent.futures.TimeoutError` from the first call, asserts the cache stays empty, then the second call succeeds and populates.

**BadFunctionCallOutput pairing.** Now sits alongside `ContractLogicError` in the deterministic tuple. Native-tracker parametrize test covers both shapes.

**Note 1 (wiring coverage).** Added `assert result[ResponseKey.ASSET_ADDRESS.value] == "0xUSDCToken"` in `test_check_offchain_requester_balance_success` — the crux line at handlers.py:2906 is now covered end-to-end.

**Note 2 (transient-branch test).** Same test as above covers it.

565 tests pass (+2 from the parametrize expansion + no-cache test), full CI battery green locally.

OjusWiZard (approved):
Re-approving — verified against `6b2a05ef7`.

The critical comment is fixed as suggested: the transient branch now returns `ZERO_ADDRESS` directly instead of falling through to the cache write, so only deterministic outcomes persist and the next request retries the read. `BadFunctionCallOutput` cache-pairs with `ContractLogicError`, which closes the coupling I flagged — a native tracker on a bare-`0x` provider caches its identity instead of re-RPCing on every 402.

Both coverage gaps are closed too. `test_resolve_balance_tracker_asset_transient_failure_is_not_cached` pins the no-cache behaviour, and the new `ASSET_ADDRESS` assertion in `test_check_offchain_requester_balance_success` means deleting the wiring line at `handlers.py:2906` now fails the suite rather than passing silently — that was the mutation gap.

Checked locally: 504 pass in `packages/valory/skills/task_execution/tests`, and the fingerprints are re-locked across `packages.json` and all four dependent yamls. CI is green on 22/22.

atepem (approved):

bennyjo (approved):
Verified against the diff: deriving the 402 asset from the tracker's immutable token() removes a config primitive that was never wired and whose native fallback was only accidentally correct on Gnosis; deterministic outcomes cache per tracker, transient failures fall back uncached so they self-heal, and a wrong native fallback can only trip the client-side asset guard into a refusal, never a wrong-asset deposit. One non-blocking note: token_res.get("token_address") or ZERO_ADDRESS sits on the caching success path, so a wrapper that ever returned a dict without that key would latch native for a token tracker until restart; indexing instead would raise into the uncached fallback and self-heal.

PR #478, opened 2026-08-13
fix(task_execution): populate in_memory_requests on the on-chain path

DIvyaNautiyal07:
## Summary

The mech's `_build_predict_api_event` (behaviours.py:1822) reads request-side content — `prompt`, `raw_content`, `model`, `tool_params` — from `in_memory_requests`. That cache is populated at accept time by the off-chain handler (`handlers._enqueue_offchain_request:2787`). The on-chain path had no equivalent hook, so the cache lookup always returned empty for on-chain requests and the payload builder fell back to `prompt='[offchain request]'` and `raw_content={}`.

## Impact

Reproduced end-to-end against production predict-api on 2026-08-13. **100% of `source='mech_onchain'` rows carry the placeholder shape:**
- Gnosis (chain 100): 2,594 / 2,594 rows
- Polygon (chain 137): 107 / 107 rows
- Spread across every requester (many different Safes including trader agents), so it isn't a per-client issue.

Downstream: mech-analytics's ETL correctly extracts NULL for `question_title` and `market_id` from placeholder content, which cascades into the olas-website parity check failing row_parity by 1,446 rows on a 2-day recent window. The pre-cutover parity artifact ran cleanly only because the tested window predated the on-chain live poller (100% `ipfs_historical` source).

The bug has existed since the mech first started writing to predict-api (commit `06cab97d`, 2026-06-23) — this is not a regression, just an on-chain path that was never wired.

## Fix

In `_handle_get_task`, after the validated IPFS content passes the prompt-size check, write it to `in_memory_requests[str(req_id)]` in the same shape the off-chain handler already uses (JSON-encoded task metadata keyed by request_id string):

```python
self.context.shared_state.setdefault(IN_MEMORY_REQUESTS, {})[
    str(req_id)
] = json.dumps(task_data)
```

## Test plan

- [x] New unit test `test_handle_get_task_populates_in_memory_requests` — verifies content lands in the cache with the same key + shape as the off-chain writer.
- [x] Sibling `handle_get_task` tests (7 total including the new one) all pass.
- [x] `autonomy packages lock --check` clean.
- [x] `autonomy push-all` succeeded (see PR CI for the hashes).

## Rollout notes

- Fixes go-forward. Every new `mech_onchain` row after this ships will land with real content.
- Existing predict-api rows with the placeholder are not overwritten by this change. Backfilling those requires the predict-api on-chain backfill (which fetches IPFS and populates content) to run over the affected window, followed by a mech-analytics one-shot to re-extract `question_title` and `market_id` on already-scored rows. Both are separate work.

### DIFF
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1441,6 +1441,10 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             self._invalid_request = True
             return
 
+        self.context.shared_state.setdefault(IN_MEMORY_REQUESTS, {})[
+            str(req_id)
+        ] = json.dumps(task_data)
+
         my_mech = self._get_designated_marketplace_mech_address().lower()
         exec_prio = str(executing_task.get("priorityMech", "")).lower()
         # Off-chain tasks never carry ``priorityMech`` (the field is set
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1441,9 +1441,9 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             self._invalid_request = True
             return
 
-        self.context.shared_state.setdefault(IN_MEMORY_REQUESTS, {})[
-            str(req_id)
-        ] = json.dumps(task_data)
+        self.context.shared_state.setdefault(IN_MEMORY_REQUESTS, {})[str(req_id)] = (
+            json.dumps(task_data)
+        )
 
         my_mech = self._get_designated_marketplace_mech_address().lower()
         exec_prio = str(executing_task.get("priorityMech", "")).lower()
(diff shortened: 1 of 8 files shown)


packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1441,6 +1441,10 @@ def _handle_get_task(self, message: IpfsMessage, dialogue: Dialogue) -> None:
             self._invalid_request = True
             return
 
+        self.context.shared_state.setdefault(IN_MEMORY_REQUESTS, {})[str(req_id)] = (
+            json.dumps(task_data)
+        )
```
OjusWiZard, line 1446: it's cleaned up later, right?

OjusWiZard (approved):

Adamantios (approved):

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Yes on the happy path — _finalize_done_task at L1777 pops via pop(str(req_id)) after the analytics payload is built. Cleanup key matches the write key exactly.

PR #480, opened 2026-08-24
fix(task_execution): populate offchain_request_responses for on-chain deliveries

DIvyaNautiyal07:
## Summary

- The on-chain branch of `_handle_done_task` did not write the response envelope into `shared_state[OFFCHAIN_REQUEST_RESPONSES]`; the off-chain branch already did.
- `_build_predict_api_event` reads `response["result"]` off this shared-state slot for both paths, so on-chain deliveries landed in the predict-api data lake with `status='failed'` and `result=None` regardless of whether the tool succeeded or failed.
- Add a symmetric write on the on-chain branch. Envelope shape `{request_id, response}` matches what the event builder already consumes. Tool result flows through on success, tool error string flows through on failure.

## Test plan

- [x] `test_handle_done_task_onchain_stores_response_envelope_in_shared_state` — on-chain success stores real tool result under `response["result"]`
- [x] `test_handle_done_task_onchain_invalid_request_stores_reason_in_shared_state` — on-chain invalid-request stores failure reason under `response["result"]`
- [x] All existing `_handle_done_task` tests continue to pass (17/17)
- [x] Full `packages/valory/skills/task_execution/tests/` suite passes (511/511)
- [x] `autonomy packages lock --check` — verified
- [x] `autonomy push-all` — new hashes published to IPFS
- [ ] Downstream: bump `TOOLS_TO_PACKAGE_HASH` / package refs in `mech-predict` once merged

### DIFF
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -19,6 +19,7 @@
 
 """This package contains the implementation of ."""
 
+import functools
 import json
 import threading
 import time
@@ -1284,12 +1285,11 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             self._finalize_done_task(local_cid)
             return
 
-        self._done_task["_pending_response"] = response
-
         msg, dialogue = self._build_ipfs_store_file_req(
             {str(req_id): json.dumps(response)}
         )
-        self.send_message(msg, dialogue, self._handle_store_response)
+        callback = functools.partial(self._handle_store_response, response=response)
+        self.send_message(msg, dialogue, callback)
 
     def _restart_executor(self) -> None:
         """Restarts the executor."""
@@ -1673,7 +1673,12 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         )
         return message, dialogue
 
-    def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> None:
+    def _handle_store_response(
+        self,
+        message: IpfsMessage,
+        dialogue: Dialogue,
+        response: Optional[Dict[str, Any]] = None,
+    ) -> None:
         """Handle the response from ipfs for a store response request."""
         if not self._executing_task:
             self.context.logger.error(
@@ -1686,18 +1691,20 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         self.context.logger.info(
             f"Response for request {req_id} stored on IPFS with hash {ipfs_hash}."
         )
-        done_task = self._done_task if isinstance(self._done_task, dict) else {}
-        pending_response = done_task.get("_pending_response")
-        if pending_response is not None:
-            request_id_str = str(req_id)
-            self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
-                request_id_str
-            ] = {
-                "request_id": request_id_str,
-                "status": "ok",
-                "content_cid": ipfs_hash,
-                "response": pending_response,
-            }
+        if response is not None and self.params.use_offchain:
+            mech_address = self._get_designated_marketplace_mech_address()
+            mech_config = self.params.mech_to_config.get(mech_address.lower())
+            if mech_config is not None and mech_config.is_marketplace_mech:
+                request_id_str = str(req_id)
+                envelope_status = "rejected" if self._invalid_request else "ok"
+                self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
+                    request_id_str
+                ] = {
+                    "request_id": request_id_str,
+                    "status": envelope_status,
+                    "content_cid": ipfs_hash,
+                    "response": response,
+                }
         self._finalize_done_task(ipfs_hash)
 
     def _finalize_done_task(self, cid: str) -> None:
@@ -1799,8 +1806,6 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 self.context.shared_state.get(OFFCHAIN_REQUEST_RESPONSES, {}).pop(
                     str(req_id), None
                 )
-        # Ephemeral field for the IPFS callback; keep it out of consensus.
-        done_task.pop("_pending_response", None)
         # add to done tasks, in thread safe way
         with self.done_tasks_lock:
             self.done_tasks.append(done_task)
@@ -1898,11 +1903,12 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # forcing every completed task to be tagged ``status="failed"`` in
         # the analytics row.
         # ``OFFCHAIN_REQUEST_RESPONSES`` is written under ``str(req_id)`` by
-        # ``_handle_done_task`` and ``_record_offchain_failure``; ``req_id``
-        # is ``int`` for off-chain tasks post-ingress coercion. Read with the
-        # ``str`` key too or the analytics row is tagged ``status="failed"``
-        # for every successful off-chain delivery, silently, because the
-        # poller sees the right answer but the lake row does not.
+        # ``_handle_done_task`` (off-chain), ``_handle_store_response`` (on-chain,
+        # after the IPFS CID lands) and ``_record_offchain_failure``. ``req_id``
+        # is ``int`` for off-chain tasks post-ingress coercion, so the lookup
+        # goes through ``str`` too — otherwise every successful off-chain
+        # delivery tags the analytics row ``status="failed"`` silently, since
+        # the poller sees the right answer but the lake row does not.
         response_envelope_raw = self.context.shared_state.get(
             OFFCHAIN_REQUEST_RESPONSES, {}
         ).get(request_id_str, {})
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -88,8 +88,9 @@ LAST_READ_ATTEMPT_TS = "last_read_attempt_ts"
 INFLIGHT_READ_TS = "inflight_read_ts"
 REQUEST_ID_TO_DELIVERY_RATE_INFO = "request_id_to_delivery_rate_info"
 # Shared-state keys owned by the MechHttpHandler (handlers.py); mirrored here
-# because the finalize path writes response envelopes for both off-chain
-# and on-chain deliveries from the behaviour side.
+# because the off-chain finalize path writes the response envelope for the
+# ``/fetch_offchain_info`` polling endpoint and clears the buffered request
+# metadata from the behaviour side.
 OFFCHAIN_REQUEST_RESPONSES = "offchain_request_responses"
 IN_MEMORY_REQUESTS = "in_memory_requests"
 # Shared-state keys owned by the MechHttpHandler; mirrored here because
@@ -1282,13 +1283,15 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 local_cid,
                 preimage_buffer.STATUS_DELIVERED,
             )
-            self._finalize_done_task(local_cid)
+            self._finalize_done_task(local_cid, response=response)
             return
 
         msg, dialogue = self._build_ipfs_store_file_req(
             {str(req_id): json.dumps(response)}
         )
-        callback = functools.partial(self._handle_store_response, response=response)
+        callback = functools.partial(
+            self._handle_store_response, response=response, req_id=req_id
+        )
         self.send_message(msg, dialogue, callback)
 
     def _restart_executor(self) -> None:
@@ -1678,6 +1681,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         message: IpfsMessage,
         dialogue: Dialogue,
         response: Optional[Dict[str, Any]] = None,
+        req_id: Any = None,
     ) -> None:
         """Handle the response from ipfs for a store response request."""
         if not self._executing_task:
@@ -1686,28 +1690,28 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             )
             return
 
-        ipfs_hash = to_v1(message.ipfs_hash)
-        req_id = self._executing_task["requestId"]
-        self.context.logger.info(
-            f"Response for request {req_id} stored on IPFS with hash {ipfs_hash}."
-        )
-        if response is not None and self.params.use_offchain:
-            mech_address = self._get_designated_marketplace_mech_address()
-            mech_config = self.params.mech_to_config.get(mech_address.lower())
-            if mech_config is not None and mech_config.is_marketplace_mech:
-                request_id_str = str(req_id)
-                envelope_status = "rejected" if self._invalid_request else "ok"
-                self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
-                    request_id_str
-                ] = {
-                    "request_id": request_id_str,
-                    "status": envelope_status,
-                    "content_cid": ipfs_hash,
-                    "response": response,
-                }
-        self._finalize_done_task(ipfs_hash)
+        live_req_id = self._executing_task["requestId"]
+        if req_id is not None and live_req_id != req_id:
+            # Stale callback from a previously-executed task that has since been
+            # timed out or swapped out. Dropping it prevents cross-task
+            # misattribution of the response and the CID.
+            self.context.logger.warning(
+                "Discarding stale IPFS store callback for request %s; "
+                "current executing task is %s.",
+                req_id,
+                live_req_id,
+            )
+            return
 
-    def _finalize_done_task(self, cid: str) -> None:
+        ipfs_hash = to_v1(message.ipfs_hash)
+        self.context.logger.info(
+            f"Response for request {live_req_id} stored on IPFS with hash {ipfs_hash}."
+        )
+        self._finalize_done_task(ipfs_hash, response=response)
+
+    def _finalize_done_task(
+        self, cid: str, response: Optional[Dict[str, Any]] = None
+    ) -> None:
         """Apply the CID to the in-flight done_task and publish it.
 
         Shared by the on-chain path (CID from a real IPFS store response) and the
@@ -1719,6 +1723,9 @@ class TaskExecutionBehaviour(SimpleBehaviour):
 
         :param cid: the multibase-encoded CIDv1 to record on chain via
             ``task_result``.
+        :param response: the tool response envelope the predict-api event builder
+            needs. Both call sites pass this in so the value flows down the
+            synchronous stack rather than round-tripping through ``shared_state``.
         """
         executing_task = self._executing_task
         if executing_task is None:
@@ -1798,14 +1805,8 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 done_task=done_task,
                 cid=cid,
                 executing_task=cast(Dict[str, Any], executing_task),
+                response=response,
             )
-            if not is_offchain:
-                # Off-chain envelopes are retained for the polling endpoint
-                # (pruned on same-id re-request in handlers.py); on-chain
-                # envelopes have no equivalent lifecycle and must be pruned here.
-                self.context.shared_state.get(OFFCHAIN_REQUEST_RESPONSES, {}).pop(
-                    str(req_id), None
-                )
         # add to done tasks, in thread safe way
         with self.done_tasks_lock:
             self.done_tasks.append(done_task)
@@ -1831,20 +1832,21 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         done_task: Dict[str, Any],
         cid: str,
         executing_task: Dict[str, Any],
+        response: Optional[Dict[str, Any]] = None,
     ) -> Dict[str, Any]:
         """Build the structured predict-api event payload from on-hand data.
 
         The post-settlement behaviour batches every event from one FSM round
         into a single signed POST to the predict-api data lake. Fields are
         populated best-effort from the buffered request metadata
-        (``IN_MEMORY_REQUESTS``), the offchain response
-        (``OFFCHAIN_REQUEST_RESPONSES``), the in-flight ``executing_task``,
-        and skill params. Optional fields that aren't readily available are
-        left ``None``; the predict-api server accepts a NULL there. Required
-        fields default to safe placeholders rather than blocking the row: a
-        malformed event surfaces as a 422 when the predict-api write fires (and
-        lands in the local replay buffer so it's recoverable), which is
-        preferable to silently dropping analytics data.
+        (``IN_MEMORY_REQUESTS``), the tool ``response`` dict threaded down
+        from the caller, the in-flight ``executing_task``, and skill params.
+        Optional fields that aren't readily available are left ``None``; the
+        predict-api server accepts a NULL there. Required fields default to
+        safe placeholders rather than blocking the row: a malformed event
+        surfaces as a 422 when the predict-api write fires (and lands in the
+        local replay buffer so it's recoverable), which is preferable to
+        silently dropping analytics data.
 
         Datetimes are emitted as ISO 8601 strings with UTC offset so the
         predict-api ``AwareDatetime`` parser accepts them; mech ints (Unix
@@ -1855,6 +1857,12 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         :param done_task: the done-task dict appended to ``done_tasks``.
         :param cid: the locally-computed multibase CIDv1 for the response.
         :param executing_task: the original request dict from the handler.
+        :param response: the tool response envelope threaded from the caller;
+            carries ``result``, ``executed_at``, ``metadata`` and ``cost_dict``
+            for the current task. Kept as a parameter so the value flows down
+            the synchronous stack instead of round-tripping through
+            ``shared_state``, which would otherwise open a stale-callback
+            misattribution window.
         :return: a ``MechSettlementEvent``-shaped dict ready for FSM consensus
             replication and downstream signing in the post-tx behaviour.
         """
@@ -1895,28 +1903,11 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             request_data = request_data_raw
         else:
             request_data = {}
-        # OFFCHAIN_REQUEST_RESPONSES stores an envelope
-        # ``{"request_id":..., "status":"ok", "content_cid":..., "response":<inner>}``;
-        # the tool's actual ``result`` / ``executed_at`` live on the inner
-        # ``response`` dict, not on the envelope. Reading the envelope
-        # directly (the original code) returned ``None`` for both fields,
-        # forcing every completed task to be tagged ``status="failed"`` in
-        # the analytics row.
-        # ``OFFCHAIN_REQUEST_RESPONSES`` is written under ``str(req_id)`` by
-        # ``_handle_done_task`` (off-chain), ``_handle_store_response`` (on-chain,
-        # after the IPFS CID lands) and ``_record_offchain_failure``. ``req_id``
-        # is ``int`` for off-chain tasks post-ingress coercion, so the lookup
-        # goes through ``str`` too — otherwise every successful off-chain
-        # delivery tags the analytics row ``status="failed"`` silently, since
-        # the poller sees the right answer but the lake row does not.
-        response_envelope_raw = self.context.shared_state.get(
-            OFFCHAIN_REQUEST_RESPONSES, {}
-        ).get(request_id_str, {})
-        response_envelope = (
-            response_envelope_raw if isinstance(response_envelope_raw, dict) else {}
-        )
-        inner_response = response_envelope.get("response", {})
-        response_data = inner_response if isinstance(inner_response, dict) else {}
+        # ``response`` is threaded from the caller (both off-chain and on-chain
+        # finalize paths pass it in). Keeping the read local avoids the
+        # stale-callback / cross-task misattribution window that a round-trip
+        # through ``shared_state`` would open.
+        response_data = response if isinstance(response, dict) else {}
         tool = str(done_task.get("tool") or "unknown")
         delivery_mech = str(
             done_task.get("mech_address") or self.params.mech_marketplace_address or ""
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1284,6 +1284,14 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             self._finalize_done_task(local_cid)
             return
 
+        request_id_str = str(req_id)
+        self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
+            request_id_str
+        ] = {
+            "request_id": request_id_str,
+            "response": response,
+        }
+
         msg, dialogue = self._build_ipfs_store_file_req(
             {str(req_id): json.dumps(response)}
         )
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -87,8 +87,8 @@ LAST_READ_ATTEMPT_TS = "last_read_attempt_ts"
 INFLIGHT_READ_TS = "inflight_read_ts"
 REQUEST_ID_TO_DELIVERY_RATE_INFO = "request_id_to_delivery_rate_info"
 # Shared-state keys owned by the MechHttpHandler (handlers.py); mirrored here
-# because the off-chain finalize path writes the rejection signal and clears the
-# buffered request metadata from the behaviour side.
+# because the finalize path writes response envelopes for both off-chain
+# and on-chain deliveries from the behaviour side.
 OFFCHAIN_REQUEST_RESPONSES = "offchain_request_responses"
 IN_MEMORY_REQUESTS = "in_memory_requests"
 # Shared-state keys owned by the MechHttpHandler; mirrored here because
@@ -1284,13 +1284,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             self._finalize_done_task(local_cid)
             return
 
-        request_id_str = str(req_id)
-        self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
-            request_id_str
-        ] = {
-            "request_id": request_id_str,
-            "response": response,
-        }
+        self._done_task["_pending_response"] = response
 
         msg, dialogue = self._build_ipfs_store_file_req(
             {str(req_id): json.dumps(response)}
@@ -1692,6 +1686,18 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         self.context.logger.info(
             f"Response for request {req_id} stored on IPFS with hash {ipfs_hash}."
         )
+        done_task = self._done_task if isinstance(self._done_task, dict) else {}
+        pending_response = done_task.get("_pending_response")
+        if pending_response is not None:
+            request_id_str = str(req_id)
+            self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
+                request_id_str
+            ] = {
+                "request_id": request_id_str,
+                "status": "ok",
+                "content_cid": ipfs_hash,
+                "response": pending_response,
+            }
         self._finalize_done_task(ipfs_hash)
 
     def _finalize_done_task(self, cid: str) -> None:
@@ -1786,6 +1792,15 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 cid=cid,
                 executing_task=cast(Dict[str, Any], executing_task),
             )
+            if not is_offchain:
+                # Off-chain envelopes are retained for the polling endpoint
+                # (pruned on same-id re-request in handlers.py); on-chain
+                # envelopes have no equivalent lifecycle and must be pruned here.
+                self.context.shared_state.get(OFFCHAIN_REQUEST_RESPONSES, {}).pop(
+                    str(req_id), None
+                )
+        # Ephemeral field for the IPFS callback; keep it out of consensus.
+        done_task.pop("_pending_response", None)
         # add to done tasks, in thread safe way
         with self.done_tasks_lock:
             self.done_tasks.append(done_task)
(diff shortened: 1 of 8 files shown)


packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1684,6 +1691,20 @@ def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> No
         self.context.logger.info(
             f"Response for request {req_id} stored on IPFS with hash {ipfs_hash}."
         )
+        if response is not None and self.params.use_offchain:
```
jmoreira-valory: 🟡 **F1 · MEDIUM — On-chain envelope can outlive its task and permanently poison the off-chain response ledger**

`response` is bound to the *sending* task by the partial, but `req_id` is read live off `self._executing_task`, and the prune that is meant to undo this write sits in a branch `_finalize_done_task` can return before reaching.

Concretely: `IpfsHandler.handle` invokes a store callback **even when the request-handling deadline has already expired** — `handlers.py:581-587` logs "Deadline reached ... Invoking callback for cleanup" and calls it anyway. If `_handle_timeout_task` has since reset the slot and a *different* task is executing, this block passes the `if not self._executing_task` guard at line 1683, writes `{status: "ok", content_cid: <task A's CID>, response: <task A's answer>}` under **task B's** id, and then `_finalize_done_task` finds `self._done_task is None` (cleared by `_reset_executing_task`, `behaviours.py:2175`), takes the early return at line 1747, and never reaches the `pop` at line 1806.

That entry is then permanent, with no self-healing path. `OFFCHAIN_REQUEST_RESPONSES` is not a private scratch slot: it is the serving store for `GET /fetch_offchain_info` (`handlers.py:2344`) **and** the off-chain duplicate-admission ledger — `_is_request_already_accepted` treats any entry whose `status != "rejected"` as "already accepted" (`handlers.py:2310-2315`). So the poisoned id both returns another request's answer to whoever polls it and is refused forever as a duplicate. Off-chain request ids are requester-chosen uint256s, so the two id spaces genuinely overlap.

Mitigating, and worth stating plainly: `use_offchain` is `false` at `skill.yaml:122` and is not overridden in `aea-config.yaml` or `service.yaml`, so this path is dark today. But flipping that flag is the whole point of the feature.

The clean fix is to not round-trip through shared state at all — the value only has to survive one synchronous call. Thread it as a parameter (`_finalize_done_task(cid, response=response)` -> `_build_predict_api_event(..., response_data=response)`) and drop both the write and the prune. If you would rather keep the shared-state route, then at minimum make the prune unconditional (wrap the `_finalize_done_task(ipfs_hash)` call in `try/finally`, or pop on every `return` inside `_finalize_done_task`) and reject the callback when the live task identity does not match the one the partial was built for.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1684,6 +1691,20 @@ def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> No
...
+            mech_config = self.params.mech_to_config.get(mech_address.lower())
+            if mech_config is not None and mech_config.is_marketplace_mech:
+                request_id_str = str(req_id)
+                envelope_status = "rejected" if self._invalid_request else "ok"
+                self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
+                    request_id_str
+                ] = {
+                    "request_id": request_id_str,
+                    "status": envelope_status,
+                    "content_cid": ipfs_hash,
```
jmoreira-valory: 🔵 **F2 · LOW — Envelope reuses content_cid and status with off-chain semantics that don't hold on-chain**

Only `envelope["response"]` is ever read on this path — `_build_predict_api_event` pulls `response_envelope.get("response", {})` and ignores both `status` and `content_cid`. Meanwhile each of those fields carries off-chain-specific meaning everywhere else in the skill:

- `content_cid` on an off-chain envelope is `compute_cidv1(json.dumps(response))`, a **bare single-block CIDv1** that a client is documented to re-derive from `stored["response"]` in order to verify the delivery against the on-chain commitment (`behaviours.py:1226-1236`). `ipfs_hash` here is the **directory-wrapped** store CID — the off-chain branch's own comment says it "intentionally differs from the on-chain path's directory-wrapped CID". A client applying the documented verification to this entry gets a mismatch.
- `status: "ok"` is precisely what `_is_request_already_accepted` reads to decide an id is already spoken for (`handlers.py:2314`).

So the PR writes two fields that are dead weight on the happy path and actively wrong on any path where the entry survives (see the prune finding). Write only what the consumer reads, or better, drop the envelope entirely in favour of passing `response` down as an argument.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1925,13 +1937,123 @@ def test_handle_done_task_onchain_still_uploads_to_ipfs(
...
+            "tool boom",
+            "rejected",
+            None,
+            "failed",
+            "tool boom",
+        ),
+    ],
+    ids=["success", "invalid_request_failure"],
+)
+def test_handle_done_task_onchain_end_to_end_predict_api_event(  # noqa: PLR0913
```
jmoreira-valory, line 2001: 🔵 **F3 · LOW — No test pins the use_offchain=False or prune-skipped behaviour of the new write**

`test_handle_done_task_onchain_end_to_end_predict_api_event` covers the success and invalid-request shapes, but it always runs against the fixture default `use_offchain=True` (`conftest.py:192`), so nothing pins the two behaviours that matter most for a dark-shipped flag:

1. with `params_stub.use_offchain = False`, `_handle_store_response` must leave `OFFCHAIN_REQUEST_RESPONSES` completely untouched;
2. when `_finalize_done_task` takes its invalid-`done_task` early return (set `behaviour._done_task = None` before invoking the captured callback), no entry may be left behind.

The existing `use_offchain` gate tests at lines 4075-4165 only exercise `_finalize_done_task`, not the new write, so neither is covered today. Both are a handful of lines on top of the parametrisation you already have, and (2) is the direct regression test for the prune finding above.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1684,6 +1691,20 @@ def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> No
         self.context.logger.info(
             f"Response for request {req_id} stored on IPFS with hash {ipfs_hash}."
         )
+        if response is not None and self.params.use_offchain:
+            mech_address = self._get_designated_marketplace_mech_address()
+            mech_config = self.params.mech_to_config.get(mech_address.lower())
+            if mech_config is not None and mech_config.is_marketplace_mech:
```
jmoreira-valory: ⚪ **F4 · NIT — The is_marketplace_mech guard here is dead code**

`_get_designated_marketplace_mech_address` (`behaviours.py:1078`) only ever returns a key whose config has `is_marketplace_mech=True` — otherwise it raises `ValueError` — and `Params.__init__` lowercases every `mech_to_config` key at construction (`models.py:98-101`). So `.get(mech_address.lower())` cannot miss and the flag cannot be `False`. Two statements later `_finalize_done_task` performs the identical lookup with `[...]`, so a miss would surface as a `KeyError` there regardless.

As written the condition reads like a real per-mech gate and will mislead the next reader into thinking non-marketplace deliveries are excluded at this point. Collapse it into the `use_offchain` check, or mirror `_finalize_done_task` and index directly so the two sites fail the same way.

jmoreira-valory (changes requested):

**What this does.** The mech's on-chain delivery path never stashed the tool's response envelope into the `offchain_request_responses` shared-state slot, but that slot is exactly where the predict-api analytics event builder looks for the result. Every on-chain delivery therefore landed in the data lake tagged `status='failed'` with a null result, whether the tool succeeded or not. This PR threads the response dict through the IPFS-store callback with `functools.partial`, writes the envelope once the real CID is known, and pops it again right after the analytics event is built.

CI is green (22/22 checks) at `dcf5d9d`, and the package CID cascade is complete: `task_execution` -> `mech_abci`, `task_submission_abci`, `agent/mech` and `service/mech` all point at the new hash at HEAD, with no stale references left anywhere under `packages/`.

The diagnosis in the PR body checks out against source. `_build_predict_api_event` reads `OFFCHAIN_REQUEST_RESPONSES[str(req_id)]["response"]`, the on-chain branch never wrote it, and the `status` / `result` derivation at `behaviours.py:2052-2065` then tags every on-chain row `failed`. Deferring the write to `_handle_store_response` rather than doing it in `_handle_done_task` is the right call, since the real IPFS CID is not known until the store lands. The test-side fix (`_dispatch_target` unwrapping the `functools.partial`) is clean and keeps the existing dispatch stubs honest rather than loosening them.

Scope note, not a finding: `response["executed_at"]` is written as an ISO 8601 **string** by `_handle_done_task`, but `_build_predict_api_event` only converts it when it is an `int` / `float` and otherwise falls back to `now_iso` (`behaviours.py:1946-1962`). So the lake's `executed_at` equals `delivered_at` on both paths. That is pre-existing and equally true off-chain, so it is out of scope here, but it does mean this PR restores `result` without restoring the execution timestamp.

**Findings** — 4 · 🟡 1 Medium · 🔵 2 Low · ⚪ 1 Nit

| F1 | 🟡 MEDIUM | [packages/valory/skills/task_execution/behaviours.py:1694](https://github.com/valory-xyz/mech/blob/dcf5d9d592c58bd41091b079b24939f4f65f3a47/packages/valory/skills/task_execution/behaviours.py#L1694) | **On-chain envelope can outlive its task and permanently poison the off-chain response ledger** |
| F2 | 🔵 LOW | [packages/valory/skills/task_execution/behaviours.py:1705](https://github.com/valory-xyz/mech/blob/dcf5d9d592c58bd41091b079b24939f4f65f3a47/packages/valory/skills/task_execution/behaviours.py#L1705) | **Envelope reuses content_cid and status with off-chain semantics that don't hold on-chain** |
| F3 | 🔵 LOW | [packages/valory/skills/task_execution/tests/test_behaviours.py:1984](https://github.com/valory-xyz/mech/blob/dcf5d9d592c58bd41091b079b24939f4f65f3a47/packages/valory/skills/task_execution/tests/test_behaviours.py#L1984) | **No test pins the use_offchain=False or prune-skipped behaviour of the new write** |
| F4 | ⚪ NIT | [packages/valory/skills/task_execution/behaviours.py:1697](https://github.com/valory-xyz/mech/blob/dcf5d9d592c58bd41091b079b24939f4f65f3a47/packages/valory/skills/task_execution/behaviours.py#L1697) | **The is_marketplace_mech guard here is dead code** |

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1287,7 +1288,8 @@ def _handle_done_task(self, task_result: Any) -> None:
         msg, dialogue = self._build_ipfs_store_file_req(
             {str(req_id): json.dumps(response)}
         )
-        self.send_message(msg, dialogue, self._handle_store_response)
+        callback = functools.partial(self._handle_store_response, response=response)
+        self.send_message(msg, dialogue, callback)
```
OjusWiZard: **Blocking — a late store callback writes this task's response under a *different* request's id.**

`response` is captured in the closure at send time, but `_handle_store_response` re-reads `req_id` and `_invalid_request` from the live `self._executing_task` when the callback actually fires. Those two halves can belong to different requests.

The window is opened by the shipped retry config (`skill.yaml:118-119`, `task_deadline: 240.0` / `timeout_limit: 3`):

1. Task A's STORE is in flight; the deadline (carried over from the execution phase, never reset for the STORE wait) elapses.
2. `_execute_task` (behaviours.py:874-879) fires `_handle_timeout_task()`. On a non-terminal timeout it sets `in_flight_req = False`, re-queues A, and nulls `_executing_task` (behaviours.py:1315, 1390-1391) — but never removes `req_to_callback[nonce_A]`, so A's STORE callback stays armed.
3. `act()` picks up task B and assigns `_executing_task = task_B`.
4. A's STORE reply lands. `IpfsHandler.handle` invokes the callback **unconditionally, even past its own deadline** (handlers.py:583-591, "Invoking callback for cleanup"). The `if not self._executing_task` guard at line 1683 passes, because the slot isn't empty — it's been *reassigned*.

Reproduced against this head:

```
ENVELOPE KEYS: ['200']
ENVELOPE: {'200': {'request_id': '200', 'status': 'ok', 'content_cid': 'bafyA',
                   'response': {'requestId': 100, 'result': 'prediction: yes', ...}}}
```

Task A (id 100, CID `bafyA`) lands entirely under task B's id.

Impact, since `handlers.py:2344-2351` serves `offchain_request_responses[request_id]` verbatim over `GET /fetch_offchain_info`:
- a requester polling id B is served another requester's answer with `status: "ok"`;
- `_is_already_accepted` (handlers.py:2312-2315) treats any non-`rejected` entry as accepted, so a genuine request for that id is short-circuited and never enqueued;
- B is finalized with `to_multihash(A's cid)`, and when B's *own* STORE reply arrives it hits the empty-slot guard at line 1683 and is silently dropped — B's real result is never delivered at all.

The CID misattribution pre-dates this PR; what's new is propagating the wrong **response body** into a caller-visible channel.

Cheapest fix uses the machinery already introduced here — capture `req_id` alongside `response` and no-op on mismatch:

```python
callback = functools.partial(
    self._handle_store_response, response=response, req_id=req_id
)
```

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1684,6 +1691,20 @@ def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> No
...
+                request_id_str = str(req_id)
+                envelope_status = "rejected" if self._invalid_request else "ok"
+                self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
+                    request_id_str
+                ] = {
+                    "request_id": request_id_str,
+                    "status": envelope_status,
+                    "content_cid": ipfs_hash,
+                    "response": response,
+                }
```
OjusWiZard: **Consider threading `response` as a parameter instead of round-tripping it through `shared_state`.**

`_handle_store_response` -> `_finalize_done_task` -> `_build_predict_api_event` -> pop is one synchronous call stack with no yield point. The `response` dict is already in hand right here; writing it into global mutable state only to read it back three frames later and delete it buys nothing — and it's what makes the two blocking issues on this PR possible.

`_finalize_done_task(cid, response=...)` -> `_build_predict_api_event(..., response=...)` would:
- remove the misattribution vector (nothing keyed by a possibly-stale `req_id`);
- remove the leak on the invalid-`done_task` early return, by construction;
- remove the duplicated `mech_config` lookup below;
- let the off-chain call site at behaviours.py:1285 pass the same argument, so both paths become symmetric.

The off-chain `shared_state` write at behaviours.py:1249-1256 must stay — that one is genuinely consumed by `/fetch_offchain_info`.

packages/valory/skills/task_execution/behaviours.py
OjusWiZard: This gate can't be false. `models.py:98-101` lowercases every key at construction, and `_get_designated_marketplace_mech_address` (behaviours.py:1078-1084) only ever returns a key whose config already has `is_marketplace_mech=True` — it either returns a marketplace mech or raises `ValueError`. So `.get()` cannot miss and the `is_marketplace_mech` check cannot fail.

It also duplicates the identical lookup `_finalize_done_task` performs at behaviours.py:1760-1762 a few frames later (there via direct indexing rather than `.get`).

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1684,6 +1691,20 @@ def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> No
         self.context.logger.info(
             f"Response for request {req_id} stored on IPFS with hash {ipfs_hash}."
         )
+        if response is not None and self.params.use_offchain:
+            mech_address = self._get_designated_marketplace_mech_address()
+            mech_config = self.params.mech_to_config.get(mech_address.lower())
+            if mech_config is not None and mech_config.is_marketplace_mech:
+                request_id_str = str(req_id)
+                envelope_status = "rejected" if self._invalid_request else "ok"
+                self.context.shared_state.setdefault(OFFCHAIN_REQUEST_RESPONSES, {})[
```
OjusWiZard: `envelope_status` is write-only on this path. `_build_predict_api_event` computes its own `status` from `self._invalid_request` and `result_text` and never reads the envelope's; the only consumers of the envelope `status` (`_is_already_accepted`, `/fetch_offchain_info`) are off-chain-specific and can't observe an entry popped in the same call stack. Same goes for `request_id` and `content_cid` here.

Reads like a status signal, but nothing consumes it — worth dropping so a future change doesn't wire a consumer to it.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1778,6 +1799,13 @@ def _finalize_done_task(self, cid: str) -> None:
                 cid=cid,
                 executing_task=cast(Dict[str, Any], executing_task),
             )
+            if not is_offchain:
+                # Off-chain envelopes are retained for the polling endpoint
+                # (pruned on same-id re-request in handlers.py); on-chain
+                # envelopes have no equivalent lifecycle and must be pruned here.
+                self.context.shared_state.get(OFFCHAIN_REQUEST_RESPONSES, {}).pop(
+                    str(req_id), None
+                )
```
OjusWiZard: **Blocking — this pop is unreachable on one path, and the envelope leaks.**

`_finalize_done_task` returns at behaviours.py:1735-1746 (`done_task` is `None` / not a `Dict`) via `_reset_executing_task()`, *before* ever reaching this block. Since the envelope is written in `_handle_store_response` **before** `_finalize_done_task` is called, that entry is never cleaned up.

Not hypothetical — it's the same repro as the callback race above. When a stale callback lands on a freshly-picked task, `_done_task` is `None` (just cleared by `_reset_executing_task`), the early return fires, and the bogus envelope stays in `shared_state` permanently: served by `/fetch_offchain_info` and blocking `_is_already_accepted` for that id.

When the next task happens to be off-chain, the early return instead routes through `_record_offchain_failure`, terminally rejecting a request that is still executing:

```
ENVELOPE: {'200': {'request_id': '200', 'status': 'rejected', 'reason': 'invalid done task'}}
```

Threading `response` as a parameter (see the comment above) removes this by construction. Otherwise the pop needs to move ahead of the validity check, or into a `finally`.

packages/valory/skills/task_execution/tests/test_behaviours.py
OjusWiZard, line 2001: Coverage gap: this only ever exercises `use_offchain=True` / `is_marketplace_mech=True`. `_seed_executing_task` hardcodes the latter and `conftest.py:192` defaults the former, so both parametrizations take the same path through the new gate. Untested with a non-null `response`: `use_offchain=False`, `is_marketplace_mech=False` (the legacy on-chain mechs the commit message says should be excluded), and `mech_config is None`.

More usefully: the leak flagged on `behaviours.py:1802` is invisible to the current suite. `test_handle_store_response_invalid_done_task_resets_state` (line 2072) exercises exactly that early-return branch, but passes the default `response=None`, so there's no envelope to leak. Passing a `response` there and asserting the key is gone afterwards would have caught it.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1925,13 +1937,123 @@ def test_handle_done_task_onchain_still_uploads_to_ipfs(
...
+    assert captured_envelope["response"]["result"] == expected_envelope_result
+
+    done_tasks = shared_state[beh_mod.DONE_TASKS]
+    assert len(done_tasks) == 1
+    done_task = done_tasks[0]
+    event = done_task["predict_api_event"]
+    assert event["response"]["result"] == expected_event_result
+    assert event["response"]["status"] == expected_event_status
+    assert event["response"]["error"] == expected_event_error
+    assert "_pending_response" not in done_task
```
OjusWiZard: Vestigial — `git grep _pending_response` across this branch returns only this line. The second commit replaced the `_pending_response` side-channel with `functools.partial`, so this asserts the absence of a key no code writes any more.

OjusWiZard (commented):
Reviewed the full diff and traced the call chains through `handlers.py` / `models.py`.

**Verified green:** 511/511 `task_execution` tests pass locally at `dcf5d9d`; all CI checks pass including `lock_check` on three OSes. The hash relock is internally consistent — every cross-reference in `aea-config.yaml`, `service.yaml` and the three `skill.yaml`s matches `packages.json`.

**The diagnosis is right.** `_build_predict_api_event` really was reading an empty envelope on the on-chain path, and the rows really were landing as `status='failed'` regardless of outcome. My concern is with the mechanism, not the goal.

Two blocking issues, both stemming from routing the response through `shared_state` rather than passing it down the (fully synchronous) call stack:

1. A late IPFS store callback pairs the closure-captured `response` with a live-read `req_id`, so one request's answer can be written under another's id — and served over `/fetch_offchain_info`. Reachable under the shipped `timeout_limit: 3` / `task_deadline: 240.0` defaults; I reproduced it against this head.
2. The compensating pop sits after an early return, so the envelope leaks on the invalid-`done_task` path.

Threading `response` as a parameter into `_finalize_done_task` / `_build_predict_api_event` closes both by construction, plus the two smaller points below, and shrinks the diff. Details inline.

Two things I checked that turned out **fine**, noting them so they don't get re-litigated:
- The `functools.partial` binding is correct — `IpfsHandler.handle` invokes callbacks positionally as `callback(ipfs_msg, dialogue)` (handlers.py:591), so `response` arrives by keyword against the new signature.
- On-chain / off-chain request-id collision in the shared dict is already prevented upstream: off-chain ids come from `compute_request_id` (the marketplace's own `getRequestId` formula) and the wire nonce is bound to on-chain `mapNonces[sender]` via `_bind_wire_nonce_to_chain`. The misattribution in (1) comes purely from the stale callback, not from id-space overlap.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in a7384d6f. Went with the threading approach you outlined. `response` is now a parameter to `_finalize_done_task` and `_build_predict_api_event`, so the on-chain path no longer round-trips through shared_state. I also capture `req_id` alongside `response` via the partial and drop the callback in `_handle_store_response` when the live executing-task id no longer matches, closing the misattribution vector at its root. Both the leak and the cross-task poisoning are removed by construction.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Dissolved. The on-chain shared_state write is gone entirely in a7384d6f, so there is no envelope with the wrong content_cid derivation and no `status` field to trip `_is_request_already_accepted` in the collision case.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Added in a7384d6f. `test_handle_done_task_onchain_use_offchain_false_leaves_shared_state_clean` pins the flag-off case, and `test_handle_store_response_invalid_done_task_does_not_leak_envelope` covers the invalid-done_task early return with a real response bound to the callback. There's also a `test_handle_store_response_stale_callback_is_dropped` that reproduces the misattribution scenario Ojus described.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Removed in a7384d6f. The whole gated write is gone, so the dead guard is too.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in a7384d6f, using both defenses. `req_id` is now captured alongside `response` via the partial (`functools.partial(self._handle_store_response, response=response, req_id=req_id)`), and `_handle_store_response` no-ops when the live executing-task id no longer matches. `test_handle_store_response_stale_callback_is_dropped` reproduces your exact repro (task A's callback landing after slot reassigned to task B) and asserts nothing gets written to either `OFFCHAIN_REQUEST_RESPONSES` or `done_tasks`.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Adopted in a7384d6f. `_finalize_done_task(cid, response=None)` and `_build_predict_api_event(..., response=None)` now take the envelope as an argument. Both off-chain and on-chain finalize pass it in, and the on-chain shared_state write is deleted. The off-chain write at line 1264 stays as you noted, since `/fetch_offchain_info` still consumes it. The duplicate mech_config lookup is also gone.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Right, removed in a7384d6f. The whole on-chain shared_state write block is gone with the parameter-threading refactor, so the dead guard goes with it.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Dropped in a7384d6f. The whole envelope write is gone, so there's nothing that looks like a status signal but isn't consumed.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed by construction in a7384d6f. No more shared_state write on the on-chain path, so nothing to leak on the invalid-done_task early return. `test_handle_store_response_invalid_done_task_does_not_leak_envelope` is a direct regression test for the scenario you described (real `response` bound to the callback, then `_done_task` cleared, then the callback fires).

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Both gaps closed in a7384d6f. Added `test_handle_done_task_onchain_use_offchain_false_leaves_shared_state_clean` for the flag-off case and `test_handle_store_response_invalid_done_task_does_not_leak_envelope` for the invalid-done_task early return. The stale-callback repro is in `test_handle_store_response_stale_callback_is_dropped`.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Removed in a7384d6f. The whole assertion is gone since nothing writes `_pending_response` anymore.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1671,22 +1676,42 @@ def _build_ipfs_get_file_req(
         )
         return message, dialogue
 
-    def _handle_store_response(self, message: IpfsMessage, dialogue: Dialogue) -> None:
+    def _handle_store_response(
+        self,
+        message: IpfsMessage,
+        dialogue: Dialogue,
+        response: Optional[Dict[str, Any]] = None,
+        req_id: Any = None,
```
jmoreira-valory, line 1684: ⚪ **F6 · NIT — req_id is typed Any, which disables the type check on the new guard**

`req_id: Any = None` opts the new comparison out of mypy entirely. The value is `executing_task["requestId"]`, which `_execute_task` normalises to `int` (bytes are converted at `behaviours.py:919-921`) before the slot is assigned, and the sibling `response` parameter right above is already precisely typed. Tightening it keeps the guard's `live_req_id != req_id` comparison under the type checker, which is where an accidental `str`/`int` drift on this exact line would show up first.

```suggestion
        req_id: Optional[int] = None,
```

jmoreira-valory (approved):

**What this does.** The mech's predict-api analytics event used to pull the tool's answer out of the `offchain_request_responses` shared-state slot, which the on-chain delivery path never wrote - so every on-chain row landed in the data lake as `status='failed'` with a null result. This revision drops the shared-state round-trip entirely and instead threads the `response` dict as a plain parameter down `_handle_store_response` -> `_finalize_done_task` -> `_build_predict_api_event`, and adds a request-id guard so a late IPFS store callback belonging to a swapped-out task is discarded instead of being attributed to whatever task now holds the slot.

**All four of my round-1 findings are resolved at `a7384d6`, most of them by construction rather than by patch.** Dropping the on-chain `shared_state` write altogether means F1 (envelope outliving its task), F2 (off-chain `content_cid` / `status` semantics reused on-chain) and F4 (dead `is_marketplace_mech` guard) have nothing left to be wrong about, and F3 is closed by three new tests (`..._use_offchain_false_leaves_shared_state_clean`, `..._stale_callback_is_dropped`, `..._invalid_done_task_does_not_leak_envelope`). @OjusWiZard's two blockers are covered by the same refactor. Net diff vs `main` is now smaller than round 1's, which is the right direction.

**What I verified at HEAD, beyond the delta:**

- CI 22/22 green, including `lock_check` on three OSes. CID cascade is complete and internally consistent - `mech_abci`, `task_submission_abci`, `agents/mech/aea-config.yaml` and `services/mech/service.yaml` all point at the new `task_execution` hash, and the two cascaded `skill.yaml`s carry dependency-hash bumps only, no source change.
- The stale-callback guard uses `req_id is not None`, not a truthiness test. That matters here: `requestId == 0` is a real value in this codebase (`_handle_timeout_task` has its own comment about `"0"` being falsy), and a truthy check would have silently disabled the guard for it. Correct as written.
- `_record_offchain_failure` resets the slot via `_reset_executing_task()` and never routes through `_finalize_done_task`, so removing the `OFFCHAIN_REQUEST_RESPONSES` read does not regress analytics on the off-chain failure path - that path never produced an event before either.
- Dropping a stale callback does not wedge the agent: `IpfsHandler.handle` clears `params.in_flight_req` after invoking the callback regardless of what the callback does (`handlers.py:592`), and the dropped task is still sitting in `pending_tasks` from the timeout re-queue.
- Off-chain behaviour is byte-identical: the object threaded into `_build_predict_api_event` is the same `response` dict that gets stored under `envelope["response"]` at `behaviours.py:1266`, so the value the event builder sees is unchanged for that path.

**One consequence worth being aware of, not a defect:** on the on-chain path `event["response"]["raw_content"]` was previously `{}` (empty envelope) and now carries the full tool response, capped at `MAX_RAW_CONTENT_BYTES` (256 KiB). That payload rides Tendermint consensus replication via `done_tasks`. It is exactly what the off-chain path already does and it is gated on `use_offchain`, but on-chain delivery volume is the higher one, so it is worth watching block sizes when the flag is turned on.

The unchecked box in the test plan - the downstream `TOOLS_TO_PACKAGE_HASH` / package-ref bump in `mech-predict` - still needs to follow this merge, otherwise the deployed mechs keep serving the old `task_execution`.

**Findings** — 2 · 🔵 1 Low · ⚪ 1 Nit

| F5 | 🔵 LOW | `packages/valory/skills/task_execution/behaviours.py:2059` | **cost_dict, params_used and execution_latency_ms still land NULL in every lake row** |
| F6 | ⚪ NIT | [packages/valory/skills/task_execution/behaviours.py:1684](https://github.com/valory-xyz/mech/blob/a7384d6fc0168d83cf6cc97fc10dd24603904b66/packages/valory/skills/task_execution/behaviours.py#L1684) | **req_id is typed Any, which disables the type check on the new guard** |

---

- 🔵 **F5 · LOW — cost_dict, params_used and execution_latency_ms still land NULL in every lake row** Now that `response` is threaded in, three more analytics fields are one line away from being correct - and the new `:param response:` docstring (lines 1859-1864) already asserts they are consumed:

> carries ``result``, ``executed_at``, ``metadata`` and ``cost_dict`` for the current task

Only `result` actually is. The other three are read off `done_task`, where nothing ever writes them:

- `cost_dict` (line 2059) reads `done_task.get("cost_dict")`. `_done_task` is built as `{...static fields..., **executing_task}` and afterwards only gains `transaction`, `dynamic_tool_cost`, `is_marketplace_mech`, `task_result` and `start_time`. The `cost_dict` produced by `TokenCounterCallback` goes onto `response`, never onto `done_task`. This is independently confirmed by `get_cost_for_done_task` (`utils/cost_calculation.py:38-41`), which logs `"Cost dict not found in done task"` and falls back to `DEFAULT_PRICE` on the same read.
- `params_used` (line 2127) reads `done_task.get("used_params")`. The resolved params live at `response["metadata"]["params"]`.
- `execution_latency_ms` (line 2125) is hardcoded `None`, while `response["metadata"]["execution_latency_ms"]` is computed and populated a few frames earlier.

Also still true, and I flagged it as a scope note last round rather than a finding: `executed_at` (line 1938) is only converted when it is an `int`/`float`, but `_handle_done_task` always writes it as an ISO 8601 string via `_iso_z`, so `executed_at` in the lake always equals `delivered_at`.

None of this blocks the fix - it is pre-existing and equally true off-chain, and the PR's stated goal (`result` / `status`) is achieved. But this PR is the one that put `response` in scope at the call site, and the docstring now claims these fields are used, so either the sourcing or the docstring should move. Suggested sourcing:

```python
response_metadata = response_data.get("metadata")
if not isinstance(response_metadata, dict):
    response_metadata = {}
cost_dict = (
    response_data.get("cost_dict")
    if isinstance(response_data.get("cost_dict"), dict)
    else None
)
```

and in the returned `"response"` block:

```python
"execution_latency_ms": response_metadata.get("execution_latency_ms"),
"params_used": (
    response_metadata.get("params")
    if isinstance(response_metadata.get("params"), dict)
    else None
),
```

If you would rather keep the field sourcing as-is and do it in a follow-up, please trim the `:param response:` docstring to `carries ``result`` for the current task` so it stops describing behaviour the function does not have.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1671,22 +1676,42 @@ def _build_ipfs_get_file_req(
...
+            # timed out or swapped out. Dropping it prevents cross-task
+            # misattribution of the response and the CID.
+            self.context.logger.warning(
+                "Discarding stale IPFS store callback for request %s; "
+                "current executing task is %s.",
+                req_id,
+                live_req_id,
+            )
+            return
+
```
OjusWiZard, line 1705: Not blocking — this closes what I raised, and the note below is pre-existing.

The guard compares `requestId` only, so it can't distinguish two *attempts* at the same request. When a task times out non-terminally it is re-queued and re-picked under the same id, and `live_req_id == req_id` then lets attempt 1's late callback through. Two outcomes depending on where attempt 2 is:

- attempt 2 still executing (`_done_task is None`) — `_finalize_done_task` takes the invalid-`done_task` early return and calls `_reset_executing_task()`, clearing the slot while the executor is still running. The task is dropped and only recovered on the next marketplace poll.
- attempt 2 already stored — the task is finalized from attempt 1. Verified against this head:

```
done_tasks published: 1
  request_id  : 100
  task_result : 8240              <- attempt 1's CID committed on chain
  event.result: ANSWER-FROM-ATTEMPT-1
executing_task after: None
done_tasks after attempt-2 callback: 1   <- attempt 2's callback dropped
```

The delivery is self-consistent where it matters — the committed CID does hash the response being reported — but `self._done_task` at that moment is attempt 2's, so `done_task["transaction"]` comes from the other attempt. I haven't traced how far that carries through `TransactionPreparationRound`, so flagging rather than asserting impact.

To be clear this is **not** a regression from this PR: before the guard, `req_id` was read live and equalled the same value in this scenario, and the same held on `main`. If anything this PR improves the second case — `event.result` now matches the committed CID instead of being null. Distinguishing attempts would need an execution nonce in the closure alongside `req_id`, which is a bigger change than belongs here. Worth a follow-up issue; happy to open one.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1925,13 +1937,178 @@ def test_handle_done_task_onchain_still_uploads_to_ipfs(
...
+    behaviour._ipfs_error_reason = None
+
+    # Task A's late store callback fires; it must not attribute A's response to B.
+    captured_callback[0](SimpleNamespace(ipfs_hash="bafyA"), MagicMock())
+
+    # Nothing lands under task B's id (the mis-attribution vector), and the
+    # slot is left untouched so task B can finalize on its own callback later.
+    assert "200" not in shared_state.get(beh_mod.OFFCHAIN_REQUEST_RESPONSES, {})
+    assert shared_state[beh_mod.DONE_TASKS] == []
+    assert behaviour._executing_task["requestId"] == 200
```
OjusWiZard, line 2065: Nit, non-blocking. I checked this test is not vacuous — deleting the guard does make it fail — but only the third assertion does that work.

- `"200" not in OFFCHAIN_REQUEST_RESPONSES` passes with or without the guard: this commit removed the only on-chain writer to that dict (the two remaining writers, at `behaviours.py:1266` and `:2184`, are both off-chain), so the on-chain path can't put anything there regardless. It asserts a behaviour that no longer exists.
- `DONE_TASKS == []` passes for an incidental reason — the second `_seed_executing_task` call resets `_done_task` to `None`, so `_finalize_done_task` would early-return even with the guard gone.
- `_executing_task["requestId"] == 200` is the one that actually pins it (without the guard it raises `TypeError: 'NoneType' object is not subscriptable`).

Reads stronger than it is. Not worth holding the PR for.

OjusWiZard (approved):
Approving. Re-reviewed at `a7384d6f`.

All seven of my earlier findings are closed, and by the better route — threading `response` down the synchronous stack rather than patching the shared-state round-trip. `OFFCHAIN_REQUEST_RESPONSES` now has exactly two writers in `behaviours.py` (`behaviours.py:1266` and `:2184`), both genuinely off-chain, and the constant's comment is accurate again.

What I verified at this head:
- CI 22/22 including `linter_checks` and `lock_check`; `test_behaviours.py` 159/159 locally.
- Only two callers of `_finalize_done_task` exist repo-wide and both pass `response`, so there is no path that reaches `_build_predict_api_event` with `response=None` and silently reintroduces `status='failed'` rows.
- Off-chain `predict_api_event` coverage survives the refactor — the pre-existing `test_handle_done_task_offchain_skips_ipfs_and_finalizes_locally` still asserts `status == "complete"` and the real `result` through the new parameter path.
- The `req_id is None` escape hatch in the guard is unreachable in production: the sole call site always binds a concrete id, and `_prepare_task` would have failed earlier without one.

Bonus: the `req_id` guard also fixes a pre-existing bug I'd scoped out of the last round — task B could previously be finalized with task A's `to_multihash(cid)`.

Two follow-up notes inline, both non-blocking and both pre-existing. Also still open from @jmoreira-valory's review and untouched here (correctly out of scope): `response["executed_at"]` is written as an ISO string but `_build_predict_api_event` only converts `int`/`float`, so it falls back to `now_iso` and `executed_at == delivered_at` on both paths — this PR restores `result` without restoring the execution timestamp.

Note this approval doesn't unblock the merge on its own — @jmoreira-valory's `CHANGES_REQUESTED` predates this commit and still gates it, though all four of their findings are addressed by it.

PR #481, opened 2026-08-27
feat(mech): thread tx hashes into predict-api events

DIvyaNautiyal07:
## Summary

Phase B of the four-repo tx-hash rollout tracked in ``eager-beaming-puddle.md``. Predict-api PR (valory-xyz/predict-api#182) ships first; this PR starts sending the two new fields on every event so mech-analytics (Phase C) can carry them into ``/v1/data/scored-rows`` and the marketplace app (Phase D) can render Etherscan links on activity rows.

Also bundles a small analytics-attribution fix for the ``[offchain request]`` prompt placeholder — details in the second section below.

## What changes

**Request payload**: ``event["request"]["request_tx_hash"]`` is sourced from ``executing_task["tx_hash"]``. The marketplace-contract wrapper at ``packages/valory/contracts/mech_marketplace/contract.py:197,294`` already captures this on on-chain requests. Absent (None) for off-chain requests — which is what predict-api's nullable schema expects and what the mech-analytics UI renders as N/A.

**Response payload**: ``event["response"]["delivery_tx_hash"]`` starts as ``None`` at task-finalize time (the settlement tx isn't known yet). Immediately before signing the delivered batch, the post-settlement behaviour stamps every event with ``synchronized_data.final_tx_hash``. All events in one round share the same tx by construction — batched off-chain settlement covers many request_ids under one ``MarketplaceDeliveryWithSignatures`` — so consumers must group by request_id, not by tx hash, when deduping.

**Batch-hash contract**: ``compute_batch_hash`` in ``predict_api_events_client.py`` now also excludes the two nested fields from the hash surface, mirroring predict-api's server-side ``model_dump`` exclusion. Without this, the enricher's post-hash write of ``delivery_tx_hash`` would invalidate the signature on the batch the mech just signed. Same rollout-safety trade-off as ``source``: an audit / display field on a row the mech's EOA already signed for. Coupled with predict-api's ``BATCH_HASH_EXCLUDED_FIELDS`` constant — a rename on either side surfaces at runtime as ``BATCH_HASH_MISMATCH`` with no compile-time signal, so the docstring on this side explicitly names the twin.

## Prompt-placeholder attribution fix (bundled)

Separate behavioural fix bundled in this PR — worth pulling out here because it changes downstream analytics counts.

Before: ``[offchain request]`` was written whenever ``prompt`` was falsy (empty or missing). That silently attributed the ~40-70 ``propose-question`` rows a day (which legitimately have ``prompt=""``) to the "no IPFS content received" bucket.

After: gate on presence, not truthiness. Three distinct shapes:

- ``request_data == {}`` — mech never received IPFS content. Placeholder: ``[offchain request]``.
- ``request_data == {"raw": ...}`` — payload was non-JSON or JSON-non-dict. New placeholder: ``[unparseable payload]`` (previously fell into ``prompt=""`` and lost the signal).
- ``request_data`` has ``prompt`` key (including ``""``) — passes through as-is. Safe because predict-api's ``MechRequestPayload.prompt`` has ``max_length`` but no ``min_length``, so ``""`` is accepted.

## Semantics

| Row shape | request_tx_hash | delivery_tx_hash |
|---|---|---|
| Off-chain delivery | None (no on-chain request event) | shared batch tx |
| On-chain per-request delivery | tx of ``MarketplaceRequest`` event | tx of ``Deliver`` event |
| Request-only sweep | tx of ``MarketplaceRequest`` event | None (no delivery happened) |

## Files

- ``packages/valory/skills/task_execution/behaviours.py`` — ``_build_predict_api_event`` emits both new fields; prompt-placeholder attribution fix.
- ``packages/valory/skills/task_submission_abci/behaviours.py`` — post-settlement enricher stamps ``delivery_tx_hash`` on delivered events before the POST. Broad ``except`` narrowed to ``ValueError`` for the documented ``final_tx_hash`` absent case; ``debug`` log level so a future FSM regression (``AttributeError`` / ``TypeError``) bubbles up rather than being silenced.
- ``packages/valory/skills/task_submission_abci/predict_api_events_client.py`` — ``compute_batch_hash`` extends exclusion set to the two nested fields; coupling comment names predict-api PR#182 as the twin.
- ``packages/valory/skills/contract_subscription/handlers.py`` — ``_get_tx_args`` threads ``tx_hash`` onto the returned args dict (primary ingest path — ``use_polling: false`` in every aea-config).
- ``packages/valory/skills/contract_subscription/tests/test_handlers.py`` — new test file, three tests pinning the ``_get_tx_args`` contract (was previously untested).
- ``packages/valory/skills/task_execution/tests/test_behaviours.py`` — three ``_build_predict_api_event`` tests (on-chain carries hash, off-chain absent, delivery placeholder shape).
- ``packages/valory/skills/task_submission_abci/tests/test_predict_api_events_client.py`` — three hash tests for the nested exclusion + regression that real fields still tip the hash.
- ``packages/valory/skills/task_submission_abci/tests/test_onchain_write_path.py`` — enricher tests for empty-string ``final_tx_hash`` warning and non-dict ``response`` warning.

Skill / agent / service hashes refreshed via ``autonomy packages lock``.

## Test plan

- [x] ``uv run pytest packages/valory/skills/task_submission_abci/tests packages/valory/skills/task_execution/tests packages/valory/skills/contract_subscription/tests`` — all pass
- [ ] Rehearse end-to-end against a staging predict-api (waiting on valory-xyz/predict-api#182 deploy).

## Sequencing

1. valory-xyz/predict-api#182 — merged + deployed.
2. **this PR** — merged next, once predict-api accepts the new fields.
3. mech-analytics (Phase C) — projects the five upstream fields (``content_cid``, ``response_cid``, ``delivery_rate``, ``request_tx_hash``, ``delivery_tx_hash``) into ``per_request_scores`` and exposes them on ``/v1/data/scored-rows``. Backfill from predict-api's DB.
4. autonolas-frontend-mono marketplace app (Phase D) — consumes behind ``NEXT_PUBLIC_USE_MECH_ANALYTICS_ROWS``.

### DIFF
--- a/packages/valory/skills/contract_subscription/handlers.py
+++ b/packages/valory/skills/contract_subscription/handlers.py
@@ -127,12 +127,22 @@ class WebSocketHandler(BaseWebSocketHandler):
             self.context.logger.info(f"Added job to queue: {event_args}")
 
     def _get_tx_args(self, tx_hash: str) -> Any:
-        """Get the transaction arguments."""
+        """Get the transaction arguments.
+
+        Threads ``tx_hash`` back onto the returned dict so downstream
+        ``_build_predict_api_event`` can populate
+        ``request.request_tx_hash``. The sibling polling path in
+        ``mech_marketplace/contract.py::get_marketplace_undelivered_reqs``
+        already emits ``tx_hash`` on every task; without it here the
+        WebSocket path (the primary ingest, per aea-config.yaml
+        ``use_polling: false``) would leave every on-chain request
+        landing at predict-api with ``request_tx_hash = NULL``.
+        """
         try:
             tx_receipt: TxReceipt = self.w3.eth.get_transaction_receipt(tx_hash)
             self._last_processed_block = tx_receipt["blockNumber"]
             rich_logs = self.contract.events.Request().processReceipt(tx_receipt)  # type: ignore
-            return dict(rich_logs[0]["args"]), False
+            return {**dict(rich_logs[0]["args"]), "tx_hash": tx_hash}, False
 
         except Exception as exc:  # pylint: disable=W0718
             self.context.logger.error(
--- a/packages/valory/skills/contract_subscription/handlers.py
+++ b/packages/valory/skills/contract_subscription/handlers.py
@@ -137,6 +137,14 @@ class WebSocketHandler(BaseWebSocketHandler):
         WebSocket path (the primary ingest, per aea-config.yaml
         ``use_polling: false``) would leave every on-chain request
         landing at predict-api with ``request_tx_hash = NULL``.
+
+        :param tx_hash: hex string of the transaction the WebSocket
+            subscription surfaced. Threaded onto the returned dict so
+            downstream can populate ``request.request_tx_hash``.
+        :return: ``(args_dict, no_request)`` — ``args_dict`` holds the
+            Request event args plus a ``tx_hash`` key; ``no_request``
+            is True on any failure so the caller stops polling for
+            args on this tx.
         """
         try:
             tx_receipt: TxReceipt = self.w3.eth.get_transaction_receipt(tx_hash)
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -2030,19 +2030,25 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             nonce_int = int(nonce_raw) if nonce_raw is not None else None
         except (TypeError, ValueError):
             nonce_int = None
-        # Placeholder ONLY when the mech never received any IPFS
-        # content for this request — an empty ``request_data`` dict
-        # means the handler couldn't populate ``IN_MEMORY_REQUESTS``,
-        # so the row would land at predict-api with no request-side
-        # context at all and needs a stand-in string.
+        # Three legitimate shapes on the way in:
         #
-        # ``request_data`` being present with ``prompt=""`` is a
-        # legitimate shape: propose-question generates its own
-        # question and has no user prompt to carry through. The
-        # previous ``prompt or "[offchain request]"`` triggered the
-        # placeholder on that legit empty-string too, silently
-        # attributing 40-70 propose-question rows a day to the
-        # "empty-IPFS" bucket in downstream analytics.
+        # 1. ``request_data == {}`` — the mech never received any IPFS
+        #    content (handler couldn't populate ``IN_MEMORY_REQUESTS``).
+        #    Row lands at predict-api with no request-side context;
+        #    stand-in placeholder makes the miss legible.
+        # 2. ``request_data == {"raw": ...}`` — the payload was
+        #    non-JSON, or was valid JSON but not a dict (see the
+        #    ``json.loads`` / ``isinstance`` fallback ~line 1895 above).
+        #    Non-empty dict but no ``prompt`` key. The previous check
+        #    (``if not request_data``) missed this case and the row
+        #    shipped ``prompt=""``, losing the "we couldn't parse
+        #    this" signal in downstream analytics — same class of bug
+        #    the empty-dict placeholder guards against, just moved to
+        #    a different bucket. Emit a distinct marker.
+        # 3. ``request_data`` present with ``prompt`` explicitly ``""``
+        #    or absent-and-present-key — a legitimate shape for
+        #    ``propose-question`` and other tools that generate their
+        #    own question. Empty string ships as-is.
         #
         # Enforce the cap on the UTF-8 BYTE length, not on ``len()``
         # (code points) — a CJK prompt of 50k code points encodes to
@@ -2051,6 +2057,8 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # in this file.
         if not request_data:
             prompt = "[offchain request]"
+        elif "prompt" not in request_data:
+            prompt = "[unparseable payload]"
         else:
             prompt = str(request_data.get("prompt") or "")
         prompt_bytes = prompt.encode("utf-8")
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -2092,6 +2092,14 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 "delivery_rate": delivery_rate_str,
                 "nonce": nonce_int,
                 "content_cid": executing_task.get("request_cid") or cid,
+                # The on-chain tx that emitted the ``MarketplaceRequest``
+                # event. The marketplace-contract wrapper captures it
+                # into ``executing_task["tx_hash"]`` for on-chain
+                # requests (see
+                # ``packages/valory/contracts/mech_marketplace/contract.py:197,294``).
+                # Legitimately absent for off-chain requests (no on-chain
+                # request event exists) — predict-api stores None → NULL.
+                "request_tx_hash": executing_task.get("tx_hash"),
                 "prompt": prompt,
                 "tool": tool,
                 "model": (
@@ -2134,6 +2142,13 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                     else {"result": result_value or "", "status": status}
                 ),
                 "response_cid": cid,
+                # Placeholder. The settlement tx isn't known here —
+                # the post-tx behaviour stamps it from
+                # ``synchronized_data.final_tx_hash`` on every
+                # delivered event before signing the batch. Left as
+                # ``None`` so pydantic on the server side accepts the
+                # payload if the enricher never ran (retry-safe).
+                "delivery_tx_hash": None,
                 "delivered_at": now_iso,
             },
             "source": predict_api_source,
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -2093,12 +2093,16 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 "nonce": nonce_int,
                 "content_cid": executing_task.get("request_cid") or cid,
                 # The on-chain tx that emitted the ``MarketplaceRequest``
-                # event. The marketplace-contract wrapper captures it
-                # into ``executing_task["tx_hash"]`` for on-chain
-                # requests (see
-                # ``packages/valory/contracts/mech_marketplace/contract.py:197,294``).
-                # Legitimately absent for off-chain requests (no on-chain
-                # request event exists) — predict-api stores None → NULL.
+                # event. Both marketplace-contract event readers
+                # (``get_marketplace_request_events`` and
+                # ``get_marketplace_undelivered_reqs`` in
+                # ``packages/valory/contracts/mech_marketplace/contract.py``)
+                # write ``tx_hash`` onto every task they emit, and the
+                # WebSocket ingestion path in
+                # ``contract_subscription/handlers.py::_get_tx_args``
+                # threads it through the primary write path. Off-chain
+                # requests legitimately have no on-chain request event
+                # so this field is absent — predict-api stores None → NULL.
                 "request_tx_hash": executing_task.get("tx_hash"),
                 "prompt": prompt,
                 "tool": tool,
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -2030,13 +2030,29 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             nonce_int = int(nonce_raw) if nonce_raw is not None else None
         except (TypeError, ValueError):
             nonce_int = None
-        # The predict-api server requires ``prompt`` to be a non-empty string;
-        # placeholder rather than dropping the row. Enforce the cap on the
-        # UTF-8 BYTE length, not on ``len()`` (code points) — a CJK prompt
-        # of 50k code points encodes to ~150 KB and would otherwise sail
-        # past the 100 KB cap; matches how ``IPFS_MAX_TASK_BYTES`` is
-        # enforced elsewhere in this file.
-        prompt = str(request_data.get("prompt") or "[offchain request]")
+        # Placeholder ONLY when the mech never received any IPFS
+        # content for this request — an empty ``request_data`` dict
+        # means the handler couldn't populate ``IN_MEMORY_REQUESTS``,
+        # so the row would land at predict-api with no request-side
+        # context at all and needs a stand-in string.
+        #
+        # ``request_data`` being present with ``prompt=""`` is a
+        # legitimate shape: propose-question generates its own
+        # question and has no user prompt to carry through. The
+        # previous ``prompt or "[offchain request]"`` triggered the
+        # placeholder on that legit empty-string too, silently
+        # attributing 40-70 propose-question rows a day to the
+        # "empty-IPFS" bucket in downstream analytics.
+        #
+        # Enforce the cap on the UTF-8 BYTE length, not on ``len()``
+        # (code points) — a CJK prompt of 50k code points encodes to
+        # ~150 KB and would otherwise sail past the 100 KB cap;
+        # matches how ``IPFS_MAX_TASK_BYTES`` is enforced elsewhere
+        # in this file.
+        if not request_data:
+            prompt = "[offchain request]"
+        else:
+            prompt = str(request_data.get("prompt") or "")
         prompt_bytes = prompt.encode("utf-8")
         if len(prompt_bytes) > MAX_PROMPT_BYTES:
             prompt = prompt_bytes[:MAX_PROMPT_BYTES].decode("utf-8", errors="ignore")
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -144,13 +144,6 @@ LAST_TX = "last_tx"
 # post-tx behaviour reads, never writes).
 PENDING_TASKS = "pending_tasks"
 PAYMENT_MODEL = "payment_model"
-# Marker for the last-enriched settlement tx hash. The FSM keeps
-# ``final_tx_hash`` in ``cross_period_persisted_keys``, so a cold
-# entry into PostTxSettlementRound without a fresh settlement would
-# otherwise re-use the prior period's hash on this period's events.
-# We stash the last hash we actually used and skip enrichment when
-# the current value matches.
-_LAST_ENRICHED_TX_HASH = "predict_api_last_enriched_tx_hash"
 
 
 class OffchainKeys(Enum):
@@ -2030,21 +2023,29 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
             # settlement in this period) that raise would escape this
             # method — which the docstring at the top of
             # ``_do_predict_api_write_best_effort`` promises never to
-            # do — and wedge the FSM in a re-entering loop. Catch it
-            # the same way the sibling access at
-            # ``check_last_tx_status`` (this file, above) does.
+            # do — and wedge the FSM in a re-entering loop. Narrowly
+            # catch ``ValueError`` (documented behaviour) at ``debug``
+            # so a broader FSM regression (``AttributeError`` from a
+            # property-chain refactor, ``TypeError`` from a corrupted
+            # DB) bubbles up as a real error rather than getting
+            # silently reported as "no settlement tx hash resolved".
             try:
                 final_tx_hash = self.synchronized_data.final_tx_hash
-            except Exception:  # noqa: BLE001 — mirrors sibling access
+            except ValueError as exc:
+                self.context.logger.debug(
+                    "final_tx_hash unavailable: %s",
+                    exc,
+                )
                 final_tx_hash = None
-            # Guard on ``is not None`` rather than truthiness so a
-            # future writer that lands ``""`` (empty string) trips a
-            # warning instead of being silently indistinguishable from
-            # "unset". Empty string is not a valid tx hash and should
-            # never appear here.
-            last_enriched = self.context.shared_state.get(
-                _LAST_ENRICHED_TX_HASH
-            )
+            # No stale-hash guard: every edge into ``PostTxSettlementRound``
+            # comes from a settlement-confirmed round, and predict-api's
+            # UPSERT is one-directional idempotent (``EXCLUDED.delivery_tx_hash
+            # IS NOT NULL AND stored delivery_tx_hash IS NULL``) so a
+            # re-POST with the same hash can never clobber. A cross-period
+            # "stale hash" isn't reachable on the FSM graph and the
+            # in-memory-only marker that used to guard it was empty
+            # right after the restart it claimed to protect (see the
+            # PR#481 review thread for the full trace).
             if final_tx_hash is None:
                 self.context.logger.info(
                     "predict-api write: no settlement tx hash resolved "
@@ -2052,22 +2053,14 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
                     "backfill will fill later."
                 )
             elif not final_tx_hash:
+                # Guard on ``is not None`` upstream and truthiness here
+                # so an empty string (never a valid tx hash) surfaces as
+                # a warning rather than being silently indistinguishable
+                # from "unset" or landing on the wire.
                 self.context.logger.warning(
                     "predict-api write: final_tx_hash resolved to %r "
                     "(non-None, falsy); skipping enrichment. "
                     "Investigate upstream FSM.",
-                    final_tx_hash,
-                )
-            elif final_tx_hash == last_enriched:
-                # This period entered PostTxSettlement without a
-                # fresh settlement — the tx hash is the one we already
-                # stamped in a prior batch. Skip rather than mis-tag
-                # this period's events with the prior period's tx.
-                self.context.logger.warning(
-                    "predict-api write: final_tx_hash=%s already "
-                    "enriched a prior batch (cross-period stale from "
-                    "cross_period_persisted_keys); skipping enrichment. "
-                    "New settlement must land before the next batch.",
                     final_tx_hash,
                 )
             else:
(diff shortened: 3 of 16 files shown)


packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2003,6 +2010,72 @@ def _do_predict_api_write_best_effort(self) -> Generator:
...
+                    "cross_period_persisted_keys); skipping enrichment. "
+                    "New settlement must land before the next batch.",
+                    final_tx_hash,
+                )
+            else:
+                for event in delivered_events:
+                    response = event.get("response")
+                    if isinstance(response, dict):
+                        response["delivery_tx_hash"] = final_tx_hash
+                self.context.shared_state[_LAST_ENRICHED_TX_HASH] = final_tx_hash
```
OjusWiZard: The marker is committed here, **before** the POST on the next line — and `_post_predict_api_batch` has four early-return failure paths (digest build, signing, JSON serialisation, network). Combined with the self-loop in `rounds.py:357-364`, that makes the tx hash permanently recoverable only by backfill:

1. Run 1 stamps the events, sets this marker, POST fails → batch dropped.
2. `PostTxSettlementRound` self-loops on `NO_MAJORITY` / `ROUND_TIMEOUT` (60s) → `async_act` re-runs.
3. `_extract_offchain_events()` re-reads `done_tasks`; `AbciAppDB.get()` returns a `deepcopy`, so run 1's in-place mutation is gone — `delivery_tx_hash` is `None` again.
4. `final_tx_hash == last_enriched` → the guard above fires → batch ships with `delivery_tx_hash=None`.

Predict-api's upsert is `... WHERE mech_responses.delivery_tx_hash IS NULL AND EXCLUDED.delivery_tx_hash IS NOT NULL` (`mech_store.py:257-263`), so that null re-POST is a no-op and the column stays NULL. (That same clause is why a re-POST can never *clobber* a stored hash — so the risk is one-directional.) This also contradicts the FSM's own comment at `rounds.py:357-364`: *"The predict-api write is idempotent, so re-entry is safe."*

Separately — I don't think the scenario this guard defends is reachable. Every edge into `PostTxSettlementRound` comes from a settlement-confirmed round (`FinishedTransactionSubmissionRound`, `ValidateTransactionRound:DONE`, `CheckTransactionHistoryRound:DONE`, `CheckLateTxHashesRound:DONE`) plus the two self-loops; the "cold entry with a stale cross-period hash" the comment describes isn't on the graph. And `SharedState` (`models.py:47`) is in-memory and non-consensus, so the marker is empty after exactly the restart it claims to protect, and diverges independently across the 4 agents.

Suggest dropping the guard entirely. If you want to keep it, move this write to after a confirmed 2xx.

Also: the comment at :2040 (*"Guard on `is not None` rather than truthiness…"*) sits above the unrelated `last_enriched = ...` line and describes a guard shape that isn't in the code — it's a four-way `if/elif` chain.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2003,6 +2010,72 @@ def _do_predict_api_write_best_effort(self) -> Generator:
...
+            # entry into PostTxSettlementRound (first boot, no prior
+            # settlement in this period) that raise would escape this
+            # method — which the docstring at the top of
+            # ``_do_predict_api_write_best_effort`` promises never to
+            # do — and wedge the FSM in a re-entering loop. Catch it
+            # the same way the sibling access at
+            # ``check_last_tx_status`` (this file, above) does.
+            try:
+                final_tx_hash = self.synchronized_data.final_tx_hash
+            except Exception:  # noqa: BLE001 — mirrors sibling access
```
OjusWiZard: This doesn't mirror the sibling — `check_last_tx_status` (:376-380) logs the exception at **error** level; this discards it. Every failure then reports as `info`: *"no settlement tx hash resolved… backfill will fill later."*

So an `AttributeError` from a future refactor of the `synchronized_data` property chain, or a `TypeError` from a corrupted DB structure, becomes indistinguishable from a routine no-settlement round and never surfaces in logs. The `info` level is right for the documented `ValueError`, but it's the wrong level for everything else `except Exception` actually catches.

```suggestion
            except ValueError as exc:  # get_strict raises when the key is absent
                self.context.logger.debug("final_tx_hash unavailable: %s", exc)
                final_tx_hash = None
```

OjusWiZard (commented):
Reviewed the logic end-to-end and verified the cross-repo claims against predict-api's actual source rather than the PR description. The threading itself is correct — `compute_batch_hash` doesn't mutate caller dicts, `sort_keys=True` keeps hashes stable for events lacking the new keys, and the `tx_hash` splat in `_get_tx_args` is safe (neither `Request` nor `MarketplaceRequest` defines a `tx_hash` arg in the built ABIs, and downstream consumers treat the dict as a plain `Dict[str, Any]`).

Three things to resolve before merge, plus inline notes.

### 1. predict-api #182 is still open — merge order is a hard gate

Sequencing §1 says #182 is "merged + deployed". It isn't. On predict-api `main`, `server/src/routes/mech.py:217` is still `event.model_dump(mode="json", exclude={"source"})` — `source` only. The nested exclusion exists solely on the unmerged `feat/mech-tx-hash-capture` branch.

If this ships first, every batch the mech signs fails `BATCH_HASH_MISMATCH`, and because the write path is best-effort, the delivered rows are dropped rather than retried. #182 also currently has a blocking review finding of its own (in its backfill script, not the exclusion logic), so please confirm the deploy actually landed rather than assuming a date.

### 2. CI will fail on `black`

Three files this PR touches fail `black --check` and are clean on `main`:

- `predict_api_events_client.py` — missing blank line after the docstring, before `def _without`
- `task_submission_abci/behaviours.py:2054` — `last_enriched = ...` fits on one line
- `tests/test_onchain_write_path.py` — the two logger lambdas, and a missing blank line after the local `_LAST_ENRICHED_TX_HASH` import

`make format` fixes all three. Please re-run `autonomy packages lock` afterwards — hashes verify clean right now, but reformatting invalidates them.

### 3. A stale claim this PR half-fixed

`task_submission_abci/behaviours.py:2612-2614` still says *"The predict-api server requires `prompt` to be a non-empty string"*. That's false — `server/src/models/mech.py:322` is `prompt: str = Field(max_length=_MAX_TEXT)` with no `min_length` (contrast `tool` on the next line, which does have `min_length=1`). This PR deleted the identical claim from `task_execution/behaviours.py`; this copy should go too.

Good news from the same check: `prompt=""` will **not** be rejected, so the core of the prompt change is safe. Two caveats — no predict-api test covers an empty prompt today, so the behaviour is unpinned and a future `min_length=1` would silently break us; and batch validation is all-or-nothing (`routes/mech.py:85` binds the whole body through pydantic), so any single bad event 422s the entire batch. Worth requesting a regression test on the predict-api side to lock it in.

### On comment volume

The production files add 75 comment lines against 77 lines of code, in multi-paragraph blocks, and the content is already in this PR description almost verbatim — which is where our convention says it belongs. Two are also inaccurate as written; I've flagged those inline. Suggest cutting these back hard and letting `git blame` → commit → this PR carry the rationale.

### Test gaps

Suite passes (794 tests). Untested: the `elif not final_tx_hash:` branch; the `shared_state[_LAST_ENRICHED_TX_HASH]` write; the `isinstance(response, dict)` false branch; and `_get_tx_args` has no coverage at all repo-wide. Inline notes on the last two.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -2030,13 +2030,29 @@ def _build_predict_api_event(
...
+        # placeholder on that legit empty-string too, silently
+        # attributing 40-70 propose-question rows a day to the
+        # "empty-IPFS" bucket in downstream analytics.
+        #
+        # Enforce the cap on the UTF-8 BYTE length, not on ``len()``
+        # (code points) — a CJK prompt of 50k code points encodes to
+        # ~150 KB and would otherwise sail past the 100 KB cap;
+        # matches how ``IPFS_MAX_TASK_BYTES`` is enforced elsewhere
+        # in this file.
+        if not request_data:
```
OjusWiZard, line 2058: Confirmed against predict-api that `prompt=""` is accepted (`models/mech.py:322` has no `min_length`), so the change is safe on that axis. But gating on dict emptiness catches a case the rationale doesn't cover.

At :1895-1901, a payload that's valid JSON but not a dict — or that fails `json.loads` — becomes `request_data = {"raw": request_data_raw}`. That's a *non-empty* dict with no `prompt` key, so it takes the `else` branch and ships `prompt=""` where it previously shipped `[offchain request]`. The "we couldn't parse this" signal is lost, which is the same class of analytics misattribution this change sets out to fix — just moved to a different bucket.

Gating on whether a usable prompt key is present, or emitting a distinct `[unparsable payload]` marker, would keep both cases legible.

One more: `propose-question` — the justification for the whole change — appears nowhere in this repo outside the files this PR touches. No tool definition, no config reference. If it lives in mech-tools, a pointer would make the rationale checkable; the "40-70 rows a day" figure isn't verifiable from here either.

packages/valory/skills/task_submission_abci/predict_api_events_client.py
```diff
@@ -91,12 +91,36 @@ def compute_batch_hash(events: List[Dict[str, Any]]) -> str:
...
     :return: the 0x-prefixed 32-byte hex digest of the canonical batch.
     """
-    hash_input = [{k: v for k, v in event.items() if k != "source"} for event in events]
+    def _without(d: Dict[str, Any], drop: str) -> Dict[str, Any]:
+        return {k: v for k, v in d.items() if k != drop}
+
+    hash_input = []
+    for event in events:
+        stripped = _without(event, "source")
+        for nested_key, drop_field in (
```
OjusWiZard, line 132: The logic is correct — `_without` builds new dicts rather than mutating the caller's events, and `canonical_json_bytes` uses `sort_keys=True`, so events without the nested keys hash byte-identically to before. Verified.

Two things about the contract this depends on:

**The server side isn't merged yet.** predict-api `main` still excludes only `source` (`routes/mech.py:217`); the matching nested exclusion lives on the unmerged #182 (`routes/mech.py:230-234`). Until that's deployed, this exclusion makes the two sides disagree and every batch fails `BATCH_HASH_MISMATCH`.

**The coupling is unguarded in both directions.** predict-api's exclusion is a hand-written dict, duplicated again in its `tests/test_mech_events.py:1435-1442` — and pydantic **silently ignores unknown keys in `exclude`**. If either side renames a field, the exclusion quietly stops applying with no error on either end; it just re-enters the hash and surfaces as `BATCH_HASH_MISMATCH` in production. Worth at least a comment naming #182 as the counterpart, so the next person renaming `delivery_tx_hash` knows there's a remote twin.

packages/valory/skills/contract_subscription/handlers.py
```diff
@@ -127,12 +127,22 @@ def handle_recv(self, message: WebsocketClientMessage) -> None:
...
+        WebSocket path (the primary ingest, per aea-config.yaml
+        ``use_polling: false``) would leave every on-chain request
+        landing at predict-api with ``request_tx_hash = NULL``.
+        """
         try:
             tx_receipt: TxReceipt = self.w3.eth.get_transaction_receipt(tx_hash)
             self._last_processed_block = tx_receipt["blockNumber"]
             rich_logs = self.contract.events.Request().processReceipt(tx_receipt)  # type: ignore
-            return dict(rich_logs[0]["args"]), False
+            return {**dict(rich_logs[0]["args"]), "tx_hash": tx_hash}, False
```
OjusWiZard, line 153: This line is the primary ingest path for the whole `request_tx_hash` feature — `use_polling` is `false` across the agent, service, and skill configs — and it has no test coverage. There's no `packages/valory/skills/contract_subscription/tests/` directory at all, so nothing in the suite exercises `_get_tx_args` before or after this change.

The splat itself is safe (checked the built ABIs: neither `Request` nor `MarketplaceRequest` defines a `tx_hash` arg, so no silent overwrite, and the format matches the polling path's `0x`-prefixed lowercase hex). But a single test asserting the returned dict carries `tx_hash` would pin the one behaviour the rest of the feature is built on.

packages/valory/skills/task_submission_abci/tests/test_onchain_write_path.py
```diff
@@ -700,6 +700,175 @@ def test_egress_gate_off_with_no_events_stays_silent() -> None:
...
+        PostTxSettlementBehaviour._do_predict_api_write_best_effort(self_)
+    )
+    # Every event's response was stamped, and the POST was made with
+    # the same batch we mutated (arg-alias — behaviour passes the
+    # same list).
+    assert all(
+        e["response"]["delivery_tx_hash"] == hash_value for e in delivered_events
+    )
+    assert len(self_._post_calls) == 1  # type: ignore[attr-defined]
+    assert self_._post_calls[0]["events"] is delivered_events  # type: ignore[attr-defined]
```
OjusWiZard, line 815: This test never asserts that `shared_state[_LAST_ENRICHED_TX_HASH]` was written — so deleting `behaviours.py:2078` outright fails nothing here, even though the stale-match test below depends entirely on that write happening.

```suggestion
    assert self_._post_calls[0]["events"] is delivered_events  # type: ignore[attr-defined]
    assert self_.context.shared_state[_LAST_ENRICHED_TX_HASH] == hash_value
```

Also untested in this file: the `elif not final_tx_hash:` branch (`behaviours.py:2054-2060`) — nothing constructs `final_tx_hash=""`, so that whole warning path is dead weight until something covers it — and the `isinstance(response, dict)` false branch at :2074, for an event whose `response` key is missing or isn't a dict.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2003,6 +2010,72 @@ def _do_predict_api_write_best_effort(self) -> Generator:
...
+                    "backfill will fill later."
+                )
+            elif not final_tx_hash:
+                self.context.logger.warning(
+                    "predict-api write: final_tx_hash resolved to %r "
+                    "(non-None, falsy); skipping enrichment. "
+                    "Investigate upstream FSM.",
+                    final_tx_hash,
+                )
+            elif final_tx_hash == last_enriched:
```
Tanya-atatakai: tracing the retry path to its end: if `_post_predict_api_batch` fails and the same deliveries are re-extracted next round — are the event dicts the same objects (stamp survives) or rebuilt from state (stamp lost)? if rebuilt AND no new settlement landed, this guard skips re-stamping and the retry ships `delivery_tx_hash=None`; if rebuilt AND a new settlement DID land, the old batch gets the new tx stamped — wrong attribution. or the retry never happens and I'm missing something? lmk how the re-post path actually behaves

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -144,6 +144,13 @@
 # post-tx behaviour reads, never writes).
 PENDING_TASKS = "pending_tasks"
 PAYMENT_MODEL = "payment_model"
+# Marker for the last-enriched settlement tx hash. The FSM keeps
+# ``final_tx_hash`` in ``cross_period_persisted_keys``, so a cold
+# entry into PostTxSettlementRound without a fresh settlement would
+# otherwise re-use the prior period's hash on this period's events.
+# We stash the last hash we actually used and skip enrichment when
+# the current value matches.
+_LAST_ENRICHED_TX_HASH = "predict_api_last_enriched_tx_hash"
```
Tanya-atatakai: tiny: `shared_state` is per-process, so an agent restart loses `_LAST_ENRICHED_TX_HASH` — the exact stale-cross-period hazard the guard defends against silently comes back on the first round after a restart. acceptable (backfill catches it), or worth persisting? jsyk

packages/valory/skills/task_execution/behaviours.py
Tanya-atatakai, line 2058: the `[offchain request]` placeholder change (empty `prompt` now passes through for propose-question) is a real behavior fix but it's not mentioned anywhere in the PR description — can we add it there (or split it out)? fwiw I checked predict-api's `MechRequestPayload.prompt` on main: `Field(max_length=…)` with no `min_length`, so `""` is accepted and the deleted "server requires non-empty" comment was indeed wrong — the change is safe

Tanya-atatakai (dismissed):
approving with caveats — the hash-exclusion work is correct and well-tested from both sides (I cross-checked `compute_batch_hash`'s nested strip against predict-api's `BATCH_HASH_EXCLUDED_FIELDS`; the old-mech/null/missing shapes hash equal in both repos). the retry-path question below is the one I'd like answered before merge.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Dropped the guard entirely in `08b4b834` — you and Ojus independently traced the same non-reachability argument. Re-post safety now hangs on predict-api's one-directional UPSERT (`WHERE stored delivery_tx_hash IS NULL AND EXCLUDED.delivery_tx_hash IS NOT NULL`) which can't clobber a stored hash and can't zero a NULL back to NULL. Your retry scenario "rebuilt AND new settlement landed" would still tag the retry batch with the new hash, but the "wrong attribution" you flag is actually correct: the deliveries in that batch were re-settled under the new tx.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Same commit — dropped the marker + the guard. `shared_state` per-process meant the marker was empty on the exact restart it claimed to protect, which we agreed was safety theatre.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Added to the PR description under "Prompt-placeholder attribution fix (bundled)" — walks through the three shapes (`{}` → `[offchain request]`, `{"raw": ...}` → new `[unparseable payload]`, `prompt` key present → pass-through). Thanks for verifying the `min_length` on `MechRequestPayload.prompt`; came to the same conclusion.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Dropped the guard entirely in `08b4b834` — you're right on every leg of the trace. `_LAST_ENRICHED_TX_HASH` (the module constant + the shared_state write + the elif branch) is gone. Re-post safety hangs on predict-api's one-directional UPSERT (`WHERE stored.delivery_tx_hash IS NULL AND EXCLUDED.delivery_tx_hash IS NOT NULL`) which can't clobber a stored hash and can't zero a NULL back to NULL; the cross-period stale-hash scenario isn't reachable on the FSM graph; and shared_state is in-memory / non-consensus so the marker was empty on the exact restart it claimed to protect anyway. Also dropped the stale "Guard on `is not None`" comment that sat above unrelated code.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Narrowed in `08b4b834` — `except ValueError as exc` with `self.context.logger.debug("final_tx_hash unavailable: %s", exc)`. `AttributeError` from a property-chain refactor and `TypeError` from a corrupted DB structure now bubble as real errors instead of being silently reported as "no settlement tx hash resolved". Comment on the try/except explains which exception is documented behaviour and why the narrower catch is safer than the sibling.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed in `08b4b834`. The gate now checks three shapes explicitly:

* `request_data == {}` → `[offchain request]` (mech never received any IPFS content).
* `"prompt" not in request_data` → new `[unparseable payload]` placeholder. Covers the `{"raw": ...}` shape the JSON-non-dict fallback at :1895-1901 produces — previously fell into `prompt=""` and lost the signal, same class of analytics misattribution the change set out to fix.
* `prompt` key present (including `""`) → pass-through. propose-question sails through as intended.

On verifiability: the tool lives at `packages/valory/customs/propose_question/` in `valory-xyz/mech-predict` (not `mech-tools` as I said in an earlier commit message — my mistake). The 40-70/day figure was from an ad-hoc `predict-api` SQL count I ran during scoping and don't have saved; I shouldn't cite a specific number here without a reproducible source. The behavioural point stands — `propose-question` legitimately writes `prompt=""` because it generates its own question — but if you want the actual daily volume before signing off, happy to re-run the count against `mech_requests` and post it.

packages/valory/skills/task_submission_abci/predict_api_events_client.py
DIvyaNautiyal07: Verified against `main` on predict-api — you're right the server-side exclusion was on the unmerged #182 at the time of your review; that PR is now merged so the two sides agree in prod. On the coupling: added an explicit ``.. warning::`` block in `compute_batch_hash`'s docstring in `08b4b834` naming predict-api's `BATCH_HASH_EXCLUDED_FIELDS` as the twin, calling out the pydantic silently-ignores-unknown-keys drift risk, and listing the file paths on both sides. A rename here without matching the counterpart still surfaces at runtime as `BATCH_HASH_MISMATCH`, but the next person renaming has a signpost to check first.

packages/valory/skills/contract_subscription/handlers.py
DIvyaNautiyal07: Addressed in `08b4b834`. New `packages/valory/skills/contract_subscription/tests/test_handlers.py` with three tests: (1) legit Request emission carries `tx_hash` on the returned args dict + `no_request=False`, (2) tx that doesn't emit a Request (`processReceipt` returns `[]`) lands as `({}, True)`, (3) RPC failure on the receipt lookup also lands as `({}, True)` so the caller stops polling on the same bad tx. First test is the primary regression guard — a refactor that dropped `"tx_hash": tx_hash` from the returned dict would flip it.

packages/valory/skills/task_submission_abci/tests/test_onchain_write_path.py
DIvyaNautiyal07: Both branches covered in `08b4b834`:

* `elif not final_tx_hash:` (empty-string case) — new `test_enricher_warns_and_skips_when_final_tx_hash_is_empty_string`. Asserts the warning fires, the event's `delivery_tx_hash` stays NULL, and the batch still ships (best-effort contract). Without this branch, `""` would either ride through as a bogus tx hash or hit the `else` and stamp `response["delivery_tx_hash"] = ""`.
* `isinstance(response, dict)` false branch — new `test_enricher_warns_when_response_is_not_a_dict_and_ships_batch`. Two events, one with a dict response, one with a scalar. Asserts the dict lands the hash, the scalar is skipped intact, the warning names the type, and the batch ships. Also added an `else` branch on the isinstance check so the shape bug logs a warning instead of silently dropping.

Deleted `test_enricher_skips_when_final_tx_hash_matches_prior_period` since `_LAST_ENRICHED_TX_HASH` is gone.

packages/valory/skills/task_submission_abci/predict_api_events_client.py
OjusWiZard: The coupling half is handled well — naming `BATCH_HASH_EXCLUDED_FIELDS` as the twin that has to move in the same rollout is exactly the signal that was missing, and it points a future renamer at the one file that can't drift silently.

Leaving this thread open for the other half, which is a merge gate rather than a code change: predict-api #182 is still `OPEN` (`mergedAt: null`) as of this comment. Until it merges *and* deploys, predict-api `main` excludes only `source` (`routes/mech.py:217`), so this side's nested exclusion makes the two disagree and every signed batch fails `BATCH_HASH_MISMATCH` — dropped rather than retried, since the write path is best-effort.

Nothing to do here in code. Worth re-checking #182 actually deployed rather than just merged before this one goes in, and I'll resolve this once it has.

OjusWiZard (approved):
Approving the code. The response to the last round was genuinely good — the stale-hash guard was **removed** rather than patched around, the exception was narrowed to `ValueError` with the exception actually logged, and the branches that were bare got real tests instead of coverage padding. Resolved my five threads.

One merge gate and three small follow-ups. None of them need another review round.

---

### ⛔ Do not merge until predict-api #182 has merged **and** deployed

This is the one thing I'd ask you not to take on trust. Re-checked twice while writing this:

```
gh pr view 182 --repo valory-xyz/predict-api
  state:       OPEN
  mergedAt:    null
  closedAt:    null
  mergeCommit: null
```

And predict-api `main` at `server/src/routes/mech.py:217` is still:

```python
event.model_dump(mode="json", exclude={"source"}) for event in body.events
```

No `BATCH_HASH_EXCLUDED_FIELDS`, no nested exclusion. So as of right now the two sides do **not** agree in prod — merging this first means every signed batch fails `BATCH_HASH_MISMATCH`, and since the write path is best-effort those delivered rows are dropped rather than retried.

Flagging it plainly because the reply on the batch-hash thread said #182 was already merged; it isn't yet, and this is a silent-until-it-isn't failure mode. Worth confirming the **deploy** landed too, not just the merge — the exclusion only takes effect once the new server is actually serving.

The approval stands on the code as written; the sequencing is the part to hold.

---

### Three follow-ups (non-blocking)

**1. `[unparseable payload]` has no test.** `grep -rn "unparseable" packages/` returns exactly one hit — `task_execution/behaviours.py:2061` — and nothing covering it. Every other fix this round picked up a test (`test_enricher_warns_and_skips_when_final_tx_hash_is_empty_string`, `test_enricher_warns_when_response_is_not_a_dict_and_ships_batch`, the three new `_get_tx_args` tests), so this one reads like an oversight rather than a call. A `request_data={"raw": "<not json>"}` case alongside the existing placeholder tests would close it.

**2. Stale docstring**, `task_execution/tests/test_behaviours.py:4032`:

> The new gate only substitutes the placeholder when `request_data` is empty at the dict level.

That was true of the previous revision; there are two placeholder branches now.

**3. `task_submission_abci/behaviours.py:2619` still carries the claim this PR disproved:**

```python
# The predict-api server requires ``prompt`` to be a non-empty
# string and ``tool`` to be a non-empty string.
```

`MechRequestPayload.prompt` is `Field(max_length=…)` with no `min_length` — you and Tanya both verified this independently. The twin copy in `task_execution/behaviours.py` was deleted; this one survived the sweep. Since the sweep path's own placeholders (`[onchain undelivered]` / `unknown`) don't depend on the claim being true, deleting the first sentence is enough.

---

### One coordination note for Phase C/D

`[unparseable payload]` is a **new sentinel string entering the data lake**. If mech-analytics or the marketplace app branch on `[offchain request]` to identify placeholder rows, they now need to handle two values. Worth a heads-up on those PRs before they build the display logic, rather than after.

---

Verified locally at `c63a249e`: `black --check` clean across all three packages, 798 tests pass including the new `contract_subscription/tests` module. CI is green on all 21 checks.

packages/valory/skills/task_submission_abci/predict_api_events_client.py
DIvyaNautiyal07: Agreed on the merge gate — this side won't land until predict-api #182 is merged AND the server is deployed. Sequencing is #182 → deploy → this PR → deploy → mech-analytics #37. I'll block the merge here on the #182 deployment signal and won't wave this through on a "merged but not deployed" state.

Nothing to change in code on this side — thread's a correct gate on the rollout, not a review-blocker on the diff. Leaving open so it's visible to the merger.

Tanya-atatakai (approved):

PR #482, opened 2026-09-01
fix(task_submission): drop done_tasks from cross-period persistence

DIvyaNautiyal07:
## Summary

Splits the cross-period hand-off between `TaskPoolingRound` and `TaskPoolingBehaviour.handle_submitted_tasks` out of the fat `done_tasks` field into a new id-only `submitted_request_ids` field. Reduces the serialized `synchronized_data.db` size by the one carried copy per period that `AbciAppDB.create()` used to seed. Under typical single-event batches a `RegistrationPayload.encode()` that used to reject at the 1 MiB cap now fits.

Does not fully bound the serialized size: a batch with two max-size predict-api events in the same period still rejects. Bounding it in the strict sense requires keeping `predict_api_event` out of consensus entirely (follow-up).

## Why

`done_tasks` carries the full request/response payload per event and is in `cross_period_persisted_keys`. On period boundary `AbciAppDB.create()` seeds the new period with the previous latest value, and `AbciAppDB.update()` then appends the new period's own write; both remain in the retained period's per-key value history because `mech_abci/skill.yaml` sets `cleanup_history_depth_current: null`. So a period ends up holding two fat blobs (carried + fresh).

`serialize()` walks every retained period. `RegistrationPayload.encode()` calls it. Once the per-event payload grew (predict-api rollout on Jul 8), the two-blobs-per-period pattern was enough to push a single-event batch over `MAX_READ_IN_BYTES = 2**20`. The failure is silent until a mech restarts (or takes the reset-error → Registration path), at which point the payload is rejected and the FSM stalls. TM replay reconstructs the same fat DB.

Removing `done_tasks` from `cross_period_persisted_keys` halves the retained size per period (only the fresh write, no carry). The id-only `submitted_request_ids` carries the ~50 bytes needed for the next cycle's prune.

## Measured

Against real `AbciAppDB` with 500 KB events on the reset-error path (no `create()` / `cleanup()`):

| events in batch | pre-PR | post-PR |
|---|---|---|
| 1 | 1465 KB — exceeds | 977 KB — fits |
| 2 | 2930 KB — exceeds | 1954 KB — still exceeds |

## Changes

- `SynchronizedData.submitted_request_ids`: new property. Reads `[]` when unset. Fail-soft on invariant drift: a non-`list[str]` value logs via `self.db.logger.error` (which is the configured AEA logger, so it lands in `log.txt`) and yields `[]` rather than raising. The field is cross-period-persisted and read on every FSM cycle across all participants — a raise would crash-loop the whole fleet with no in-band recovery.
- `PostTxSettlementRound.end_block`: writes `submitted_request_ids` only when `done_tasks` is non-empty. `composition.py` wires `FinishedTransactionSubmissionRound` here for every settlement, so the delivery-rate settlement path arrives with `done_tasks == []`; overwriting a still-pending hand-off in that case would clobber a prior real settlement's ids. Emits a `warning` naming the pending id list so this benign case is distinguishable from an actual bug landing on the same branch.
- `TaskPoolingRound.end_block`: consumes the hand-off (`submitted_request_ids = []`) on both DONE and NO_TASKS. Every participant's `handle_submitted_tasks` has already read and pruned by the time the round accepts quorum, so clearing here makes the field one-shot per settlement. Without this, the "already submitted" block re-runs on stale ids every period, silently re-pruning re-swept requests before they can be pooled.
- `cross_period_persisted_keys`: swap `done_tasks` → `submitted_request_ids`.
- `TaskPoolingBehaviour.handle_submitted_tasks`: `remove_tasks_by_id` now returns the pre-prune snapshot; the caller drives its metrics loop from that snapshot. One lock acquisition, one tested surface. Prune runs before the `_fetch_tx_block_number` yield so a cancellation there can't leave shared state un-pruned. Prune log fires next to the prune, past tense.
- `remove_tasks_by_id(submitted_ids: List[str])`: tightened from `Sequence[Any]`. `str` satisfies `Sequence[str]`, so the old annotation type-checked a bare `str` and character-split it.
- Per-task `tool_delivery_time` histogram reads from local `shared_state[DONE_TASKS]`. `start_time` is `time.perf_counter()` (process-relative), so pre-PR subtracting another agent's counter produced meaningless durations; this is arguably a correctness fix. Any dashboard threshold keyed on this histogram's rate will shift on deploy (sample count drops by roughly N in an N-agent service).

## What's preserved

- On-chain requests: source of truth is the marketplace contract; re-fetched and re-executed on restart as before.
- Offchain in-memory requests: unchanged.
- Predict-api POST: reads `done_tasks` in the same period as it was set, unchanged.
- Multi-agent idempotency: server dedups on `request_id`.

## Tests

Round side:
- `SynchronizedData.submitted_request_ids` property (default, stored value, invariant-broken → `[]` with `error` log)
- `PostTxSettlementRound.end_block` writes ids on DONE, does not write when `done_tasks` empty (pending hand-off preserved), str-normalises mixed-type ids, skips falsy ids
- `TaskPoolingRound.end_block` clears `submitted_request_ids` on both DONE and NO_TASKS
- `cross_period_persisted_keys` schema invariant (contains `submitted_request_ids`, excludes `done_tasks`)

Behaviour side:
- `remove_tasks_by_id` (removes, missing ids no-op, legacy `int` in shared_state matches `str` id, returns pre-prune snapshot)
- `handle_submitted_tasks` primary path (id list drives shared_state prune), empty-ids no-op, `status=False` leaves shared_state untouched, histogram + shared_state assertions on the with-tasks path
- Skip-branch coverage: task absent from shared_state, task missing `tool` / `start_time`

Mutation-catching:
- `test_prune_acquires_done_tasks_lock` — dropping the `with` around the prune fails.
- `test_str_normalisation_matches_legacy_int_on_live_path` — dropping the `str()` on the shared_state side of the match fails on the hot path.
- `test_prune_runs_before_block_number_fetch` — moving the prune below the `_fetch_tx_block_number` yield fails.

End-to-end:
- `TestSubmittedRequestIdsRoundTrip` uses a real `SynchronizedData` (not `MagicMock`) so the property is actually exercised on the behaviour path; asserts the prune reaches `shared_state[DONE_TASKS]`.

## Follow-ups (out of scope)

- Keep `predict_api_event` out of consensus entirely, to truly bound the serialized payload regardless of event count per batch.
- Extract a shared `normalise_request_id(task)` helper for the three sites in `rounds.py` (extract_request_ids filter, dedup loop, sort key).
- Extract the metrics loop in `handle_submitted_tasks` to a plain helper.
- Labelled counter to separate "task not in shared_state (executed by another agent)" from "malformed skip".

## Test plan

- [x] Unit tests pass locally (`pytest packages/valory/skills/task_submission_abci/tests packages/valory/skills/task_execution/tests packages/valory/skills/mech_abci/tests`, 903 tests)
- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e darglint -e mypy` clean
- [x] `tomte tox -e check-abci-docstrings -e check-abciapp-specs` OK
- [x] `autonomy packages lock --check` OK
- [x] `tomte tox -e check-hash` OK
- [ ] CI: full pipeline
- [ ] Post-merge: canary on a Gnosis mech; confirm DB serialization stays under 1 MiB through many periods and reset-error cycles

### DIFF
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -344,9 +344,9 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
         if not status:
             return
 
-        submitted_ids = list(
-            cast(SynchronizedData, self.synchronized_data).submitted_request_ids
-        )
+        submitted_ids = cast(
+            SynchronizedData, self.synchronized_data
+        ).submitted_request_ids
         if len(submitted_ids) == 0:
             return
 
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -228,30 +228,35 @@ class TaskExecutionBaseBehaviour(BaseBehaviour, ABC):
 
         :param submitted_tasks: the done tasks that have already been submitted
         """
+        # Extract ids and delegate; both entry points share
+        # ``remove_tasks_by_id`` so match semantics stay consistent.
+        submitted_ids = [
+            str(task["request_id"])
+            for task in submitted_tasks
+            if task.get("request_id") is not None and task.get("request_id") != ""
+        ]
+        self.remove_tasks_by_id(submitted_ids)
+
+    def remove_tasks_by_id(self, submitted_ids: List[str]) -> None:
+        """Pop already-submitted tasks from shared state, matched by request_id.
+
+        :param submitted_ids: request ids of tasks already delivered.
+        """
         # run this in a lock
         # the amount of done tasks will always be relatively low (<<20)
         # we can afford to do this in a lock
+        submitted_id_set = {str(rid) for rid in submitted_ids}
         with self.done_tasks_lock():
             done_tasks = self.done_tasks
-            not_submitted = []
-            # Normalize both sides to ``str`` before comparing so a mix of
-            # ``int`` (on-chain requestId) and ``str`` (pre-ingress-fix
-            # off-chain requestId) doesn't silently fail equality and leave
-            # a delivered task stuck in ``done_tasks`` for the next round to
-            # re-emit. Post the :mod:`task_execution.handlers` ingress fix
-            # both sides are ``int`` and the ``str`` cast is a no-op; the
-            # cast is retained for the same restart-transition reason
-            # called out in :meth:`task_submission_abci.rounds.
-            # TaskPoolingRound.end_block`.
-            for done_task in done_tasks:
-                done_key = str(done_task["request_id"])
-                is_submitted = False
-                for submitted_task in submitted_tasks:
-                    if str(submitted_task["request_id"]) == done_key:
-                        is_submitted = True
-                        break
-                if not is_submitted:
-                    not_submitted.append(done_task)
+            # ``str``-normalise the match key. The shared-state side
+            # may hold ``int`` or ``str`` ``request_id`` (mixed types
+            # can survive an in-place restart transition); normalising
+            # both sides here keeps the prune consistent regardless.
+            not_submitted = [
+                done_task
+                for done_task in done_tasks
+                if str(done_task.get("request_id")) not in submitted_id_set
+            ]
             self.context.shared_state[DONE_TASKS] = not_submitted
 
     @property
@@ -323,43 +328,87 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
         return []
 
     def handle_submitted_tasks(self) -> Generator[None, None, None]:
-        """Handle tasks that have been already submitted before (in a prev. period)."""
+        """Handle tasks that have been already submitted before (in a prev. period).
+
+        Reads ``submitted_request_ids`` for the id list, or falls back
+        to ``done_tasks`` when the id field isn't yet in the DB (a
+        mech that resumes with pre-existing state written before this
+        code shipped will see ``None`` on the first read). Per-task
+        metrics are read from ``shared_state[DONE_TASKS]`` since the
+        id-only hand-off doesn't carry tool / start_time.
+
+        :yield: AEA protocol messages while fetching the tx block number.
+        """
         status, tx_hash = self.check_last_tx_status()
         self.context.logger.info(f"Last tx status is: {status}")
-        if status:
-            submitted_tasks = cast(
+        if not status:
+            return
+
+        # Fallback path for state that predates
+        # ``submitted_request_ids`` being written. Applies for at most
+        # one cycle per mech; subsequent cycles read the id list
+        # written by the previous ``PostTxSettlementRound``.
+        submitted_ids_raw = self.synchronized_data.db.get("submitted_request_ids", None)
+        if submitted_ids_raw is None:
+            self.context.logger.info(
+                "submitted_request_ids missing; falling back to done_tasks."
+            )
+            submitted_tasks_fallback = cast(
                 List[Dict[str, Any]], self.synchronized_data.done_tasks
             )
-            if len(submitted_tasks) > 0:
-                for task in submitted_tasks:
-                    req_id = task["request_id"]
-                    tool = task["tool"]
-                    start_time = task["start_time"]
-                    tool_delivery_time_duration = time.perf_counter() - start_time
-                    self.context.logger.info(
-                        f"Request id: {req_id} with tool: {tool} took {tool_delivery_time_duration} seconds to complete delivery"
-                    )
-                    self.observe_histogram(
-                        self.shared_state.tool_delivery_time,
-                        tool_delivery_time_duration,
-                        tool=tool,
-                    )
+            submitted_ids = [
+                str(task["request_id"])
+                for task in submitted_tasks_fallback
+                if task.get("request_id") is not None and task.get("request_id") != ""
+            ]
+        else:
+            submitted_ids = [str(rid) for rid in cast(List[Any], submitted_ids_raw)]
 
-                block_number = yield from self._fetch_tx_block_number(tx_hash)
-                self.context.logger.info(
-                    f"Block number for tx hash: {tx_hash} is {block_number}"
-                )
-                if block_number:
-                    self.set_gauge(
-                        self.shared_state.mech_delivery_last_block_number, block_number
-                    )
+        if len(submitted_ids) == 0:
+            return
 
+        # Emit per-task delivery-time metrics from in-memory state.
+        # A task not found in ``shared_state[DONE_TASKS]`` still
+        # contributes to the prune below, but its metric is skipped;
+        # the delivery already landed on-chain, only the timing
+        # dimension is dropped.
+        shared_done_tasks_by_id = {
+            str(task.get("request_id")): task for task in self.done_tasks
+        }
+        for req_id in submitted_ids:
+            task = shared_done_tasks_by_id.get(req_id)
+            if task is None:
+                continue
+            tool = task.get("tool")
+            start_time = task.get("start_time")
+            if tool is None or start_time is None:
+                continue
+            tool_delivery_time_duration = time.perf_counter() - float(start_time)
             self.context.logger.info(
-                f"Tasks {submitted_tasks} has already been submitted. The corresponding tx_hash is: {tx_hash}. "
-                f"Removing them from the list of tasks to be processed."
+                f"Request id: {req_id} with tool: {tool} took "
+                f"{tool_delivery_time_duration} seconds to complete delivery"
+            )
+            self.observe_histogram(
+                self.shared_state.tool_delivery_time,
+                tool_delivery_time_duration,
+                tool=tool,
             )
 
-            self.remove_tasks(submitted_tasks)
+        block_number = yield from self._fetch_tx_block_number(tx_hash)
+        self.context.logger.info(
+            f"Block number for tx hash: {tx_hash} is {block_number}"
+        )
+        if block_number:
+            self.set_gauge(
+                self.shared_state.mech_delivery_last_block_number, block_number
+            )
+
+        self.context.logger.info(
+            f"Tasks with ids {submitted_ids} have been submitted. "
+            f"The corresponding tx_hash is: {tx_hash}. "
+            f"Removing them from the list of tasks to be processed."
+        )
+        self.remove_tasks_by_id(submitted_ids)
 
     def check_last_tx_status(self) -> Tuple[bool, str]:
         """Check if the tx in the last round was successful or not"""
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -33,7 +33,6 @@ from typing import (
     Generator,
     List,
     Optional,
-    Sequence,
     Set,
     Tuple,
     Type,
@@ -242,15 +241,12 @@ class TaskExecutionBaseBehaviour(BaseBehaviour, ABC):
         """
         self.remove_tasks_by_id(extract_request_ids(submitted_tasks))
 
-    def remove_tasks_by_id(self, submitted_ids: Sequence[Any]) -> None:
+    def remove_tasks_by_id(self, submitted_ids: List[str]) -> None:
         """Pop already-submitted tasks from shared state, matched by request_id.
 
         :param submitted_ids: request ids of tasks already delivered.
-            Accepts ``int`` or ``str`` shapes; both sides are
-            ``str``-normalised on the equality check.
         """
-        # Amount of done tasks is small (<<20); the lock cost is trivial.
-        submitted_id_set = {str(rid) for rid in submitted_ids}
+        submitted_id_set = set(submitted_ids)
         with self.done_tasks_lock():
             done_tasks = self.done_tasks
             not_submitted = [
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -340,9 +340,7 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
         if not status:
             return
 
-        submitted_ids = cast(
-            SynchronizedData, self.synchronized_data
-        ).submitted_request_ids
+        submitted_ids = self.synchronized_data.submitted_request_ids
         if len(submitted_ids) == 0:
             return
 
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -241,20 +241,27 @@ class TaskExecutionBaseBehaviour(BaseBehaviour, ABC):
         """
         self.remove_tasks_by_id(extract_request_ids(submitted_tasks))
 
-    def remove_tasks_by_id(self, submitted_ids: List[str]) -> None:
-        """Pop already-submitted tasks from shared state, matched by request_id.
+    def remove_tasks_by_id(self, submitted_ids: List[str]) -> List[Dict[str, Any]]:
+        """Pop already-submitted tasks from shared state and return the pre-prune snapshot.
+
+        Returning the snapshot lets a caller drive per-task metrics
+        from the same list the prune ran on, without acquiring the
+        lock twice or risking a torn read against the background
+        executor. The snapshot is a ``deepcopy`` of the pre-prune
+        ``shared_state[DONE_TASKS]``.
 
         :param submitted_ids: request ids of tasks already delivered.
+        :return: the pre-prune snapshot of ``shared_state[DONE_TASKS]``.
         """
         submitted_id_set = set(submitted_ids)
         with self.done_tasks_lock():
-            done_tasks = self.done_tasks
-            not_submitted = [
-                done_task
-                for done_task in done_tasks
-                if str(done_task.get("request_id")) not in submitted_id_set
+            snapshot = self.done_tasks
+            self.context.shared_state[DONE_TASKS] = [
+                task
+                for task in snapshot
+                if str(task.get("request_id")) not in submitted_id_set
             ]
-            self.context.shared_state[DONE_TASKS] = not_submitted
+        return snapshot
 
     @property
     def mech_addresses(self) -> List[str]:
@@ -344,28 +351,22 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
         if len(submitted_ids) == 0:
             return
 
-        # Take one snapshot under the lock and prune inline, so the
-        # metrics loop below and the prune both operate on the same
-        # entries. ``shared_state[DONE_TASKS]`` is written concurrently
-        # by the background executor, so an unlocked
-        # ``deepcopy(shared_state[DONE_TASKS])`` can produce a torn
-        # read: a task landing mid-copy would silently fall into the
-        # ``task is None`` skip branch below and its metric would be
-        # dropped, making a real desync indistinguishable from a race.
-        submitted_id_set = set(submitted_ids)
-        with self.done_tasks_lock():
-            shared_snapshot = list(self.context.shared_state.get(DONE_TASKS, []))
-            self.context.shared_state[DONE_TASKS] = [
-                task
-                for task in shared_snapshot
-                if str(task.get("request_id")) not in submitted_id_set
-            ]
+        # Prune + snapshot under one lock via ``remove_tasks_by_id``.
+        # Sharing the snapshot with the metrics loop below keeps the
+        # torn-read window closed (the background executor writes
+        # ``shared_state[DONE_TASKS]`` under the same lock) and keeps
+        # both paths on the single tested prune surface.
+        shared_snapshot = self.remove_tasks_by_id(submitted_ids)
+        self.context.logger.info(
+            f"Pruned tasks with ids {submitted_ids} from shared state; "
+            f"tx_hash={tx_hash}."
+        )
 
-        # Emit per-task delivery-time metrics from the snapshot. A
-        # task absent from ``shared_state[DONE_TASKS]`` at snapshot
-        # time still counted toward the prune above; the metric is
-        # skipped only because the timing data isn't locally
-        # available (the delivery already landed on-chain).
+        # Emit per-task delivery-time metrics from the same snapshot.
+        # A task absent from the snapshot still counted toward the
+        # prune above; the metric is skipped only because the local
+        # timing data isn't available (the delivery already landed
+        # on-chain).
         shared_done_tasks_by_id = {
             str(task.get("request_id")): task for task in shared_snapshot
         }
@@ -408,12 +409,6 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
             self.set_gauge(
                 self.shared_state.mech_delivery_last_block_number, block_number
             )
-
-        self.context.logger.info(
-            f"Tasks with ids {submitted_ids} have been submitted. "
-            f"The corresponding tx_hash is: {tx_hash}. "
-            f"Removing them from the list of tasks to be processed."
-        )
 
     def check_last_tx_status(self) -> Tuple[bool, str]:
         """Check if the tx in the last round was successful or not"""
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -346,20 +346,37 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
         if len(submitted_ids) == 0:
             return
 
-        # Emit per-task delivery-time metrics from in-memory state.
-        # A task not found in shared_state[DONE_TASKS] still
-        # contributes to the prune below, but its metric is skipped;
-        # the delivery already landed on-chain, only the timing
-        # dimension is dropped.
+        # Take one snapshot under the lock and prune inline, so the
+        # metrics loop below and the prune both operate on the same
+        # entries. ``shared_state[DONE_TASKS]`` is written concurrently
+        # by the background executor, so an unlocked
+        # ``deepcopy(shared_state[DONE_TASKS])`` can produce a torn
+        # read: a task landing mid-copy would silently fall into the
+        # ``task is None`` skip branch below and its metric would be
+        # dropped, making a real desync indistinguishable from a race.
+        submitted_id_set = set(submitted_ids)
+        with self.done_tasks_lock():
+            shared_snapshot = list(self.context.shared_state.get(DONE_TASKS, []))
+            self.context.shared_state[DONE_TASKS] = [
+                task
+                for task in shared_snapshot
+                if str(task.get("request_id")) not in submitted_id_set
+            ]
+
+        # Emit per-task delivery-time metrics from the snapshot. A
+        # task absent from ``shared_state[DONE_TASKS]`` at snapshot
+        # time still counted toward the prune above; the metric is
+        # skipped only because the timing data isn't locally
+        # available (the delivery already landed on-chain).
         shared_done_tasks_by_id = {
-            str(task.get("request_id")): task for task in self.done_tasks
+            str(task.get("request_id")): task for task in shared_snapshot
         }
         for req_id in submitted_ids:
             task = shared_done_tasks_by_id.get(req_id)
             if task is None:
                 self.context.logger.debug(
                     "tool_delivery_time skipped for request_id=%s: "
-                    "not in shared_state[DONE_TASKS]",
+                    "not in shared_state[DONE_TASKS] snapshot",
                     req_id,
                 )
                 continue
@@ -399,7 +416,6 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
             f"The corresponding tx_hash is: {tx_hash}. "
             f"Removing them from the list of tasks to be processed."
         )
-        self.remove_tasks_by_id(submitted_ids)
 
     def check_last_tx_status(self) -> Tuple[bool, str]:
         """Check if the tx in the last round was successful or not"""
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -27,7 +27,18 @@ from collections import defaultdict
 from copy import deepcopy
 from datetime import datetime, timezone
 from enum import Enum
-from typing import Any, Dict, Generator, List, Optional, Set, Tuple, Type, cast
+from typing import (
+    Any,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Type,
+    cast,
+)
 
 from aea.helpers.cid import CID, to_v1
 from prometheus_client import Counter, Gauge, Histogram
@@ -75,6 +86,7 @@ from packages.valory.skills.task_submission_abci.rounds import (
     TaskPoolingRound,
     TaskSubmissionAbciApp,
     TransactionPreparationRound,
+    extract_request_ids,
 )
 from packages.valory.skills.transaction_settlement_abci.payload_tools import (
     hash_payload_to_hex,
@@ -228,30 +240,19 @@ class TaskExecutionBaseBehaviour(BaseBehaviour, ABC):
 
         :param submitted_tasks: the done tasks that have already been submitted
         """
-        # Extract ids and delegate; both entry points share
-        # ``remove_tasks_by_id`` so match semantics stay consistent.
-        submitted_ids = [
-            str(task["request_id"])
-            for task in submitted_tasks
-            if task.get("request_id") is not None and task.get("request_id") != ""
-        ]
-        self.remove_tasks_by_id(submitted_ids)
+        self.remove_tasks_by_id(extract_request_ids(submitted_tasks))
 
-    def remove_tasks_by_id(self, submitted_ids: List[str]) -> None:
+    def remove_tasks_by_id(self, submitted_ids: Sequence[Any]) -> None:
         """Pop already-submitted tasks from shared state, matched by request_id.
 
         :param submitted_ids: request ids of tasks already delivered.
+            Accepts ``int`` or ``str`` shapes; both sides are
+            ``str``-normalised on the equality check.
         """
-        # run this in a lock
-        # the amount of done tasks will always be relatively low (<<20)
-        # we can afford to do this in a lock
+        # Amount of done tasks is small (<<20); the lock cost is trivial.
         submitted_id_set = {str(rid) for rid in submitted_ids}
         with self.done_tasks_lock():
             done_tasks = self.done_tasks
-            # ``str``-normalise the match key. The shared-state side
-            # may hold ``int`` or ``str`` ``request_id`` (mixed types
-            # can survive an in-place restart transition); normalising
-            # both sides here keeps the prune consistent regardless.
             not_submitted = [
                 done_task
                 for done_task in done_tasks
@@ -330,12 +331,11 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
     def handle_submitted_tasks(self) -> Generator[None, None, None]:
         """Handle tasks that have been already submitted before (in a prev. period).
 
-        Reads ``submitted_request_ids`` for the id list, or falls back
-        to ``done_tasks`` when the id field isn't yet in the DB (a
-        mech that resumes with pre-existing state written before this
-        code shipped will see ``None`` on the first read). Per-task
-        metrics are read from ``shared_state[DONE_TASKS]`` since the
-        id-only hand-off doesn't carry tool / start_time.
+        Reads ``submitted_request_ids`` for the id list and prunes
+        ``shared_state[DONE_TASKS]`` accordingly. Per-task
+        ``tool_delivery_time`` metrics are read from
+        ``shared_state[DONE_TASKS]`` since the id-only hand-off
+        doesn't carry tool / start_time.
 
         :yield: AEA protocol messages while fetching the tx block number.
         """
@@ -344,31 +344,14 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
         if not status:
             return
 
-        # Fallback path for state that predates
-        # ``submitted_request_ids`` being written. Applies for at most
-        # one cycle per mech; subsequent cycles read the id list
-        # written by the previous ``PostTxSettlementRound``.
-        submitted_ids_raw = self.synchronized_data.db.get("submitted_request_ids", None)
-        if submitted_ids_raw is None:
-            self.context.logger.info(
-                "submitted_request_ids missing; falling back to done_tasks."
-            )
-            submitted_tasks_fallback = cast(
-                List[Dict[str, Any]], self.synchronized_data.done_tasks
-            )
-            submitted_ids = [
-                str(task["request_id"])
-                for task in submitted_tasks_fallback
-                if task.get("request_id") is not None and task.get("request_id") != ""
-            ]
-        else:
-            submitted_ids = [str(rid) for rid in cast(List[Any], submitted_ids_raw)]
-
+        submitted_ids = list(
+            cast(SynchronizedData, self.synchronized_data).submitted_request_ids
+        )
         if len(submitted_ids) == 0:
             return
 
         # Emit per-task delivery-time metrics from in-memory state.
-        # A task not found in ``shared_state[DONE_TASKS]`` still
+        # A task not found in shared_state[DONE_TASKS] still
         # contributes to the prune below, but its metric is skipped;
         # the delivery already landed on-chain, only the timing
         # dimension is dropped.
@@ -378,10 +361,22 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
         for req_id in submitted_ids:
             task = shared_done_tasks_by_id.get(req_id)
             if task is None:
+                self.context.logger.debug(
+                    "tool_delivery_time skipped for request_id=%s: "
+                    "not in shared_state[DONE_TASKS]",
+                    req_id,
+                )
                 continue
             tool = task.get("tool")
             start_time = task.get("start_time")
             if tool is None or start_time is None:
+                self.context.logger.debug(
+                    "tool_delivery_time skipped for request_id=%s: "
+                    "missing field(s) tool=%r start_time=%r",
+                    req_id,
+                    tool,
+                    start_time,
+                )
                 continue
             tool_delivery_time_duration = time.perf_counter() - float(start_time)
             self.context.logger.info(
--- a/packages/valory/skills/task_submission_abci/rounds.py
+++ b/packages/valory/skills/task_submission_abci/rounds.py
@@ -20,7 +20,6 @@
 """This package contains the rounds of TaskSubmissionAbciApp."""
 
 import json
-import logging
 from enum import Enum
 from typing import Any, Dict, FrozenSet, List, Optional, Set, Tuple, cast
 
@@ -118,7 +117,7 @@ class SynchronizedData(BaseSynchronizedData):
         if not isinstance(value, list) or not all(
             isinstance(item, str) for item in value
         ):
-            logging.getLogger(__name__).error(
+            self.db.logger.error(
                 "submitted_request_ids invariant broken: expected list[str], "
                 "got %s=%r; degrading to [] for this cycle",
                 type(value).__name__,
@@ -338,20 +337,21 @@ class PostTxSettlementRound(CollectSameUntilThresholdRound):
     def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
         """Process the end of the block."""
         if self.threshold_reached:
-            done_tasks = cast(SynchronizedData, self.synchronized_data).done_tasks
+            sd = cast(SynchronizedData, self.synchronized_data)
+            done_tasks = sd.done_tasks
             if not done_tasks:
-                # ``PostTxSettlementRound`` is also reached from paths
-                # that carry no task batch: the delivery-rate settlement
-                # (``composition.py`` wires ``FinishedTransactionSubmissionRound``
-                # here for every settlement, not just task delivery) and
-                # a settlement-internal ``ResetRound`` retry (which
-                # drops ``done_tasks`` from the DB via ``db.create``'s
-                # allow-list). Writing ``submitted_request_ids = []``
-                # in those cases would clobber a still-pending hand-off
-                # from a prior real settlement, and the delivered batch
-                # would be re-pooled and re-submitted on-chain next
-                # cycle. When we have no task batch to record, leave
-                # any pending hand-off in place.
+                # ``composition.py`` wires
+                # ``FinishedTransactionSubmissionRound`` here for every
+                # settlement, not just task delivery, so the
+                # delivery-rate path arrives with ``done_tasks == []``.
+                # Overwriting a still-pending hand-off with ``[]`` here
+                # would clobber a prior real settlement's ids and the
+                # delivered batch would be re-pooled next cycle.
+                sd.db.logger.warning(
+                    "PostTxSettlementRound reached with empty done_tasks; "
+                    "preserving pending hand-off submitted_request_ids=%s",
+                    sd.submitted_request_ids,
+                )
                 return self.synchronized_data, Event.DONE
             submitted_ids = extract_request_ids(done_tasks)
             return (
--- a/packages/valory/skills/task_submission_abci/rounds.py
+++ b/packages/valory/skills/task_submission_abci/rounds.py
@@ -99,9 +99,24 @@ class SynchronizedData(BaseSynchronizedData):
         across periods because it holds per-event request/response
         payload data that would inflate DB serialization.
 
+        The value is validated on read: writers must go through
+        :func:`extract_request_ids` so that only ``str`` ids land in
+        the DB. A future writer that bypasses the helper and stores a
+        non-list or non-``str`` entries surfaces as ``TypeError`` here
+        rather than as silent type drift at the consumer.
+
         :return: the list of request ids from the most recent settlement.
+        :raises TypeError: if the persisted value is not ``list[str]``.
         """
-        return cast(List[str], self.db.get("submitted_request_ids", []))
+        value = self.db.get("submitted_request_ids", [])
+        if not isinstance(value, list) or not all(
+            isinstance(item, str) for item in value
+        ):
+            raise TypeError(
+                "submitted_request_ids must be a list[str]; got "
+                f"{type(value).__name__}={value!r}"
+            )
+        return value
 
     @property
     def final_tx_hash(self) -> str:
--- a/packages/valory/skills/task_submission_abci/rounds.py
+++ b/packages/valory/skills/task_submission_abci/rounds.py
@@ -70,6 +70,22 @@ class SynchronizedData(BaseSynchronizedData):
         return cast(List[Dict[str, Any]], self.db.get("done_tasks", []))
 
     @property
+    def submitted_request_ids(self) -> List[str]:
+        """Return request ids of tasks submitted in the most recent settlement.
+
+        Written by :class:`PostTxSettlementRound` end_block and read by
+        :meth:`TaskPoolingBehaviour.handle_submitted_tasks` next period
+        to prune ``shared_state[DONE_TASKS]`` of already-delivered
+        entries. Cross-period-persisted because it's small and the next
+        cycle needs it; the full ``done_tasks`` list is not carried
+        across periods because it holds per-event request/response
+        payload data that would inflate DB serialization.
+
+        :return: the list of request ids from the most recent settlement.
+        """
+        return cast(List[str], self.db.get("submitted_request_ids", []))
+
+    @property
     def final_tx_hash(self) -> str:
         """Get the verified tx hash."""
         return cast(str, self.db.get_strict("final_tx_hash"))
@@ -264,18 +280,18 @@ class PostTxSettlementRound(CollectSameUntilThresholdRound):
     NO_MAJORITY arm only fires if the agents can't agree on having reached
     this round at all, which is the same shape as every consensus round.
 
-    ``done_tasks`` MUST survive this round untouched. It is in
-    ``cross_period_persisted_keys`` because
-    ``TaskPoolingBehaviour.handle_submitted_tasks`` reads it at the START
-    of the next cycle to know which ``request_id``s to prune from
-    ``shared_state[DONE_TASKS]``. Clearing here on DONE strands the
-    delivered tasks in shared state, so every subsequent cycle re-pools
-    and re-delivers the same batch — the Safe tx succeeds each time but
-    the inner ``mech.deliverMulti`` reverts on the already-delivered
-    ids, burning gas without emitting new Deliver events. Any re-fire
-    concern on the NO_MAJORITY / ROUND_TIMEOUT self-loop must be handled
-    inside the behaviour (idempotence flag) rather than by mutating a
-    field the next behaviour depends on.
+    On DONE the round writes ``submitted_request_ids`` — the id-only
+    hand-off used by the next cycle's
+    :meth:`TaskPoolingBehaviour.handle_submitted_tasks` to prune
+    ``shared_state[DONE_TASKS]``. ``done_tasks`` itself stays in the
+    period it was set and is not carried across; the ID hand-off is
+    what rides consensus between cycles.
+
+    Do NOT mutate ``done_tasks`` here. It is still read by predict-api
+    write and log emission earlier in this same behaviour, and the
+    behaviour-side prune in the next cycle depends on
+    ``shared_state[DONE_TASKS]`` reflecting only tasks that were not
+    settled — mutating the consensus field here breaks that contract.
     """
 
     payload_class = PostTxSettlementPayload
@@ -286,7 +302,30 @@ class PostTxSettlementRound(CollectSameUntilThresholdRound):
     def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
         """Process the end of the block."""
         if self.threshold_reached:
-            return self.synchronized_data, Event.DONE
+            # Extract the id list from this period's done_tasks and
+            # carry it via ``submitted_request_ids`` for the next
+            # cycle's prune. ``str``-normalise for parity with the
+            # ``str``-keyed dedup rule in
+            # ``TaskPoolingRound.end_block``; downstream match sites
+            # normalise the same way.
+            done_tasks = cast(
+                List[Dict[str, Any]],
+                cast(SynchronizedData, self.synchronized_data).done_tasks,
+            )
+            submitted_ids = [
+                str(task["request_id"])
+                for task in done_tasks
+                if task.get("request_id") is not None and task.get("request_id") != ""
+            ]
+            return (
+                self.synchronized_data.update(
+                    synchronized_data_class=SynchronizedData,
+                    **{
+                        get_name(SynchronizedData.submitted_request_ids): submitted_ids,
+                    },
+                ),
+                Event.DONE,
+            )
         if not self.is_majority_possible(
             self.collection, self.synchronized_data.nb_participants
         ):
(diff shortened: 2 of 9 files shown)


packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -286,7 +336,17 @@ class PostTxSettlementRound(CollectSameUntilThresholdRound):
     def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
         """Process the end of the block."""
         if self.threshold_reached:
-            return self.synchronized_data, Event.DONE
+            done_tasks = cast(SynchronizedData, self.synchronized_data).done_tasks
+            submitted_ids = extract_request_ids(done_tasks)
```
OjusWiZard, line 356: **Critical — this write is unconditional, so it clobbers a hand-off that hasn't been consumed yet.**

`done_tasks` is now period-scoped (`AbciAppDB.get` only reads `_data[reset_index]`, and it's no longer in `cross_period_persisted_keys`). So any entry into `PostTxSettlementRound` in a period where `TaskPoolingRound` never ran resolves `done_tasks` to the `[]` default and then *overwrites* a still-pending `submitted_request_ids`.

Two reachable routes:

**(a) Delivery-rate settlement.** `composition.py:52` maps `FinishedTransactionSubmissionRound -> PostTxSettlementRound`, and that's the *only* mapping for that final state — so the delivery-rate tx path (`composition.py:43` `FinishedWithTxRound -> RandomnessTransactionSubmissionRound`) lands here too. Sequence:

1. Period N settles tasks A, B -> `submitted_request_ids = ["A","B"]`. `shared_state[DONE_TASKS]` still holds A, B (prune is deferred to next cycle by design).
2. `ResetAndPauseRound` times out / `NO_MAJORITY` -> `FinishedResetAndPauseErrorRound` -> `RegistrationRound` -> `UpdateDeliveryRateRound`. New period; `["A","B"]` correctly carried.
3. Delivery-rate tx settles -> `PostTxSettlementRound` -> `done_tasks == []` -> `submitted_request_ids = []`.
4. Next `handle_submitted_tasks` returns early on the empty list. A and B are never pruned, get re-pooled, and are re-submitted on-chain.

The process never restarted, so `shared_state[DONE_TASKS]` is still populated across all of this.

**(b) The settlement-internal `ResetRound`.** `transaction_settlement_abci/rounds.py:606` calls `db.create()` with a hard-coded allow-list; `AbciAppDB.create` keeps that list *plus* `cross_period_persisted_keys`. Verified empirically against this branch:

```
POST-PR (submitted_request_ids): done_tasks after settlement ResetRound = MISSING
PRE-PR  (done_tasks):            done_tasks after settlement ResetRound = [{'request_id': 7, ...}]
```

`ResetRound` is entered on `NO_MAJORITY` from four settlement rounds and its DONE arm retries via `RandomnessTransactionSubmissionRound`, which can still reach `PostTxSettlementRound`. Same clobber — plus `PostTxSettlementBehaviour` then reads `done_tasks == []` and silently drops the whole cycle's predict-api batch (`behaviours.py:2005-2011` logs it as "a truly quiet round", indistinguishable from a wiped DB).

Caveat on (b): `most_voted_tx_hash` isn't preserved across that `create()` either, so the retry may already fail earlier for unrelated reasons. But the invariant this PR now relies on — "`done_tasks` set by `TaskPoolingRound` is still visible at `PostTxSettlementRound`" — is not guaranteed by the FSM, and the `db_pre_conditions` comment below asserts it as fact.

This is the exact failure the docstring you replaced warned about: "burning gas without emitting new Deliver events."

Minimal fix: don't write when `submitted_ids` is empty — return `self.synchronized_data, Event.DONE` unchanged. See the note on the prune call in `behaviours.py` though; these two want fixing together.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -323,43 +329,81 @@ def get_done_tasks(self, timeout: float) -> Generator[None, None, List[Dict]]:
...
+            self.set_gauge(
+                self.shared_state.mech_delivery_last_block_number, block_number
+            )
+
+        self.context.logger.info(
+            f"Tasks with ids {submitted_ids} have been submitted. "
+            f"The corresponding tx_hash is: {tx_hash}. "
+            f"Removing them from the list of tasks to be processed."
+        )
+        self.remove_tasks_by_id(submitted_ids)
```
OjusWiZard: **Critical — the hand-off is never cleared, so this prune runs against a stale id list forever.**

`submitted_request_ids` is written only by `PostTxSettlementRound.end_block` and nothing resets it. `final_tx_hash` is likewise cross-period-persisted and never reset, so `check_last_tx_status()` returns `True` indefinitely after the first settlement. Net effect: the whole "already submitted" path — receipt fetch, gauge set, the INFO at line 401 asserting tasks "have been submitted", and this prune — executes in *every* period from then on, including periods where nothing was settled.

Pre-PR this was bounded to one cycle, because `TaskPoolingRound.end_block` rewrote `done_tasks` on **every** branch, including to `[]` on `NO_TASKS` (`rounds.py:276`). The staleness is now unbounded.

Two consequences:

**Silent permanent task loss.** `_handle_get_undelivered_reqs` (`task_execution/handlers.py:703`) re-enqueues requests whose delivery never landed on-chain. Those re-execute and re-appear in `shared_state[DONE_TASKS]` under the **same** `request_id` — where the stale id list prunes them again before they can ever be pooled. Execute -> pruned -> re-swept -> execute, indefinitely; the requester is never served, compute is burned each round, and the only log is the INFO above falsely claiming they were submitted.

**Stale metric that looks fresh.** Lines 392-399 re-fetch the receipt for the same old `tx_hash` every period and call `set_gauge`, which does `metric.set_to_current_time()` before setting the value. `mech_delivery_last_block_number` therefore reports a freshly-timestamped, hours-old block number on a mech that has delivered nothing — a "deliveries stalled" alert can never fire. Plus one wasted RPC round-trip per period.

Note this interacts with the `end_block` finding: guarding that write on a non-empty list makes the staleness here *more* persistent. They need one combined fix — consume the hand-off after a successful prune (write `submitted_request_ids: []`), or key it to the `final_tx_hash` it was derived from and skip when that hash hasn't changed since the last prune.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -228,30 +240,24 @@ def remove_tasks(self, submitted_tasks: List[Dict[str, Any]]) -> None:
 
         :param submitted_tasks: the done tasks that have already been submitted
         """
-        # run this in a lock
-        # the amount of done tasks will always be relatively low (<<20)
-        # we can afford to do this in a lock
+        self.remove_tasks_by_id(extract_request_ids(submitted_tasks))
+
+    def remove_tasks_by_id(self, submitted_ids: Sequence[Any]) -> None:
```
OjusWiZard: `Sequence[Any]` is satisfied by a bare `str` (a `str` *is* a `Sequence[str]`), so this type-checks under CI mypy:

```python
self.remove_tasks_by_id(some_request_id)   # some_request_id: str = "1234"
```

and produces `submitted_id_set == {"1","2","3","4"}`. The intended task is not pruned, survives into the next `TaskPoolingRound`, and is re-submitted for delivery. I confirmed both halves — it runs, and `mypy` reports `Success: no issues found`. `bytes` has the same hole, and `Sequence[str]` / `Iterable[str]` / `Collection[str]` do **not** close it.

Beyond the hole, `Any` erases the invariant this PR just paid to establish: the chain is `extract_request_ids -> List[str]` into `Sequence[Any]`, the widest possible type at exactly the boundary where the narrowed type arrives. The tell is line 253 — both call sites already guarantee `str`, so that `str(rid)` is provably a no-op today and exists only because the signature doesn't say so.

`List[str]` closes the hole and lets line 253 become `set(submitted_ids)`. (The `str()` on line 259 is a different matter and genuinely needed — `context.shared_state` is untyped and may hold `int`.)

If you want the invariant actually enforced rather than conventional, `RequestId = NewType("RequestId", str)` threaded through `extract_request_ids` / this parameter / the property would get you mypy enforcement at zero runtime cost — `task_submission_abci` is in mypy scope, and `NewType` JSON-serialises as `str` so the DB is unaffected.

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -69,6 +87,37 @@ def done_tasks(self) -> List[Dict[str, Any]]:
...
+        rather than as silent type drift at the consumer.
+
+        :return: the list of request ids from the most recent settlement.
+        :raises TypeError: if the persisted value is not ``list[str]``.
+        """
+        value = self.db.get("submitted_request_ids", [])
+        if not isinstance(value, list) or not all(
+            isinstance(item, str) for item in value
+        ):
+            raise TypeError(
```
OjusWiZard: Raising from the *read* side of a cross-period-persisted key is a rough failure mode for this particular key.

This property is read at the top of `TaskPoolingBehaviour.async_act` (via `handle_submitted_tasks`), i.e. the first round of every FSM cycle. The synchronized DB is byte-identical across participants, so a bad value makes every agent raise deterministically on the same block — the whole service crash-loops simultaneously rather than degrading. And because the key is cross-period-persisted, `db.create()` copies the bad value forward across period resets, while the only code that could overwrite it (`PostTxSettlementRound.end_block`) is downstream of `TaskPoolingRound`. That's a poison pill with no in-band recovery.

Two more things worth weighing:

- Five lines below, `check_last_tx_status` wraps a structurally identical `SynchronizedData` property read in a bare `except Exception` and returns `(False, "")`. That's an established local pattern for "property reads may throw, absorb it" — a future maintainer applying it here turns this guard into a silent no-prune, which is the bug class this PR fixes.
- The `-> List[str]` annotation enforces nothing on its own: `AbciAppDB.get` returns `Any`, and `abstract_round_abci` is under `ignore_errors = True` in `tox.ini:78`. The write side goes through `update(**kwargs: Any)` keyed by a `get_name` string. This runtime check is the entire enforcement.

Validating at the single write site in `end_block` inverts the tradeoff: you still have the un-normalised `done_tasks` in hand to log, and the bad value hasn't entered consensus or cross-period storage yet. If you'd rather keep a read-side check, `logger.error` + `return []` keeps the drift detectable without taking the fleet down.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -323,43 +329,81 @@ def get_done_tasks(self, timeout: float) -> Generator[None, None, List[Dict]]:
...
 
-                block_number = yield from self._fetch_tx_block_number(tx_hash)
-                self.context.logger.info(
-                    f"Block number for tx hash: {tx_hash} is {block_number}"
-                )
-                if block_number:
-                    self.set_gauge(
-                        self.shared_state.mech_delivery_last_block_number, block_number
-                    )
+        submitted_ids = cast(
```
OjusWiZard: Two things here.

**1. The PR description's compat story doesn't match this code.** The description says `handle_submitted_tasks` "falls back to `done_tasks` when `submitted_request_ids` isn't in the DB" and lists a test for that fallback path. Neither exists — this reads `submitted_request_ids` only, returns early on an empty list, and never touches `done_tasks`; `grep -rn fallback` in the skill turns up two unrelated hits.

The good news is the fallback isn't needed: `ContractHandler.setup` resets `shared_state[DONE_TASKS] = []` on every process start (`task_execution/handlers.py:608`), and a package-hash bump forces a restart, so an upgraded agent boots with nothing to re-deliver. So this is a description fix, not a code fix — but the "Compat" and "Tests" sections should be corrected before merge rather than claiming a mechanism that isn't there.

**2. Minor:** the `cast` is redundant. `TaskExecutionBaseBehaviour.synchronized_data` is already declared `-> SynchronizedData` (line 206-208) and `TaskPoolingBehaviour` inherits it, so this collapses to:

```python
submitted_ids = self.synchronized_data.submitted_request_ids
```

(The sibling cast at `rounds.py:339` *is* needed — `AbstractRound.synchronized_data` is typed `BaseSynchronizedData`.)

packages/valory/skills/task_submission_abci/tests/test_rounds.py
```diff
@@ -138,6 +138,17 @@ def test_final_tx_hash_missing_raises(self) -> None:
         with pytest.raises(ValueError):
             _ = sd.final_tx_hash
 
+    def test_submitted_request_ids_defaults_to_empty_list(self) -> None:
+        """``submitted_request_ids`` reads ``[]`` when the key is unset."""
+        sd = _make_sync_data()
+        assert sd.submitted_request_ids == []
+
+    def test_submitted_request_ids_returns_stored_value(self) -> None:
```
OjusWiZard, line 147: Coverage on the new code is otherwise strong, but two gaps line up with the risk in this PR.

**The `TypeError` branch is the single uncovered statement in the module** — `rounds.py` reports 99%, missing line 115. These two tests only exercise the happy path. It's trivially reachable:

```
SynchronizedData(AbciAppDB({... "submitted_request_ids": [[1, 2]]})).submitted_request_ids
-> TypeError: submitted_request_ids must be a list[str]; got list=[1, 2]
```

Worth a two-case parametrised test (non-list value; list with non-`str` entries). That branch is the PR's stated defence against silent type drift — unexercised, a refactor could delete it with nothing failing.

**Nothing joins the two halves of the new contract.** The round tests run against a real `SynchronizedData`; the behaviour tests replace `synchronized_data` wholesale with a `MagicMock`. The only thing linking write to read is the attribute name appearing in both files. A test that takes the `new_sync_data` returned by `_make_post_tx_round(...).end_block()`, assigns it to `b._synchronized_data`, seeds `shared_state[DONE_TASKS]`, and asserts the right rows are pruned would pin the whole hand-off — which is the actual bug class this PR is fixing.

Neither of the two correctness findings I left on `rounds.py:340` and `behaviours.py:406` is covered either; both are cheap to pin at the round level.

OjusWiZard (commented):
Full multi-perspective review of #482 — correctness, tests, silent failures, types, comments, and simplification. Inline notes below.

Baseline is green: 874 tests pass, `black` / `isort` / `flake8` / `darglint` clean, `autonomy packages lock --check` verified.

The core idea is sound and the happy path works — `submitted_request_ids` does survive `ResetAndPauseRound` into the next `TaskPoolingRound`, and the id-only hand-off genuinely bounds `db.serialize()`. The two blocking problems are in the hand-off's **lifecycle**, not its shape:

1. **`PostTxSettlementRound.end_block` writes unconditionally** (`rounds.py:340`), so entering that round in a period without `done_tasks` — via the delivery-rate settlement path, or across the settlement-internal `ResetRound` — overwrites an unconsumed hand-off with `[]`. Delivered tasks then stay in `shared_state[DONE_TASKS]`, get re-pooled, and are re-delivered on-chain.
2. **The hand-off is never cleared** (`behaviours.py:406`), so the prune runs on a stale id list every period forever. Pre-PR, `done_tasks` was rewritten by `TaskPoolingRound.end_block` on every branch, so the list self-cleared each cycle.

These interact — guarding (1) on a non-empty list makes (2) more persistent — so they want one combined fix rather than two independent ones.

Also worth doing before merge: the **Compat / Tests sections of the description** claim a `done_tasks` fallback in `handle_submitted_tasks` and a test for it; neither exists. The fallback isn't actually needed (`shared_state[DONE_TASKS]` is reset on process start), so it's a description fix.

The rest is typing (`Sequence[Any]` admits a bare `str` and character-splits it), the placement of the `TypeError` guard, two test gaps, and comment volume/accuracy against CLAUDE.md's bar.

Two things I checked and want to explicitly *not* raise:
- Adding `submitted_request_ids` to `cross_period_persisted_keys` looks like it could raise `ABCIAppInternalError` at `ResetAndPause` on paths that skip `PostTxSettlementRound` — but `final_tx_hash` has identical writer reachability and already behaves the same way, so it's pre-existing, not a regression here.
- The "replay buffer" and `use_offchain` inaccuracies in the `PostTxSettlementRound` docstring are pre-existing (commit `06cab97d`), not from this PR.

packages/valory/skills/task_submission_abci/tests/test_behaviours.py
```diff
@@ -659,9 +714,15 @@ def _make_b(self) -> "_DummyPooling":
...
 
-    def _patch_sd(self, b: Any, done_tasks: Any) -> Any:
+    def _patch_sd(self, b: Any, submitted_ids: List[str]) -> Any:
+        """Wire the mocked synchronized_data with a fixed id list.
+
+        :param b: the behaviour under test to attach the mock onto.
+        :param submitted_ids: value the property returns.
+        :return: a no-op context manager so the caller's ``with`` block stays uniform.
+        """
         mock_sd = MagicMock()
```
OjusWiZard, line 720: Because `cast()` is a runtime no-op, setting `submitted_request_ids` on a bare `MagicMock` means the real property — including its `list[str]` validation — never executes on the behaviour path. These tests would keep passing if the property were deleted, renamed on the round side only, or if its return type changed.

Using a real `SynchronizedData` here costs one helper call and closes this plus the missing round-trip coverage.

Two smaller ones in this class:

- `test_status_true_with_tasks_no_block_number` and its sibling below don't patch `remove_tasks_by_id`, so the real prune actually runs and `ctx.shared_state[DONE_TASKS]` ends as `[]` — but neither asserts it. Adding `assert ctx.shared_state[DONE_TASKS] == []` turns the strongest integration point in the file from incidental into intentional.
- `test_status_false_removes_nothing` asserts nothing about removal despite its name. `patch.object(b, "remove_tasks_by_id")` + `assert_not_called()` would match it.

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -264,18 +313,19 @@ class PostTxSettlementRound(CollectSameUntilThresholdRound):
...
-    inside the behaviour (idempotence flag) rather than by mutating a
-    field the next behaviour depends on.
+    On DONE the round writes ``submitted_request_ids`` — the id-only
+    hand-off used by the next cycle's
+    :meth:`TaskPoolingBehaviour.handle_submitted_tasks` to prune
+    ``shared_state[DONE_TASKS]``. ``done_tasks`` itself stays in the
+    period it was set and is not carried across; the ID hand-off is
+    what rides consensus between cycles.
+
+    Do NOT mutate ``done_tasks`` here. It is still read by
```
OjusWiZard: This paragraph is wrong on both halves, which matters because it's the warning the next reader will act on.

> It is still read by `PostTxSettlementBehaviour` **earlier in the same period**

If the behaviour already read it earlier, a mutation in `end_block` can't affect it. The stated ordering disproves the stated hazard.

> the behaviour-side prune in the next cycle depends on `shared_state[DONE_TASKS]` reflecting only tasks that were not settled — mutating the consensus field here breaks that contract

`shared_state[DONE_TASKS]` is agent-local process memory (`behaviours.py:225`). Mutating the *consensus* `done_tasks` field has no effect on it at all; the two are connected only through `submitted_request_ids`.

The real hazard is one line below the paragraph: `end_block` itself calls `extract_request_ids(done_tasks)`, so clearing `done_tasks` first would yield an empty id list and silently disable the next cycle's prune. That's the coupling worth a line — and it's the only part of this docstring I'd keep.

The rest of what this PR adds to the class docstring (the "On DONE the round writes..." paragraph) restates the four lines of `end_block` directly below it.

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -376,16 +436,21 @@ class TaskSubmissionAbciApp(AbciApp[Event]):
...
             get_name(SynchronizedData.final_tx_hash),
         ]
     )
     db_pre_conditions: Dict[AppState, Set[str]] = {
         TaskPoolingRound: set(),
-        # Entered from composition after settlement; relies on done_tasks
-        # being on synchronized_data from the prior TaskPoolingRound cycle
-        # (cross_period_persisted_keys above keeps it alive). No additional
-        # pre-condition fields beyond what the FSM already carries.
+        # Entered from composition after settlement. Reads
```
OjusWiZard: This comment asserts as fact the invariant that the `end_block` finding above disproves: `done_tasks` being "present in the current period's DB slot" is **not** guaranteed — the settlement-internal `ResetRound` (`transaction_settlement_abci/rounds.py:606`) opens a new period mid-settlement and drops it, and the delivery-rate path reaches `PostTxSettlementRound` in a period where `TaskPoolingRound` never ran at all.

So this is worse than redundant — it will actively mislead the next reader into assuming a guarantee the FSM doesn't provide.

Also, it describes what the round *does* rather than why the pre-condition set is empty, which is the only thing a reader stopped at `PostTxSettlementRound: set()` would be wondering.

Same for the `cross_period_persisted_keys` comment a few lines up: it duplicates the property docstring at lines 97-100, and the decision it records is what the commit message and PR body are for.

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -41,6 +41,24 @@
 )
 
 
+def extract_request_ids(tasks: List[Dict[str, Any]]) -> List[str]:
```
OjusWiZard, line 44: Three notes on this helper.

**The predicate now exists three times in this file** and they have to stay in lockstep:
- here, line 58 — filters, no log
- `TaskPoolingRound.end_block`, line 188 — skips, logs a `warning`, with a comment calling it "a real protocol violation worth surfacing"
- the sort key, line 220 — `str(x.get("request_id", ""))`, a third spelling of the same normalisation

They agree today. If they ever drift, a task kept by the `end_block` dedup but dropped here settles on-chain, never enters `submitted_request_ids`, is never pruned from `shared_state[DONE_TASKS]`, and gets re-collected and re-delivered — i.e. the exact bug this PR fixes, reintroduced by a one-line edit in either copy. One shared predicate used by all three sites would close that.

**The silent drop is asymmetric with its twin.** Line 188 warns; this one doesn't. At the `end_block` call site a dropped id becomes a permanently stuck row in `shared_state[DONE_TASKS]`, re-serialised into the pooling payload every round and never delivered, with nothing counting or alerting.

**Subscript/`get` mix.** The filter uses `task.get("request_id")` twice and the projection uses `task["request_id"]`. Correct today for plain dicts only because comprehension evaluation order runs the `if` first — but a plausible tidy-up of the guard to `task["request_id"] is not None` introduces a `KeyError` inside a consensus `end_block`. Binding once (`raw = task.get("request_id")` in a plain loop) removes the trap and the docstring paragraph justifying `is None or == ""` over `not`.

Small thing in that docstring: "Falsy ids (missing key, `None`, empty string) are dropped" and "Legit `0` (int) is kept" contradict each other — `0` is falsy. (`False` is also kept, and stringifies to `"False"`.)

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -323,43 +329,81 @@ def get_done_tasks(self, timeout: float) -> Generator[None, None, List[Dict]]:
...
+        ).submitted_request_ids
+        if len(submitted_ids) == 0:
+            return
 
+        # Emit per-task delivery-time metrics from in-memory state.
+        # A task not found in shared_state[DONE_TASKS] still
+        # contributes to the prune below, but its metric is skipped;
+        # the delivery already landed on-chain, only the timing
+        # dimension is dropped.
+        shared_done_tasks_by_id = {
```
OjusWiZard, line 370: A few things about the metrics block.

**Built outside the lock.** `self.done_tasks` reads `shared_state[DONE_TASKS]` unlocked, while the background executor appends under `done_tasks_lock` (`task_execution/behaviours.py:1811-1812`). Impact is metrics-only, but a torn read drops a task into the silent `debug` branch below, making a real desync and a benign race look identical. Reuse the snapshot `remove_tasks_by_id` already takes under the lock.

**`debug` is too quiet for those skip branches** — but they're also conflating two different things. `task is None` fires legitimately for tasks executed by *other* agents (pooling merges every participant's `done_tasks`), and it also fires on a genuine consensus-DB-vs-shared-state desync. Missing `tool`/`start_time` is a third case: `shared_state[DONE_TASKS]` has exactly one append site and it always sets both, so that one means a malformed entry. A labelled counter (`reason="not_in_shared_state"` vs `reason="malformed"`) would separate them; as-is, if the whole id list misses, `tool_delivery_time` drops to zero samples with nothing anywhere explaining why while line 401 still logs "have been submitted".

**Sample count changes ~1/N.** Metrics previously came from the consensus `done_tasks` (every agent emitted one observation per task in the batch); they now come from local `shared_state[DONE_TASKS]`. This is arguably a *correctness fix* — `start_time` is `time.perf_counter()`, which is process-relative, so subtracting another agent's `start_time` from the local counter produced meaningless durations. Flagging only so it's a conscious change: any dashboard or alert threshold keyed on this histogram's rate will shift on deploy. Worth a line in the PR body.

**Shape.** This loop needs no `yield`, so pulling it into a plain `_observe_delivery_times(submitted_ids)` makes the generator a linear four-step read. Using `tasks_by_id.get(req_id, {})` then collapses the two skip branches into the single `tool is None or start_time is None` guard, and the four-line comment above stops being needed because there's no non-obvious two-branch shape left to explain.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -228,30 +240,24 @@ def remove_tasks(self, submitted_tasks: List[Dict[str, Any]]) -> None:
...
+        self.remove_tasks_by_id(extract_request_ids(submitted_tasks))
+
+    def remove_tasks_by_id(self, submitted_ids: Sequence[Any]) -> None:
+        """Pop already-submitted tasks from shared state, matched by request_id.
+
+        :param submitted_ids: request ids of tasks already delivered.
+            Accepts ``int`` or ``str`` shapes; both sides are
+            ``str``-normalised on the equality check.
+        """
+        # Amount of done tasks is small (<<20); the lock cost is trivial.
```
OjusWiZard: This reads as justifying why the lock is *affordable*, which implies it's discretionary. It isn't — `shared_state[DONE_TASKS]` is written concurrently by the background `task_execution` thread, so the lock is required for correctness regardless of cost. `<<20` is also an unsourced bound that nothing enforces, and the comment sits on the `submitted_id_set` line, which is outside the `with`.

Per CLAUDE.md this one can just go. Same for the two-line normalisation paragraph in the docstring above — `str(rid)` on the next line says it, and if the signature tightens to `List[str]` (see my note on line 245) the paragraph describes a contract that no longer exists.

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -376,16 +436,21 @@ class TaskSubmissionAbciApp(AbciApp[Event]):
     }
     cross_period_persisted_keys: FrozenSet[str] = frozenset(
         [
-            get_name(SynchronizedData.done_tasks),
+            # Hand-off signal only. ``done_tasks`` is intentionally
+            # not cross-period-persisted because its per-entry payload
```
bennyjo: This removes the carried copy, but the current period's `done_tasks` is still serialized when `RegistrationRound` is entered on the reset-error path, so this halves the payload rather than bounding it. I reproduced with the real DB: one max-size event per batch now fits (525 KB), two still reject both the registration and the pooling payload. Could the body and this comment say that, and is keeping `predict_api_event` out of consensus a follow-up you'd track?

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -376,16 +436,21 @@ class TaskSubmissionAbciApp(AbciApp[Event]):
...
+            # data is large enough to inflate DB serialization on the
+            # next registration.
+            get_name(SynchronizedData.submitted_request_ids),
```
bennyjo, line 455: The tx-settlement skill's `ResetRound` (NO_MAJORITY from keeper selection or signature collection) calls `db.create()` mid-cycle, and only cross-period keys survive it. With `done_tasks` removed from this set, a settlement retry reaches `PostTxSettlementRound` with an empty list: no predict-api events for the batch, and `submitted_request_ids=[]` means the delivered batch is re-pooled next period. Reproduced with the real `AbciAppDB`. Keeping the payload out of consensus and leaving a small `done_tasks` here as a cross-period key would fix both this and the size issue.

packages/valory/skills/task_submission_abci/rounds.py
bennyjo: Raising here crash-loops every agent on the same consensus value, and a restart replays it. Since `extract_request_ids` is the only writer, would a warning plus `str` normalisation be enough? Happy to leave it if the loud failure is deliberate.

packages/valory/skills/task_submission_abci/tests/test_behaviours.py
```diff
@@ -686,30 +747,131 @@ def test_status_true_with_tasks_no_block_number(self) -> None:
...
             result = _run_gen(b.handle_submitted_tasks())
         assert result is None
+        # Histogram MUST fire for the task in shared_state; a regression
+        # that flips the ``if task is None`` guard would silently skip
+        # the emit without this assertion.
+        mock_hist.assert_called_once()
+        _, kwargs = mock_hist.call_args
+        assert kwargs["tool"] == "t1"
 
     def test_status_true_with_tasks_and_block_number(self) -> None:
```
bennyjo, line 767: Could this test also assert `shared_state[DONE_TASKS] == []` after the call? Nothing currently checks that the prune actually lands end to end.

bennyjo (commented):
Tests pass locally (874). Main ask: the tx-settlement ResetRound drops done_tasks mid-cycle, see inline. Also the 'bounded' claim needs softening, and the PR body still describes the fallback that b9dcc04 removed.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -323,43 +329,81 @@ def get_done_tasks(self, timeout: float) -> Generator[None, None, List[Dict]]:
...
-                    f"Block number for tx hash: {tx_hash} is {block_number}"
-                )
-                if block_number:
-                    self.set_gauge(
-                        self.shared_state.mech_delivery_last_block_number, block_number
-                    )
+        submitted_ids = cast(
+            SynchronizedData, self.synchronized_data
+        ).submitted_request_ids
+        if len(submitted_ids) == 0:
```
bennyjo, line 351: Since `submitted_request_ids` is never cleared, every idle period re-enters this branch for the last success: a receipt RPC, a "removing them" log line, and a no-op prune. Before the PR the carry went empty after an idle period because `TaskPoolingRound` always rewrote `done_tasks`; clearing the id list in `TaskPoolingRound.end_block` would restore that. Minor, happy to leave it if you'd rather not touch the round.

bennyjo (commented):
One minor follow-up from the prune path.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Fixed in dd421666. Guarded the write on non-empty done_tasks, so the delivery-rate path and the settlement-internal ResetRound retry no longer clobber a still-pending hand-off. Added test_no_write_when_done_tasks_empty which seeds a pending value and asserts it survives the empty-batch DONE.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in dd421666 (paired with the write-side guard above). TaskPoolingRound.end_block now clears submitted_request_ids on both DONE and NO_TASKS, so by the time PostTxSettlement runs again the previous cycle's hand-off has been consumed. This restores the one-shot semantics that TaskPoolingRound.end_block used to give done_tasks pre-PR. Coverage in test_submitted_request_ids_cleared_on_done and _cleared_on_no_tasks.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 97c2e84b. Tightened to List[str] and dropped the redundant str(rid) normalisation on the parameter side (the str() on the shared_state lookup stays because context.shared_state is untyped). Adjusted test_mixed_type_ids_match_normalised to test_int_shared_state_matches_str_id since the caller-side legacy-int case is now rejected by the type contract. NewType is a nice follow-up but I left it out of this round to keep the diff scoped.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Fixed in 97c2e84b. Downgraded to logger.error + return []. Added parametrised test_submitted_request_ids_degrades_to_empty_on_bad_shape covering non-list, non-str entries, and mixed cases, asserting both the empty return and the error log record.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Both fixed. PR body corrected in the same update round, the false 'fallback to done_tasks' claim is gone (the fallback never existed and isn't needed since ContractHandler.setup resets shared_state[DONE_TASKS] on process start). Redundant cast removed in a305fbf7.

packages/valory/skills/task_submission_abci/tests/test_rounds.py
DIvyaNautiyal07: Both gaps closed. TypeError-branch coverage in 97c2e84b (test_submitted_request_ids_degrades_to_empty_on_bad_shape). Round-trip integration test in 1b94cd73 (TestSubmittedRequestIdsRoundTrip) uses a real SynchronizedData so the property is actually exercised on the behaviour path, seeds shared_state[DONE_TASKS] with mixed present/absent ids, and asserts the prune reaches shared_state. The write-side guard and the TaskPooling clear are also directly tested (test_no_write_when_done_tasks_empty, test_submitted_request_ids_cleared_on_done, _on_no_tasks).

packages/valory/skills/task_submission_abci/tests/test_behaviours.py
DIvyaNautiyal07: Round-trip test in 1b94cd73 uses a real SynchronizedData so the property runs on the behaviour path. On the two smaller ones: test_status_true_with_tasks_no_block_number and _and_block_number now assert shared_state[DONE_TASKS] == [] end to end (0518c00e), and test_status_false_removes_nothing now seeds shared_state and asserts it's untouched instead of only checking the return value.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Fixed in a305fbf7. Dropped the 'Do NOT mutate done_tasks / earlier in the same period' paragraph, you were right on both halves. Kept the top of the docstring that describes what the round is for (predict-api post, always DONE on threshold).

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Fixed in a305fbf7. Removed the pre_conditions comment (it asserted an invariant the FSM does not guarantee, per your finding on line 340) and the cross_period_persisted_keys comment (duplicated the property docstring).

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Docstring contradiction fixed in a305fbf7: reworded to 'None or empty string are dropped; any other value including 0 and False is kept and stringified', which matches the code.

On the three-site duplication and the get/subscript mix: leaving those in this round. The three sites do slightly different actions (filter silent, filter + warn, sort-key with '' default), so a shared predicate would only cover the falsy check and the three action shapes would still diverge. The subscript in the projection is safe today because comprehension order runs the filter first, but I hear the concern about a future tidy-up introducing a KeyError. Happy to bind once (raw = task.get(...)) in a follow-up if it's a blocker.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Snapshot-under-the-lock is fixed in 0518c00e: the metrics loop and the prune now operate on the same list captured under done_tasks_lock, so a torn read can't drop a task into the None branch. Sample-count change noted in the PR body (correctness improvement since start_time is process-relative, but any dashboard threshold on this histogram will shift on deploy).

On the labelled counter to distinguish 'not_in_shared_state' from 'malformed' and the _observe_delivery_times extraction: leaving both for a follow-up. The counter is genuine metrics scope creep, and after the inline prune the loop is short and linear enough that extracting a helper only saves a nesting level. Happy to revisit if either is a blocker.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in 97c2e84b. Dropped both the discretionary-cost comment and the two-line normalisation paragraph. The tightened List[str] type and the set() call on the next line make the intent obvious without the prose.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Softened in the PR body. Now describes the reduction accurately: one max-size predict-api event per batch fits (about 525 KB), two still don't. Keeping predict_api_event out of consensus is tracked as the natural follow-up.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Fixed in dd421666. PostTxSettlementRound.end_block now guards the write on non-empty done_tasks, so the settlement-internal ResetRound retry (and the delivery-rate settlement path, same shape) no longer clobbers a pending submitted_request_ids with []. The delivered batch's ids are preserved for the next TaskPoolingRound to consume.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Downgraded to logger.error + return [] in 97c2e84b, exactly for the crash-loop reason you called out. Since extract_request_ids is the sole writer today and normalises to str, the degrade path is mostly a defense-in-depth signal for a future writer that bypasses the helper. Coverage in test_submitted_request_ids_degrades_to_empty_on_bad_shape.

packages/valory/skills/task_submission_abci/tests/test_behaviours.py
DIvyaNautiyal07: Done in 0518c00e. Both test_status_true_with_tasks_no_block_number and _and_block_number now assert shared_state[DONE_TASKS] == []. Also added test_reads_submitted_request_ids_when_present which seeds two tasks and asserts only the one whose id was in submitted_ids is pruned, so the mixed-selection case is pinned too.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Same fix. TaskPoolingRound.end_block clears submitted_request_ids on DONE and NO_TASKS (dd421666), so an idle period sees the id list already empty and the whole 'already submitted' block returns early on the len(submitted_ids) == 0 guard: no receipt RPC, no log, no no-op prune.

DIvyaNautiyal07:
Thanks for the deep reviews. Inline replies on each thread. Summary of what changed since the last commit (76e92ff3):

Two correctness fixes bundled as one shape in dd421666:
- PostTxSettlementRound.end_block guards the write on non-empty done_tasks, so the delivery-rate settlement and the settlement-internal ResetRound retry no longer clobber a pending hand-off with [].
- TaskPoolingRound.end_block clears submitted_request_ids on DONE and NO_TASKS, so the field is one-shot per settlement and the "already submitted" block in handle_submitted_tasks stops re-executing on stale ids every period.

Type hole (Sequence[Any] admits a bare str and character-splits it) tightened to List[str], and the property TypeError downgraded to logger.error + return [] so a byte-identical bad value can't crash-loop the whole fleet (97c2e84b).

Metrics loop and prune now share one snapshot taken under done_tasks_lock, so a torn read can't drop a task into the None branch (0518c00e).

Round-trip integration test in 1b94cd73 uses a real SynchronizedData (not a MagicMock), so the property is actually exercised on the behaviour path and the prune is verified end to end via shared_state[DONE_TASKS].

Docs pass in a305fbf7: dropped the "Do NOT mutate done_tasks / earlier in the same period" paragraph, the misleading db_pre_conditions comment, the redundant cast in handle_submitted_tasks, and fixed the extract_request_ids docstring contradiction.

PR body updated: removed the false "fallback to done_tasks" claim, softened "bounded" to describe the actual reduction (one max blob fits, two don't), and noted the histogram sample-count semantic change so any dashboard threshold keyed on it can be updated on deploy.

Two things intentionally deferred with rationale on the respective threads: the shared falsy-id predicate for extract_request_ids (three sites do slightly different actions) and the metrics helper extraction / labelled counter (CLAUDE.md "no premature abstraction"). Happy to fold either in if it's a blocker.

OjusWiZard (commented):
Second round on the fix commits (`dd421666`..`a305fbf7`). All four fixes are behaviourally correct, and the tree is green at `a305fbf7`: 880 tests, `black`/`isort`/`flake8`/`darglint`/`mypy`, `check-abciapp-specs`, `autonomy packages lock --check`, FSM spec byte-unchanged. Ten of my twelve threads from round 1 are resolved.

Three new inline notes, plus follow-ups on the two threads I left open.

The one I'd most want a decision on before merge is the serialization bound — bennyjo's point, which I reproduced: the change **halves** the serialized payload rather than bounding it, so the summary's "stays bounded regardless of per-event payload size" overclaims. One max-size event now fits under `MAX_READ_IN_BYTES`; two still reject.

Not filed inline, but flagged for the record:

- **The fail-soft error log never reaches `log.txt`.** `rounds.py:121` uses `logging.getLogger(__name__)`, but `aea-config.yaml:59-80` attaches the `logfile`/`console` handlers to the `aea` logger only, with no `root` section — I reproduced the config and that logger resolves to **zero** handlers, falling through to `logging.lastResort` (unformatted stderr, nothing in the agent logfile). Since detectability is the whole justification for degrading instead of raising, this makes the new path quieter than the `TypeError` it replaced. `AbciAppDB` is constructed with the AEA logger and stores it as `self.logger`, so `self.db.logger.error(...)` from the property lands in `log.txt` and keeps the `caplog` test valid.
- **The one-shot clear removed a self-healing property.** The prune is fully synchronous before the first `yield`, so a merely-slow agent has already pruned — the residual is `act()`-loop starvation. Narrow, but it used to self-correct on the next cycle and now doesn't. Mutation shows the ordering itself is unpinned: moving the prune below the `_fetch_tx_block_number` yield leaves all 300 tests passing.
- **Stale log ordering.** `behaviours.py:412-416` still says "Removing them from the list of tasks to be processed" as the *last* statement, ~55 lines after the prune it announces and after the RPC yield — so a generator cancelled at that yield mutates shared state with no record of what was pruned. Moving it up and putting it in past tense fixes both.

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -376,16 +452,12 @@ class TaskSubmissionAbciApp(AbciApp[Event]):
     }
     cross_period_persisted_keys: FrozenSet[str] = frozenset(
```
OjusWiZard, line 453: **The PR doesn't reach the bound it claims — it halves the payload rather than bounding it.** bennyjo raised this on the `cross_period_persisted_keys` thread; I verified it independently and want to record the mechanism and the numbers, because the summary still says `db.serialize()` "stays bounded regardless of per-event payload size."

Three facts combine:

1. `AbciAppDB.serialize()` dumps **all** retained `_data`, not just the latest period.
2. `AbciAppDB.update()` *appends* to a key's per-period value history (`data.setdefault(key, []).append(value)`) — a period holds every value the key took, not just the last.
3. `cleanup_history_depth_current: null` in `mech_abci/skill.yaml:92`, so that within-period history is never trimmed.

So pre-PR a period held the `create()`-seeded carried copy **plus** `TaskPoolingRound`'s own write — two fat `done_tasks` copies. Post-PR nothing seeds it, so one. That is exactly a halving.

Measured against the real `AbciAppDB` with 500 KB events, simulating the reset-error path (no `create()`, so no `cleanup()`), against `MAX_READ_IN_BYTES = 1048576`:

| events in batch | pre-PR | post-PR |
|---|---|---|
| 1 | 1465 KB — exceeds | 977 KB — fits |
| 2 | 2930 KB — exceeds | 1954 KB — exceeds |

Which matches bennyjo's 525 KB measurement in shape: one max-size event now fits, two still reject.

The reason this path is the one that matters: `RegistrationRound` is reachable *only* via `composition.py:56` `FinishedResetAndPauseErrorRound -> RegistrationRound` — i.e. precisely when `ResetAndPauseRound` did **not** complete, so `create()`/`cleanup()` never ran and the current period is un-trimmed. `RegistrationPayload.encode()` then serializes it whole.

Net: this buys exactly one max-size event of headroom. It's a real improvement and worth landing, but the bound comes from keeping `predict_api_event` out of consensus, not from this change. Could the title and summary say "reduces" rather than "bounds", and is the payload-out-of-consensus work something you'd track as a follow-up?

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -323,43 +325,95 @@ def get_done_tasks(self, timeout: float) -> Generator[None, None, List[Dict]]:
...
+        # read: a task landing mid-copy would silently fall into the
+        # ``task is None`` skip branch below and its metric would be
+        # dropped, making a real desync indistinguishable from a race.
+        submitted_id_set = set(submitted_ids)
+        with self.done_tasks_lock():
+            shared_snapshot = list(self.context.shared_state.get(DONE_TASKS, []))
+            self.context.shared_state[DONE_TASKS] = [
+                task
+                for task in shared_snapshot
+                if str(task.get("request_id")) not in submitted_id_set
```
OjusWiZard: **This line's `str()` is load-bearing and completely untested, while the near-identical line in `remove_tasks_by_id` is covered.**

`0518c00e` inlined the prune here and dropped the trailing `remove_tasks_by_id(submitted_ids)` call, so this is now the copy that runs on the hot path — `remove_tasks_by_id` is reachable in production only through `remove_tasks` at the simulation-failure site (`behaviours.py:1465`).

Mutation run against the 300-test suite, in a throwaway worktree:

```
drop str() here (line 361)                  -> 300 passed   # unguarded
drop str() in remove_tasks_by_id (line 255) ->   2 failed   # guarded
```

So `test_int_shared_state_matches_str_id` — renamed in this round specifically to pin the legacy-`int` case — is guarding the copy that barely runs, and the live path has no coverage at all. If that normalisation regresses, the prune silently no-ops and the delivered batch is re-pooled and re-delivered on-chain: the exact failure this PR exists to prevent.

Two related gaps from the same duplication:

- Replacing `with self.done_tasks_lock():` on line 356 with `if True:` also leaves **300 passed** — the commit titled "snapshot done_tasks under the lock" has nothing pinning the lock.
- The two copies have already diverged in behaviour: `remove_tasks_by_id` reads through `self.done_tasks` (a `deepcopy`) and writes the copies back, whereas this block writes the executor's original dict references. Nothing depends on that identity today, but it's day-one drift.

Cheapest consolidation that keeps the single-snapshot-under-one-lock property the fix was after — have the method return the pre-prune snapshot:

```python
def remove_tasks_by_id(self, submitted_ids: List[str]) -> List[Dict[str, Any]]:
    submitted_id_set = set(submitted_ids)
    with self.done_tasks_lock():
        snapshot = self.done_tasks
        self.context.shared_state[DONE_TASKS] = [
            t for t in snapshot
            if str(t.get("request_id")) not in submitted_id_set
        ]
    return snapshot
```

and here: `shared_snapshot = self.remove_tasks_by_id(submitted_ids)`. One lock acquisition, one filter, one tested surface — and `test_int_shared_state_matches_str_id` starts guarding the live path again.

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -286,7 +338,31 @@ class PostTxSettlementRound(CollectSameUntilThresholdRound):
     def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
         """Process the end of the block."""
         if self.threshold_reached:
-            return self.synchronized_data, Event.DONE
+            done_tasks = cast(SynchronizedData, self.synchronized_data).done_tasks
+            if not done_tasks:
```
OjusWiZard, line 342: The guard itself is right and I'm glad it's here — but the comment justifies it with a path it does not actually protect, and that's the half a future reader will trust.

**The `ResetRound` half doesn't hold.** Trace it with `dd421666` in place:

1. `TaskPoolingRound.end_block` sets `done_tasks=[T1]` **and** `submitted_request_ids=[]` (rounds.py:239-245) — the one-shot clear.
2. Settlement takes `NO_MAJORITY` -> the settlement-internal `ResetRound` -> `db.create(allow_list)` drops `done_tasks`.
3. Retry settles -> here, `done_tasks == []` -> early return -> `submitted_request_ids` stays `[]`.

There was no pending hand-off to preserve; the clear in step 1 already emptied it. So on the `ResetRound` path this guard changes nothing — `T1` is never recorded, never pruned, and gets re-pooled and re-delivered next cycle. The comment says that case is handled; it isn't.

**The delivery-rate half does hold**, and it's the reason to keep the guard: previous cycle writes `[A,B]` -> `ResetAndPauseRound` errors -> `RegistrationRound` -> `UpdateDeliveryRateRound` -> `FinishedWithTxRound` -> settlement -> here, with `TaskPoolingRound` never having run in between, so `[A,B]` is genuinely still pending and the early return preserves it.

Worth noting the `ResetRound` path may already be dead for an unrelated reason: the same `db.create()` also drops `most_voted_tx_hash`, so `CollectSignatureBehaviour` would raise on the retry. That's an argument for deleting the claim rather than trying to make the guard cover it.

Suggestion: keep one line naming the composition fact — `composition.py:51` wires `FinishedTransactionSubmissionRound -> PostTxSettlementRound` for *every* settlement, not just task delivery — which is the genuinely non-obvious bit that stops someone deleting this branch as dead defensive code. Drop the rest; the consequence is already stated verbatim in the commit, the PR, and `test_no_write_when_done_tasks_empty`'s own docstring.

Separately: this early return is completely silent. A batch lost to the `ResetRound` drop and a benign delivery-rate settlement take the identical path with zero log output. A `warning` here that names the pending `submitted_request_ids` and `final_tx_hash` would separate the two.

packages/valory/skills/task_submission_abci/rounds.py
OjusWiZard: Following up after the fix commits. The docstring contradiction is fixed — `0` and `False` are now correctly described as kept — so that part is closed. Two structural points from this thread are still open, and I'd rather flag them than let the thread be resolved on the docstring alone:

- **The accessor mix survives.** `str(task["request_id"])` in the projection with `task.get("request_id")` twice in the predicate. Still correct today purely because comprehension evaluation order runs the `if` first, and still a trap: tidying the guard to `task["request_id"] is not None` puts a `KeyError` inside a consensus `end_block`.
- **The predicate is still in three places** — here, the dedup loop at `rounds.py:197`, and the sort key at `rounds.py:229` (whose `""` default is now dead, since everything reaching `unique_objects` already passed the `:197` guard). A divergence between the first two strands a settled task in `shared_state[DONE_TASKS]` permanently.

Also worth noting the asymmetry is unchanged: `:197` emits `logger.warning` on a dropped id, this one drops silently. Both live call sites currently feed this from data that already passed the `:197` filter, so the silent branch is unreachable in practice — which argues for making it loud rather than tolerated.

One helper used by all three sites closes all of it:

```python
def normalise_request_id(task: Dict[str, Any]) -> Optional[str]:
    raw = task.get("request_id")
    if raw is None or raw == "":
        return None
    return str(raw)
```

Happy for this to be a follow-up rather than this PR — just don't want it lost.

packages/valory/skills/task_submission_abci/behaviours.py
OjusWiZard: Following up after the fix commits. The main thing in this thread is properly fixed: `0518c00e` takes one snapshot under `done_tasks_lock` and drives both the prune and the metrics map from it, so the torn read is gone — and the prune moving above the metrics loop and the `_fetch_tx_block_number` yield also means a metrics exception or a hung RPC can no longer skip it. The ~1/N sample-count change is now documented in the PR body too. Both good.

Two smaller points from this thread are still open:

- **The skip branches are still `logger.debug`**, and `LOG_LEVEL` defaults to `INFO`, so they are absent rather than merely quiet. They also still conflate two different conditions: a task executed by *another* agent (normal — pooling unions every participant's payload) and this agent's own executor not having finished the request yet. The second is not benign: the prune leaves no tombstone, and now that `TaskPoolingRound` clears `submitted_request_ids` the id list can't act as a suppression set either, so a late-finishing task lands back in `DONE_TASKS` and gets re-pooled. `_get_deliver_tx`'s simulation catches that for non-marketplace tasks (`behaviours.py:1457-1466`) but the marketplace batch path has no per-task gate. One aggregate line per cycle — `pruned N/M; K ids not present locally` — would give the number an operator actually needs.
- **The metrics loop still sits inline in the generator** despite containing no `yield`. Extracting it to a plain `_emit_delivery_time_metrics(submitted_ids, shared_snapshot)` would make `handle_submitted_tasks` read as four steps. Combined with the `remove_tasks_by_id`-returns-the-snapshot suggestion I left on line 361, the body between the id read and the block-number fetch collapses to two lines.

Both are advisory — fine as follow-ups.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Fixed the summary and rounds.py:453 area, and added the numbers you measured to the PR body. Title kept for now since it still describes the direction of the change. New wording in the summary: 'Reduces the serialized synchronized_data.db size by the one carried copy per period... Does not fully bound the serialized size: a batch with two max-size predict-api events in the same period still rejects. Bounding it in the strict sense requires keeping predict_api_event out of consensus entirely (follow-up).' Predict_api_event follow-up is now listed explicitly in the PR body.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in f7cfd2d5. remove_tasks_by_id now returns the pre-prune snapshot; handle_submitted_tasks calls it and drives the metrics loop from that snapshot. One lock acquisition, one filter, one tested surface. test_int_shared_state_matches_str_id now guards the live path.

Added three mutation-catching tests to lock down the invariants you found unpinned:
- test_prune_acquires_done_tasks_lock — dropping the `with` now fails.
- test_str_normalisation_matches_legacy_int_on_live_path — dropping the str() on the shared_state side of the match now fails end-to-end.
- test_prune_runs_before_block_number_fetch — moving the prune below the yield now fails.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Both fixed in f7cfd2d5. Shortened the comment to just the composition-wiring fact (the genuinely non-obvious bit) and dropped the ResetRound claim you're right doesn't hold. Added a warning log at the branch that names the pending submitted_request_ids so a benign delivery-rate settlement is distinguishable from an actual clobber landing on the same path.

Agree the ResetRound path is probably dead for the most_voted_tx_hash reason anyway. Kept the guard for the delivery-rate case which does hold.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Both structural points logged as explicit follow-ups in the PR body (predicate DRY + get/[] accessor mix). Not folding into this PR round to keep the diff scoped, but tracked so they don't get lost. Your normalise_request_id helper is a good shape for that follow-up.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Both points logged as explicit follow-ups in the PR body. The 'skip branches at debug' one is genuine — with the id list one-shot-cleared now, a late-finishing task really can't be suppressed by the id set, so aggregate 'pruned N/M; K ids not present locally' logging at info level is worth doing. Deferring to a follow-up rather than folding it in here.

DIvyaNautiyal07:
Thanks for the second pass. Inline replies on each thread. Round-3 pushed as f7cfd2d5.

Correctness fixes for the three real issues in the review body:

- Fail-soft error log now goes through `self.db.logger` (the AEA-configured logger), so an invariant-broken value actually lands in `log.txt` instead of `logging.lastResort`. Detectability is restored.
- `handle_submitted_tasks` prune consolidated back onto `remove_tasks_by_id`, which now returns the pre-prune snapshot and drives the metrics loop from it. One lock acquisition, one tested surface. The str-normalisation and lock use are both on the hot path again with pinning tests.
- The "pruned" log now fires next to the actual prune, past tense, before the `_fetch_tx_block_number` yield. A cancellation at the yield no longer leaves a silent mutation.

Inline fixes:

- rounds.py:342 empty-batch guard comment shortened to the composition-wiring fact; dropped the ResetRound claim you're right doesn't hold. Added a `warning` naming the pending id list so the delivery-rate benign case is distinguishable from an actual clobber.
- PR summary updated. Added your measured table (500 KB events, single vs two-event batch) so the "reduces vs bounds" distinction is spelled out concretely.

Three mutation-catching tests added that would have caught what you found:
- `test_prune_acquires_done_tasks_lock` — dropping the `with` around the prune fails.
- `test_str_normalisation_matches_legacy_int_on_live_path` — dropping the `str()` on the live path fails end-to-end.
- `test_prune_runs_before_block_number_fetch` — moving the prune below the yield fails.

Advisory items logged as follow-ups in the PR body: (a) predicate DRY + get/[] accessor mix in `extract_request_ids`, (b) metrics loop extraction, (c) aggregate `pruned N/M; K missing locally` log at info level, (d) predict_api_event out of consensus to truly bound the serialize size. Happy to fold any in if you'd rather they land in this PR.

OjusWiZard (approved):
## Approving

Reviewed across three rounds. The hand-off redesign is sound and the code now matches what the PR claims. Verified at `f7cfd2d5`, not taken on trust:

- **883 tests pass**; `black` / `isort` / `flake8` / `darglint` / `mypy` clean; `check-abciapp-specs` OK; `autonomy packages lock --check` verifies; the FSM spec is byte-unchanged across the whole PR.
- **The round-2 fixes are genuinely pinned.** I re-ran the three mutants that previously survived: dropping `str()` on the prune now fails 3 tests, dropping `done_tasks_lock` fails 1, moving the prune below the `_fetch_tx_block_number` yield fails 1. All three passed 303/303 before this round.
- **The `db.logger` swap works.** `SharedState.setup` passes `logger=self.context.logger` (`abstract_round_abci/models.py:509-513`), so the fail-soft invariant error reaches `log.txt` — the previous `logging.getLogger(__name__)` resolved to zero handlers and fell through to `logging.lastResort`.

Three things I raised earlier that I checked and **withdraw** — recording them so nobody re-opens them:

- The `deepcopy` under the lock is not a contention problem: `deepcopy` treats `str` as atomic, so the large `predict_api_event` payloads aren't copied (~0.5 ms for 50 tasks × 350 KB), and `get_done_tasks` already deepcopies the same list *unlocked* once per second.
- The deep-copy write-back can't lose a mutation: every field write on `done_task` happens before the guarded append, and `_reset_executing_task` rebinds rather than mutates.
- The unlocked `del storage[idx]` in `task_execution/handlers.py` isn't a race with the prune: the append runs on the main `act()` loop after `_async_result.done()`, so append, prune and eviction are all main-thread.

### Not blocking, but worth a look

1. **The simulation-failure prune is unasserted** — `behaviours.py:1460`. Deleting `self.remove_tasks([task])` leaves all 883 tests passing. `test_skips_deliver_with_failed_simulation` seeds `_synchronized_data.done_tasks` but never `shared_state[DONE_TASKS]`, so the prune runs against an empty dict. It's the one call site of the newly-changed signature that was left alone; two lines pin it.

2. **A simulation-failed task is recorded as delivered fleet-wide** — `behaviours.py:1453-1461` with `rounds.py:356`. The task is dropped from the multisend but stays in consensus `done_tasks`, so its id rides into `submitted_request_ids` and every agent prunes it as delivered. All three deliver paths are affected (the offchain and marketplace paths `continue` without a local prune, but the id still carries). On-chain requests recover via the marketplace sweep; offchain in-memory ones don't. **Pre-existing** — the pre-PR code pruned off the same id set — so not a regression here, but it's the inverse of the guarantee this PR is built around and shouldn't be inherited silently.

3. **`remove_tasks_by_id` returns the wrong list** — `behaviours.py:244-264`. It returns the full pre-prune snapshot, but only the removed tasks are ever consumed: the metrics map is built over everything while lookups are only `for req_id in submitted_ids`, and a surviving task's key can never match by construction. Returning the popped tasks would make the name honest and let the rationale docstring go.

4. **Two logs slightly overstate.** `behaviours.py:360-363` asserts a completed local prune using the fleet-wide consensus id set without computing what was actually removed — after a restart it logs a clean `Pruned tasks with ids [...]` while removing zero. And `rounds.py:350-354` reads `preserving pending hand-off submitted_request_ids=[]` on the benign delivery-rate path, which is byte-identical to the real-clobber case it was added to distinguish; branching on truthiness would separate them.

5. **Comment volume crept back** against the repo's minimal-comments rule — the 6-line block above the prune call, the `remove_tasks_by_id` rationale paragraph, and the three new test docstrings (one embeds review-thread history and pins commit `0518c00e`; one sentence doesn't parse). The `composition.py` wiring comment at `rounds.py:343-345` is the one that clearly earns its place.

The known limitation is now honestly stated in the summary — this reduces the serialized payload by the one carried copy per period rather than bounding it, and two max-size events in a batch still reject. Keeping `predict_api_event` out of consensus, the `normalise_request_id` helper, the metrics-loop extraction and the labelled skip counter are all recorded under Follow-ups.

bennyjo still has open threads on `rounds.py:440` and `rounds.py:455` — those are theirs to close, not mine.

packages/valory/skills/task_submission_abci/rounds.py
bennyjo: Closing this one: the retry path can't reach `PostTxSettlementRound` today, because the settlement `ResetRound` also drops `most_voted_tx_hash` and `SignatureBehaviour` raises on it first. Pre-existing and out of scope here. If that path is ever revived, the id hand-off gap returns, and a small cross-period `pending_request_ids` written at pooling time and promoted at settlement is the shape that survives it.

bennyjo (approved):
Round 3 closes everything I raised. 883 tests pass locally on f7cfd2d5. The settlement ResetRound gap is real in isolation but unreachable today (see thread), and the payload-out-of-consensus follow-up is tracked in the body.

Adamantios (approved):

PR #483, opened 2026-09-02
fix(task_submission): recognise FixedPriceTokenUSDC paymentType

DIvyaNautiyal07:
## Summary

Two independent fixes on the same `task_submission` / `task_execution` surface, bundled because they landed against the same open PR:

1. **Recognise `FixedPriceTokenUSDC` paymentType** so USDC-denominated mechs route their withdraw through the ERC20 branch instead of the native-transfer fallback. Fixes the polygon-21 GS013 loop.
2. **Write the mech's paymentType hash on every predict-api event** so `mech_requests.payment_type` stops being NULL on 100% of rows. Closes the downstream marketplace-app "Payment cell blank on off-chain rows" gap Tanya flagged on autonolas-frontend-mono#446.

Each fix stands alone; either can be dropped without affecting the other.

---

### Background — the on-chain failure this fixes

The polygon 21 mech at `0x76A9A29441c7ACd072b03e63911F0E177DE56Ab7` is a `MechFixedPriceTokenUSDC` (`autonolas-marketplace/contracts/mechs/token/usdc/MechFixedPriceTokenUSDC.sol`). Its `paymentType()` returns `keccak256("FixedPriceTokenUSDC") = 0x6406bb5f31a732f898e1ce9fdd988a80a808d36ab5d9a4a4805a8be8d197d5e3`.

Before this fix, that value wasn't in `behaviours.py:107-118`, so the withdraw branch at line 680 (`if mech_type.hex() in [PAYMENT_TYPE_TOKEN_NVM, PAYMENT_TYPE_TOKEN]`) evaluated False and fell through to `_get_transfer_tx` at line 701 — the NATIVE-transfer path. That encoded `mech.exec(to=service_owner, value=<usdc_amount>, data=0x, operation=CALL)`, attempting to send `<usdc_amount>` wei of POL from the mech contract.

The mech contract had just received USDC from `processPaymentByMultisig` inside the same Safe multisend, but zero POL. The native transfer reverted, the atomic multisend rolled back with `GS013`, and the FSM entered a settlement retry loop rebuilding the same failing multisend every ~90s. Symptom in agent logs: `Payload is invalid for 0x...! Cannot continue. Received: ContractLogicError('execution reverted: GS013', ...)`.

Confirmed live on polygon 21 on 2026-09-02 after agent-deployments PR #336 rolled out the withdraw trigger; rolled back via agent-deployments PR #338.

### What changed

- `packages/valory/skills/task_submission_abci/behaviours.py`
  - Add `PAYMENT_TYPE_TOKEN_USDC = "6406bb5f..."` constant next to the existing four
  - Include it in the ERC20-branch list at line 680 alongside `PAYMENT_TYPE_TOKEN` and `PAYMENT_TYPE_TOKEN_NVM`
- `packages/valory/skills/task_submission_abci/tests/test_behaviours.py`
  - New regression test `test_fixed_price_token_usdc_routes_to_erc20_branch` that pins the routing: asserts `_get_token_transfer_tx` is called AND `_get_transfer_tx` (native) is NOT called for a USDC paymentType

### What deliberately did NOT change

- `behaviours.py:619` (NVM `_adjust_mech_balance` branch) — USDC does not use `tokenCreditRatio`. The USDC contract at `autonolas-marketplace/contracts/mechs/token/usdc/` reuses `BalanceTrackerFixedPriceToken.sol` (no dedicated `BalanceTrackerFixedPriceTokenUSDC.sol` exists), so no ratio adjustment applies.
- Existing misleading constant `PAYMENT_TYPE_TOKEN_NVM` — its actual hash matches `NvmSubscriptionTokenUSDC`, not `NvmSubscriptionToken`. Not renaming to avoid churn on working code.

### Round-3 review pickups (all on Fix 1)

- **Warn on unrecognised paymentType.** The withdraw-path else-branch used to silently fall through to `_get_transfer_tx` for any hash outside the token-payment list — the exact regression this PR fixes per-type. Split the token check from a new `KNOWN_NATIVE_PAYMENT_TYPES` set; the else-branch warns when the paymentType is in neither set, with the observed hash + mech address + remediation string (`add it to TOKEN_PAYMENT_TYPES to avoid a GS013 revert on-chain`). Legitimate native rails stay silent.
- **`TOKEN_PAYMENT_TYPES` frozenset.** Inline `[PAYMENT_TYPE_TOKEN_NVM, PAYMENT_TYPE_TOKEN, PAYMENT_TYPE_TOKEN_USDC]` at the check site was two-place maintenance; extracted to a module-level `frozenset` alongside the constants.
- **Sentinel test cleanup.** Dropped the belt-and-suspenders `MagicMock(side_effect=AssertionError(...))` on `_get_transfer_tx` in the USDC routing test; rely on `.assert_not_called()` alone for a clean pytest failure message.
- **New regression tests.** `test_unrecognised_payment_type_warns_before_native_fallback` and `test_known_native_payment_type_does_not_warn` pin the warning fires on unknown hashes and doesn't fire on legitimate native rails.

---

### The gap this fixes

Predict-api's `mech_requests.payment_type` was NULL on **100% of rows** — measured directly against prod on 2026-09-02: 0 of 11,326,264 rows populated, including all 14,813 rows written since the mech-analytics v0.0.19 deploy went live. The NULL propagates through mech-analytics' ETL projection into `per_request_scores.payment_type`, and on the marketplace FE that means:

- `mapPaymentToFee` cannot pick correct decimals for `delivery_rate` (the same `"1000000"` reads as 1 USDC at 6-decimal or 0.000000000001 xDAI at 18-decimal depending on the rail)
- On-chain rows fall back to the marketplace-subgraph twin's `feeUnit` / `feeRaw`, so they render OK
- **Off-chain rows have no twin**, so their Payment cell is always blank — the exact failure mode Tanya flagged on autonolas-frontend-mono#446 as the caveat blocking the default-ON flip

### Root cause

`_build_predict_api_event` in `task_execution/behaviours.py` read `executing_task.get("payment_type")` and coerced through `str(x or "") or None`. No code path in the marketplace-contract readers (`get_marketplace_undelivered_reqs` / `get_marketplace_request_events` in `packages/valory/contracts/mech_marketplace/contract.py`) or in the off-chain request handler ever populated `payment_type` on the task dict, so the fallback was always None. The mech was faithfully sending `payment_type=None` on every event; the schema was NULL because upstream never wrote anything.

### What changed

- `packages/valory/skills/task_execution/behaviours.py` (`_build_predict_api_event`, line 2115)
  - Read `self.payment_model.hex()` instead of `executing_task.get("payment_type")`. `self.payment_model` is the process-wide cache in `shared_state[PAYMENT_MODEL]`, populated once at startup by `task_execution.handlers._handle_get_undelivered_reqs` from the mech contract's `paymentType()` view. Zero contract calls per event.
  - Bytes32 hash is emitted as lowercase hex without the `0x` prefix (matching how the constants are declared in `packages/valory/contracts/olas_mech/contract.py:49-54`). No human-name conversion in the pipeline — the marketplace FE (`apps/marketplace/common-util/mechAnalytics/service-activity.ts`) does the hash → `native` / `usdc` / `nvm_subscription` label mapping at render time in a follow-up autonolas-frontend-mono PR.
- `packages/valory/skills/task_execution/tests/test_behaviours.py`
  - `test_build_predict_api_event_writes_hex_payment_type_from_shared_state` — cache set to bytes → event carries lowercase hex without `0x`.
  - `test_build_predict_api_event_writes_none_payment_type_when_shared_state_empty` — startup-race guard: `self.payment_model` is None → event's `payment_type` is None (no `AttributeError` on `.hex()`).
  - `test_build_predict_api_event_payment_type_ignores_executing_task_key` — a bogus `executing_task['payment_type']` does not shadow the cache. Prior shape was coupled to that key; the new authoritative read is not.

### What deliberately did NOT change

- `task_submission_abci/behaviours.py:2691` (`_build_request_only_event`) keeps `"payment_type": None`. The sweep's `priority_mech` may be another mech whose payment model we don't authoritatively cache, and the delivery event's `ON CONFLICT UPDATE` fills the row's `payment_type` when the actual deliverer writes. Emitting a possibly-wrong hash from the sweep would risk shadowing the correct value under race conditions.
- Pipeline-side conversion of hash → human name. The mech carries only bytes32; every downstream layer (predict-api storage, mech-analytics projection, mech-analytics API) forwards the hash unchanged. This makes new payment rails additive — a new marketplace rail flows end-to-end without any pipeline code change; only display consumers need to map the new hash.

### Why zero contract calls per event

`self.payment_model` is populated once at agent start via the `get_mech_type` handler path. `_get_mech_payment_type` in `task_submission_abci/behaviours.py:797` explicitly returns the cache without a contract call when the address is the mech's own (line 800-801). For step-in flows, `_filter_out_incompatible_reqs` guarantees the mech only processes tasks whose priority mech shares the mech's own payment model, so `self.payment_model` is the authoritative value for every delivered row regardless of whether the delivery originated as own-mech-priority or step-in.

---

## Local checks passed

- `tomte tox -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` — all OK
- `tomte tox -e check-abci-docstrings -e check-abciapp-specs -e check-handlers` — all OK
- `tomte tox -e check-doc-hashes` — all OK
- `autonomy packages lock --check` — verification successful
- `pytest packages/valory/skills/task_submission_abci/tests/test_behaviours.py packages/valory/skills/task_execution/tests/test_behaviours.py` — 357 passed (Fix 1 regression tests + round-3 warn tests + 3 new tests for Fix 2)

Packages relocked via `autonomy packages lock`; `autonomy push-all` still needed before the branch is pushed for review — the IPFS registry entries under the new hashes have to exist before Propel can pull them.

## Rollout after merge

1. Merge this PR
2. Bump the service hash on `single_polygon_mech_mm_predict_21` in the agent-deployments repo to pick up the new mech-service package: `bafybeicfblue2fsikhlqvw52viwyszbracowbdf5loelir73pjrbnxwatq` (relocked and pushed to IPFS by `autonomy push-all` after both fixes + round-3 review pickups landed on the branch)
3. Re-open the lower-`PROFIT_SPLIT_BALANCE` config change (was in agent-deployments PR #336, rolled back in PR #338) alongside the service-hash bump
4. Redeploy `single_polygon_mech_mm_predict_21` on Propel
5. Watch agent logs for **Fix 1**: expect `Splitting profits` → `Split ... into {service_owner: N}` → Safe multisend lands on-chain → `Total mechs balances: 0` on the next split-check
6. Watch **Fix 2** rollout by querying `predict-api-db` after the redeploy: `SELECT count(*), count(payment_type) FROM mech_requests WHERE created_at > '<redeploy timestamp>'`. Expect `count(payment_type)` to grow with `count(*)` on rows written by the redeployed mech. The autonolas-frontend-mono follow-up PR that maps hex → label needs to land + deploy before the FE Payment cell renders correctly for those rows.
7. If Fix 1 is clean, extend the same treatment to `polygon_mech_mm_predict_25` and `polygon_mech_mm_predict_44`. Fix 2 benefits from any mech redeploy identically — the more mechs on the new package, the faster the tail of NULL payment_type shrinks.

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -9,12 +9,12 @@
         "connection/valory/websocket_client/0.1.0": "bafybeid4cnbjpmzcsw3usg7umufgqyod2u3xfw5uso3vukt3n5hfuxvdea",
         "skill/valory/contract_subscription/0.1.0": "bafybeihaaqld7ntndyhoxu2o4jecjocclo4gsvmqdz3dpgmqeckq6j5cb4",
         "skill/valory/delivery_rate_abci/0.1.0": "bafybeifjiazvf3qhr2264i55u4tlvmyf6ztgak2ndqolujt4kw6jiy7l3m",
-        "skill/valory/mech_abci/0.1.0": "bafybeiehxuzywr7ld3bosooqvhewyrpbagmz2kljep6i3mls43jmtnktie",
+        "skill/valory/mech_abci/0.1.0": "bafybeid53ko6e7f2iqieve4t7cky4w7qpbztbf6pdlee3vnry2mjrnszra",
         "skill/valory/task_execution/0.1.0": "bafybeifbqbo5fmgxxrs4gpkuqrukg4qvt6q2l2sd4pgpqsvlmcz2zbo3fy",
-        "skill/valory/task_submission_abci/0.1.0": "bafybeidxzddmqxhovneuxtplkdcmlbf7xlm3h75jhps7eddrwuuqnozeee",
+        "skill/valory/task_submission_abci/0.1.0": "bafybeiarzinfeymv6kiymkdsfvreivod7sf5hkudfilamsrl5a5eeqvkci",
         "skill/valory/websocket_client/0.1.0": "bafybeiht2ify5vkseeulutzjwkw5vcmhcwtun2o5ehn6hya6uz4eanua2m",
-        "agent/valory/mech/0.1.0": "bafybeighyndmb65rgptrifklfuimhvotz27dciemqyrnh2xbehovq6t5hm",
-        "service/valory/mech/0.1.0": "bafybeietnx4ctnxqswbtrcw7z6qrdncsiimf7xsbgp7pyw7mr7rblvxuhi"
+        "agent/valory/mech/0.1.0": "bafybeigsq263pcxuxisq47kk2yasl4o3xzdb2dputd4yqzpi6efeyepnvi",
+        "service/valory/mech/0.1.0": "bafybeiazchfmdlyc32hypspvg4zpydxjubkvuggvjzsbzq3t7g4lhkqyii"
     },
     "third_party": {
         "protocol/open_aea/signing/1.0.0": "bafybeifsjmldwyki3beqyvdt5lzenrg6wyrqaar5plc5rpnvtc4zlentye",
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -9,12 +9,12 @@
         "connection/valory/websocket_client/0.1.0": "bafybeid4cnbjpmzcsw3usg7umufgqyod2u3xfw5uso3vukt3n5hfuxvdea",
         "skill/valory/contract_subscription/0.1.0": "bafybeihaaqld7ntndyhoxu2o4jecjocclo4gsvmqdz3dpgmqeckq6j5cb4",
         "skill/valory/delivery_rate_abci/0.1.0": "bafybeifjiazvf3qhr2264i55u4tlvmyf6ztgak2ndqolujt4kw6jiy7l3m",
-        "skill/valory/mech_abci/0.1.0": "bafybeid53ko6e7f2iqieve4t7cky4w7qpbztbf6pdlee3vnry2mjrnszra",
-        "skill/valory/task_execution/0.1.0": "bafybeifbqbo5fmgxxrs4gpkuqrukg4qvt6q2l2sd4pgpqsvlmcz2zbo3fy",
-        "skill/valory/task_submission_abci/0.1.0": "bafybeiarzinfeymv6kiymkdsfvreivod7sf5hkudfilamsrl5a5eeqvkci",
+        "skill/valory/mech_abci/0.1.0": "bafybeialhw2p6dqyjtjuwzboizswh2wtpimwf556aje2e3sczzxkq7yzgi",
+        "skill/valory/task_execution/0.1.0": "bafybeihgusy4jnab5zak6ouikblzmffbwttzeag6xkjpdzrbx4rtscibwu",
+        "skill/valory/task_submission_abci/0.1.0": "bafybeiabk5sffb4i2doypwjbznr6zw6kwr3pnjttwm4onrydfbfyavrzce",
         "skill/valory/websocket_client/0.1.0": "bafybeiht2ify5vkseeulutzjwkw5vcmhcwtun2o5ehn6hya6uz4eanua2m",
-        "agent/valory/mech/0.1.0": "bafybeigsq263pcxuxisq47kk2yasl4o3xzdb2dputd4yqzpi6efeyepnvi",
-        "service/valory/mech/0.1.0": "bafybeiazchfmdlyc32hypspvg4zpydxjubkvuggvjzsbzq3t7g4lhkqyii"
+        "agent/valory/mech/0.1.0": "bafybeicgbuajc22iy47fgaulxi2b5alfyhxq3z6znqgyumxaxgqqc2iruu",
+        "service/valory/mech/0.1.0": "bafybeihltyh5umxwagb2kqrlic7sezyex7osiosuvw4w6um7242wzs43ny"
     },
     "third_party": {
         "protocol/open_aea/signing/1.0.0": "bafybeifsjmldwyki3beqyvdt5lzenrg6wyrqaar5plc5rpnvtc4zlentye",
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -9,12 +9,12 @@
         "connection/valory/websocket_client/0.1.0": "bafybeid4cnbjpmzcsw3usg7umufgqyod2u3xfw5uso3vukt3n5hfuxvdea",
         "skill/valory/contract_subscription/0.1.0": "bafybeihaaqld7ntndyhoxu2o4jecjocclo4gsvmqdz3dpgmqeckq6j5cb4",
         "skill/valory/delivery_rate_abci/0.1.0": "bafybeifjiazvf3qhr2264i55u4tlvmyf6ztgak2ndqolujt4kw6jiy7l3m",
-        "skill/valory/mech_abci/0.1.0": "bafybeialhw2p6dqyjtjuwzboizswh2wtpimwf556aje2e3sczzxkq7yzgi",
+        "skill/valory/mech_abci/0.1.0": "bafybeifuuif7ctpas6u7j75qkhda6zpm5uzrnotetzoux4uujvjyflk5pi",
         "skill/valory/task_execution/0.1.0": "bafybeihgusy4jnab5zak6ouikblzmffbwttzeag6xkjpdzrbx4rtscibwu",
-        "skill/valory/task_submission_abci/0.1.0": "bafybeiabk5sffb4i2doypwjbznr6zw6kwr3pnjttwm4onrydfbfyavrzce",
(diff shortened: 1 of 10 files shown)


Adamantios (approved):

OjusWiZard (approved):
Approving — both fixes are correct and are strict improvements over `main`. Nothing below blocks the merge.

### Verified

- `PAYMENT_TYPE_TOKEN_USDC` = `keccak256("FixedPriceTokenUSDC")` ✅ — recomputed independently.
- The step-in claim holds: `_filter_out_incompatible_reqs` only keeps a task when `req_mech_pm == self.payment_model` (`task_execution/behaviours.py:757`), so the mech's own payment model really is authoritative for every delivered row.
- The four regression tests genuinely pin the fix — I checked out the merge-base `behaviours.py` in both skills and re-ran them; all four fail against pre-fix code. 828 tests pass at HEAD.

### Worth acting on (two one-liners)

**1. `payment_model` is annotated `-> None` but returns `bytes`, and it's on this PR's critical path.**

`task_submission_abci/behaviours.py:244-247`:

```python
@property
def payment_model(self) -> None:
    """Get the mech's payment model."""
    return self.context.shared_state.get(PAYMENT_MODEL)
```

`_get_mech_payment_type` (line 820-824) returns this as `Optional[bytes]`. Because `None` inhabits `Optional[bytes]`, mypy raises nothing — the annotation is actively wrong rather than merely loose, and a reader trusting the signature would conclude the cache branch always falls through to the network fetch. That value feeds straight into the new `mech_type.hex() in TOKEN_PAYMENT_TYPES` check at line 697. Pre-existing, but this PR makes it the routing decision. Fix: `-> Optional[bytes]`.

**2. The new test docstring's "startup race" framing is inaccurate.**

`task_execution/tests/test_behaviours.py` (the `..._when_shared_state_empty` test) says `_ensure_payment_model` "gates the `act` step on this cache being populated". It doesn't — `task_execution/behaviours.py:489-490` returns `True` immediately when `not use_mech_marketplace`, never requesting the model at all. `use_offchain` is an independent kwarg (`models.py:145`) and `mech_marketplace_address` defaults to `ZERO_ADDRESS` in every `service.yaml` / `aea-config.yaml` entry, so on a `use_offchain=True` + no-marketplace deployment `payment_type` stays `null` *permanently*, silently — the same 100%-NULL symptom this fix targets, just on a different config axis. Either correct the docstring or decouple the payment-model lookup from `use_mech_marketplace`.

### Worth knowing

**3. `NvmSubscriptionToken` is missing from the repo entirely.** Recomputing all six hashes:

| preimage | hash | present? |
|---|---|---|
| `FixedPriceTokenUSDC` | `6406bb5f…` | ✅ added here |
| `NvmSubscriptionTokenUSDC` | `0d6fd99a…` | ✅ (as `PAYMENT_TYPE_TOKEN_NVM`) |
| `NvmSubscriptionToken` | `0eafaed2…` | ❌ **absent** |

So the real `NvmSubscriptionToken` is unhandled at *both* line 636 (no `tokenCreditRatio` balance adjustment) and line 697 (falls through to native → GS013). Exactly the bug class this PR fixes. The new warning is precisely what would surface it, which is a good argument for the warning having been worth adding.

**4. Warn-and-proceed is the only non-fatal branch in `get_split_profit_txs`.** Every sibling failure does `return None` (lines 611, 624, 638, 704, 728). The new branch logs that it is about to encode a transfer it names as GS013-reverting, then encodes it. That tx is batched via `all_txs.extend(split_profit_txs)` (line 1443) alongside `offchain_deliver_txs` (1448) and `marketplace_deliver_txs` (1461), so the revert fails the **entire multisend**, taking unrelated already-completed deliveries down with it. `main` did this silently too, so the warning is still a net gain — but `logger.error(...)` + `return None` would be more consistent with every neighbour. Worth a follow-up rather than a change here.

### Nits (non-blocking)

- **Comments.** The block comments above `TOKEN_PAYMENT_TYPES` and `KNOWN_NATIVE_PAYMENT_TYPES` restate the GS013 root cause the PR description already covers at length. Same for the 5–8 line test docstrings — the immediately adjacent `test_token_mech_success_with_token_transfer` uses a one-liner.
- **Test duplication.** The two warn/no-warn tests are near-identical, as are the three predict-api ones; `pytest.mark.parametrize` is already used in both files and would take ~180 lines to ~55. (The nested `if` inside `else:` is right as-is — an `elif` chain would duplicate the `_get_transfer_tx` call.)
- **Coverage.** `PAYMENT_TYPE_TOKEN_NVM` is never exercised through the refactored `TOKEN_PAYMENT_TYPES` check, and `PAYMENT_TYPE_NATIVE_NVM` is never tested against the no-warn branch.
- The warning fires inside `for receiver_address, amount in split_funds.items()`, so it's once per payout rather than once per mech.

PR #484, opened 2026-09-02
fix(task_submission): move predict_api_event out of consensus

DIvyaNautiyal07:
## Summary

Stashes the per-task predict-api event in agent-local `shared_state[PREDICT_API_EVENTS]` keyed by `str(request_id)` instead of embedding it in the `done_task` dict. The consensus-replicated `done_task` now carries only its small fields (request_id, tool, task_result, is_offchain, etc); the ~30-60 KB `raw_content` payload no longer rides Tendermint replication.

## Why

Follow-up to [mech#482](https://github.com/valory-xyz/mech/pull/482) which reduced but did not bound the serialized `done_tasks`. A batch of 15-20 tasks each carrying a 30-60 KB `predict_api_event` still pushed one period's `done_tasks` past `MAX_READ_IN_BYTES = 1 MiB` and rejected the next `RegistrationPayload`. Reproduced on `single_mech_mm_predict_2359` earlier today: 3 crash-loop restarts in 6 minutes with individual done_tasks measured at 17 KB / 28 KB / 57 KB.

Moving the payload out of consensus caps the consensus payload size regardless of per-event size.

## Changes

- `task_execution._finalize_done_task`: writes the event to `shared_state[PREDICT_API_EVENTS][str(req_id)] = {"event": ..., "written_at": time.time()}` instead of `done_task["predict_api_event"] = ...`. On every write, runs a TTL sweep that drops entries older than `PREDICT_API_EVENT_TTL_SECONDS` (24h) and any malformed entries (non-dict, or non-numeric `written_at`). This bounds cache growth even for an agent whose settlements keep failing and never trigger the primary prune.
- `task_execution.ContractHandler.setup`: initialises `shared_state[PREDICT_API_EVENTS] = {}`.
- `task_submission_abci.PostTxSettlementBehaviour._extract_offchain_events`: iterates `done_tasks` and looks each event up in local `shared_state` by request_id. A miss on a task another agent executed logs DEBUG (normal, that agent posts it). A miss on a task **this** agent executed (identified by `task_executor_address == self.context.agent_address`) logs WARNING, since it's an ingress-side drop.
- `task_submission_abci.TaskPoolingBehaviour.handle_submitted_tasks`: after `remove_tasks_by_id(submitted_ids)`, drops the submitted request_ids from `shared_state[PREDICT_API_EVENTS]`. The prune is intentionally NOT inside `remove_tasks_by_id`, since that method is also called pre-settlement from `TransactionPreparationBehaviour.get_payload_content` on the simulation-skip path.
- Divergence-warning check (`use_offchain=False` on the egress side with ingress producing events) now watches `shared_state[PREDICT_API_EVENTS]` instead of scanning `done_tasks`.

## Intentional semantic change

Pre-PR, a deliver-tx simulation failure would still POST the event (the event rode consensus and `remove_tasks` only touched `shared_state[DONE_TASKS]`). Post-PR, a simulation-skipped task never gets a delivered event POSTed. This is the intended behaviour: no delivered event for a task that was never delivered, and `_sweep_pending_undelivered` still emits a request-only event later. Pinned by `test_pre_settlement_remove_tasks_preserves_predict_api_cache`.

## Multi-agent + durability tradeoff

Pre-PR every agent had every event in consensus (durable at TaskPoolingRound quorum, replayable from Tendermint blockchain) and every agent POST'd; predict-api dedups on `request_id`. Post-PR only the executing agent has the event locally, from `_finalize_done_task` completion through the entire settlement pipeline (randomness, select-keeper, collect-signature, finalize, validate, post-settlement). If that agent restarts anywhere in that window the row is lost, and there is no cross-agent redundancy on the POST — this is a materially longer and less-redundant window than the pre-PR one.

For Phase 1 dark rollout the subsystem is explicitly fail-soft; the kv_store replay buffer is a known follow-up that will close this window.

## Test plan

- [x] Unit tests pass locally (895 tests across `task_execution` + `task_submission_abci`).
- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e darglint -e mypy` clean.
- [x] `tomte tox -e check-abci-docstrings -e check-abciapp-specs` OK.
- [x] `autonomy packages lock --check` OK.
- [ ] CI: full pipeline.
- [ ] Post-merge: canary on `single_mech_mm_predict_2359` (currently crash-looping on this exact edge case); confirm RegistrationPayload stays under 1 MiB across many settlement cycles and no `1048576` errors.

### DIFF
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -87,6 +87,7 @@ LAST_SUCCESSFUL_EXECUTED_TASK = "last_successful_executed_task"
 LAST_READ_ATTEMPT_TS = "last_read_attempt_ts"
 INFLIGHT_READ_TS = "inflight_read_ts"
 REQUEST_ID_TO_DELIVERY_RATE_INFO = "request_id_to_delivery_rate_info"
+PREDICT_API_EVENTS = "predict_api_events"
 # Shared-state keys owned by the MechHttpHandler (handlers.py); mirrored here
 # because the off-chain finalize path writes the response envelope for the
 # ``/fetch_offchain_info`` polling endpoint and clears the buffered request
@@ -1773,40 +1774,18 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         # pop the data key value as it's bytes which causes issues
         # with json dumps and not required anywhere
         done_task.pop("data", None)
-        # Attach the offchain predict-api event payload before the
-        # IN_MEMORY_REQUESTS pop below; the post-settlement behaviour in
-        # task_submission_abci reads it off ``synchronized_data.done_tasks``
-        # (i.e. after consensus replication) and posts it to the predict-api
-        # data lake, but the buffered request metadata it needs is local
-        # to this agent's shared_state and goes away at the pop.
-        #
-        # Gate the whole event build on ``use_offchain``. When the flag
-        # is off, no payload is built and nothing rides Tendermint
-        # consensus replication — so the ingress-side consensus cost
-        # matches the egress-side HTTP write cost (nothing on either
-        # side). The paired gate in
-        # :py:meth:`task_submission_abci.behaviours.PostTxSettlementBehaviour._do_predict_api_write_best_effort`
-        # additionally warns if it receives events under a locally-off
-        # flag, so a config drift between the two skill copies of
-        # ``use_offchain`` is alertable rather than silent.
-        # Build the event for both off-chain HTTP deliveries
-        # (``is_offchain=True``) and on-chain marketplace deliveries
-        # (``is_offchain=False`` AND ``is_marketplace_delivery=True``).
-        # The ``source`` field on the event (set by
-        # ``_build_predict_api_event``) disambiguates the two paths on
-        # the predict-api side: ``mech_offchain`` for the paid HTTP path,
-        # ``mech_onchain`` for the on-chain rails. See
-        # ``autonolas-marketplace/docs/onchain_write_path_scope.md`` for
-        # the agreed shape.
         predict_api_mode_enabled = self.params.use_offchain
         is_marketplace_delivery = bool(done_task.get("is_marketplace_mech"))
         if predict_api_mode_enabled and (is_offchain or is_marketplace_delivery):
-            done_task["predict_api_event"] = self._build_predict_api_event(
+            event = self._build_predict_api_event(
                 done_task=done_task,
                 cid=cid,
                 executing_task=cast(Dict[str, Any], executing_task),
                 response=response,
             )
+            self.context.shared_state.setdefault(PREDICT_API_EVENTS, {})[
+                str(req_id)
+            ] = event
         # add to done tasks, in thread safe way
         with self.done_tasks_lock:
             self.done_tasks.append(done_task)
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -88,6 +88,7 @@ LAST_READ_ATTEMPT_TS = "last_read_attempt_ts"
 INFLIGHT_READ_TS = "inflight_read_ts"
 REQUEST_ID_TO_DELIVERY_RATE_INFO = "request_id_to_delivery_rate_info"
 PREDICT_API_EVENTS = "predict_api_events"
+PREDICT_API_EVENT_TTL_SECONDS = 24 * 60 * 60
 # Shared-state keys owned by the MechHttpHandler (handlers.py); mirrored here
 # because the off-chain finalize path writes the response envelope for the
 # ``/fetch_offchain_info`` polling endpoint and clears the buffered request
@@ -1785,7 +1786,7 @@ class TaskExecutionBehaviour(SimpleBehaviour):
             )
             self.context.shared_state.setdefault(PREDICT_API_EVENTS, {})[
                 str(req_id)
-            ] = event
+            ] = {"event": event, "written_at": time.time()}
         # add to done tasks, in thread safe way
         with self.done_tasks_lock:
             self.done_tasks.append(done_task)
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1784,9 +1784,23 @@ class TaskExecutionBehaviour(SimpleBehaviour):
                 executing_task=cast(Dict[str, Any], executing_task),
                 response=response,
             )
-            self.context.shared_state.setdefault(PREDICT_API_EVENTS, {})[
-                str(req_id)
-            ] = {"event": event, "written_at": time.time()}
+            events_by_id = self.context.shared_state.setdefault(PREDICT_API_EVENTS, {})
+            # TTL sweep on write bounds cache growth even if settlement
+            # is stuck and no post-settlement prune ever fires.
+            cutoff = time.time() - PREDICT_API_EVENT_TTL_SECONDS
+            stale = [
+                rid
+                for rid, entry in events_by_id.items()
+                if not isinstance(entry, dict)
+                or not isinstance(entry.get("written_at"), (int, float))
+                or float(entry["written_at"]) < cutoff
+            ]
+            for rid in stale:
+                events_by_id.pop(rid, None)
+            events_by_id[str(req_id)] = {
+                "event": event,
+                "written_at": time.time(),
+            }
         # add to done tasks, in thread safe way
         with self.done_tasks_lock:
             self.done_tasks.append(done_task)
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -67,6 +67,7 @@ from packages.valory.skills.abstract_round_abci.behaviours import (
     BaseBehaviour,
 )
 from packages.valory.skills.abstract_round_abci.io_.store import SupportedFiletype
+from packages.valory.skills.task_execution.behaviours import PREDICT_API_EVENTS
 from packages.valory.skills.task_execution.utils.ipfs import to_multihash
 from packages.valory.skills.task_submission_abci.models import Params
 from packages.valory.skills.task_submission_abci.payloads import (
@@ -278,6 +279,10 @@ class TaskExecutionBaseBehaviour(BaseBehaviour, ABC):
                 for task in snapshot
                 if str(task.get("request_id")) not in submitted_id_set
             ]
+        events_by_id = self.context.shared_state.get(PREDICT_API_EVENTS)
+        if isinstance(events_by_id, dict):
+            for rid in submitted_id_set:
+                events_by_id.pop(rid, None)
         return snapshot
 
     @property
@@ -1990,21 +1995,11 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
             ``return`` anything.
         """
         if not self.params.use_offchain:
-            # Divergence safety net. ``use_offchain`` is a duplicated
-            # param (declared on both this skill and ``task_execution``);
-            # if an operator flips ``task_execution.use_offchain=true`` but
-            # forgets this one, the ingress side builds
-            # ``predict_api_event`` entries and rides Tendermint
-            # consensus replication for them — and then the egress side
-            # here silently drops the entire batch. Warn loudly when we
-            # can see that has happened so the divergence is alertable
-            # rather than invisible.
-            if any(
-                isinstance(t, dict) and t.get("predict_api_event")
-                for t in cast(List[Dict[str, Any]], self.synchronized_data.done_tasks)
-            ):
+            # Divergence safety net: the ingress side wrote events into
+            # local shared_state but the egress side has the flag off.
+            if self.context.shared_state.get(PREDICT_API_EVENTS):
                 self.context.logger.warning(
-                    "predict_api_event entries present in done_tasks but "
+                    "predict_api_event entries present in shared_state but "
                     "task_submission_abci.use_offchain is False — the "
                     "ingress-side (task_execution) and egress-side "
                     "(task_submission_abci) use_offchain flags are "
@@ -2447,30 +2442,29 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
         ).inc()
 
     def _extract_offchain_events(self) -> List[Dict[str, Any]]:
-        """Return the ``predict_api_event`` payload from every done_task that carries one.
+        """Return the ``predict_api_event`` payload for every done_task that has one.
 
-        Both off-chain HTTP deliveries (``source = 'mech_offchain'``) and
-        on-chain marketplace deliveries (``source = 'mech_onchain'``) now
-        ride the same predict-api write path; the per-event ``source`` field
-        set inside :py:meth:`task_execution.behaviours._build_predict_api_event`
-        is what disambiguates the two on the predict-api side. The filter
-        here keys on the presence of a ``predict_api_event`` field, which
-        :py:meth:`task_execution.behaviours._finalize_done_task` sets only
-        when ``use_offchain`` is on AND the task is either an off-chain
-        HTTP delivery (``is_offchain=True``) or an on-chain marketplace
-        delivery (``is_marketplace_delivery=True``). On-chain
-        non-marketplace tasks on a legacy mech contract stay out of the
-        lake by construction.
+        Looks each event up in agent-local
+        ``shared_state[PREDICT_API_EVENTS]`` keyed by ``str(request_id)``.
+        Only the agent that executed a task locally has its event; a
+        task another agent executed is skipped here and posted by that
+        agent instead. Predict-api dedups on ``request_id`` so a single
+        POST per task is sufficient.
 
         :return: an ordered list of ``MechEvent``-shaped dicts.
         """
         done_tasks = cast(List[Dict[str, Any]], self.synchronized_data.done_tasks)
+        events_by_id: Dict[str, Dict[str, Any]] = self.context.shared_state.get(
+            PREDICT_API_EVENTS, {}
+        )
         events: List[Dict[str, Any]] = []
         for task in done_tasks:
-            event = task.get("predict_api_event")
-            if not isinstance(event, dict):
+            req_id = task.get("request_id")
+            if req_id is None:
                 continue
-            events.append(event)
+            event = events_by_id.get(str(req_id))
+            if isinstance(event, dict):
+                events.append(event)
         return events
 
     def _sweep_pending_undelivered(
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -67,7 +67,10 @@ from packages.valory.skills.abstract_round_abci.behaviours import (
     BaseBehaviour,
 )
 from packages.valory.skills.abstract_round_abci.io_.store import SupportedFiletype
-from packages.valory.skills.task_execution.behaviours import PREDICT_API_EVENTS
+from packages.valory.skills.task_execution.behaviours import (
+    PREDICT_API_EVENTS,
+    PREDICT_API_EVENT_TTL_SECONDS,
+)
 from packages.valory.skills.task_execution.utils.ipfs import to_multihash
 from packages.valory.skills.task_submission_abci.models import Params
 from packages.valory.skills.task_submission_abci.payloads import (
@@ -282,6 +285,15 @@ class TaskExecutionBaseBehaviour(BaseBehaviour, ABC):
         events_by_id = self.context.shared_state.get(PREDICT_API_EVENTS)
         if isinstance(events_by_id, dict):
             for rid in submitted_id_set:
+                events_by_id.pop(rid, None)
+            cutoff = time.time() - PREDICT_API_EVENT_TTL_SECONDS
+            stale = [
+                rid
+                for rid, entry in events_by_id.items()
+                if not isinstance(entry, dict)
+                or float(entry.get("written_at", 0)) < cutoff
+            ]
+            for rid in stale:
                 events_by_id.pop(rid, None)
         return snapshot
 
@@ -2462,7 +2474,10 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
             req_id = task.get("request_id")
             if req_id is None:
                 continue
-            event = events_by_id.get(str(req_id))
+            entry = events_by_id.get(str(req_id))
+            if not isinstance(entry, dict):
+                continue
+            event = entry.get("event")
             if isinstance(event, dict):
                 events.append(event)
         return events
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -67,10 +67,7 @@ from packages.valory.skills.abstract_round_abci.behaviours import (
     BaseBehaviour,
 )
 from packages.valory.skills.abstract_round_abci.io_.store import SupportedFiletype
-from packages.valory.skills.task_execution.behaviours import (
-    PREDICT_API_EVENTS,
-    PREDICT_API_EVENT_TTL_SECONDS,
-)
+from packages.valory.skills.task_execution.behaviours import PREDICT_API_EVENTS
 from packages.valory.skills.task_execution.utils.ipfs import to_multihash
 from packages.valory.skills.task_submission_abci.models import Params
 from packages.valory.skills.task_submission_abci.payloads import (
@@ -282,19 +279,6 @@ class TaskExecutionBaseBehaviour(BaseBehaviour, ABC):
                 for task in snapshot
                 if str(task.get("request_id")) not in submitted_id_set
             ]
-        events_by_id = self.context.shared_state.get(PREDICT_API_EVENTS)
-        if isinstance(events_by_id, dict):
-            for rid in submitted_id_set:
-                events_by_id.pop(rid, None)
-            cutoff = time.time() - PREDICT_API_EVENT_TTL_SECONDS
-            stale = [
-                rid
-                for rid, entry in events_by_id.items()
-                if not isinstance(entry, dict)
-                or float(entry.get("written_at", 0)) < cutoff
-            ]
-            for rid in stale:
-                events_by_id.pop(rid, None)
         return snapshot
 
     @property
@@ -391,6 +375,13 @@ class TaskPoolingBehaviour(TaskExecutionBaseBehaviour, ABC):
         # ``shared_state[DONE_TASKS]`` under the same lock) and keeps
         # both paths on the single tested prune surface.
         shared_snapshot = self.remove_tasks_by_id(submitted_ids)
+        # Prune must fire only after ``PostTxSettlementBehaviour`` has
+        # POSTed. Do not move into ``remove_tasks_by_id``: its
+        # pre-settlement caller strands the row otherwise.
+        events_by_id = self.context.shared_state.get(PREDICT_API_EVENTS)
+        if isinstance(events_by_id, dict):
+            for rid in submitted_ids:
+                events_by_id.pop(rid, None)
         self.context.logger.info(
             f"Pruned tasks with ids {submitted_ids} from shared state; "
             f"tx_hash={tx_hash}."
@@ -1951,14 +1942,15 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
     agent-local ``shared_state[PREDICT_API_EVENTS]`` (keyed by
     ``str(request_id)``, populated on the offchain path inside
     :py:meth:`task_execution.behaviours._build_predict_api_event`).
-    Each cache entry is a ``{event, written_at}`` wrapper so the sibling
-    TTL sweep in :py:meth:`TaskExecutionBaseBehaviour.remove_tasks_by_id`
-    can drop stale entries. Only the agent that executed a task
-    locally has its event; other agents skip and rely on the executor
-    to POST. The predict-api server is idempotent on ``request_id``,
-    so a single POST per task is sufficient — each agent posts its own
-    events under its own EOA signature; the per-EOA rate limiter on the
-    server keeps the co-owners in separate budgets.
+    Each cache entry is a ``{event, written_at}`` wrapper so the TTL
+    sweep on write in
+    :py:meth:`task_execution.behaviours.TaskExecutionBehaviour._finalize_done_task`
+    can drop stale entries. Only the agent that executed a task locally
+    has its event; other agents skip and rely on the executor to POST.
+    The predict-api server is idempotent on ``request_id``, so a single
+    POST per task is sufficient — each agent posts only the subset it
+    executed, under its own EOA signature; the per-EOA rate limiter on
+    the server keeps the co-owners in separate budgets.
 
     The behaviour is fail-soft by construction. A predict-api outage / 5xx
     / network drop does NOT stop the FSM — the settlement already landed
@@ -2462,14 +2454,7 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
         ).inc()
 
     def _extract_offchain_events(self) -> List[Dict[str, Any]]:
-        """Return the ``predict_api_event`` payload for every done_task that has one.
-
-        Looks each event up in agent-local
-        ``shared_state[PREDICT_API_EVENTS]`` keyed by ``str(request_id)``.
-        Only the agent that executed a task locally has its event; a
-        task another agent executed is skipped here and posted by that
-        agent instead. Predict-api dedups on ``request_id`` so a single
-        POST per task is sufficient.
+        """Return every locally-cached ``predict_api_event`` for the round's done_tasks.
 
         :return: an ordered list of ``MechEvent``-shaped dicts.
         """
@@ -2477,6 +2462,7 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
         events_by_id: Dict[str, Dict[str, Any]] = self.context.shared_state.get(
             PREDICT_API_EVENTS, {}
         )
+        agent_address = self.context.agent_address
         events: List[Dict[str, Any]] = []
         for task in done_tasks:
             req_id = task.get("request_id")
@@ -2486,10 +2472,17 @@ class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
             event = entry.get("event") if isinstance(entry, dict) else None
             if isinstance(event, dict):
                 events.append(event)
+                continue
+            if task.get("task_executor_address") == agent_address:
+                self.context.logger.warning(
+                    "no local predict_api_event for request_id=%s but this "
+                    "agent executed it — event dropped before POST",
+                    req_id,
+                )
             else:
                 self.context.logger.debug(
                     "no local predict_api_event for request_id=%s; "
-                    "assuming executed by another agent",
+                    "executed by another agent",
                     req_id,
                 )
         return events
--- a/packages/valory/skills/task_submission_abci/behaviours.py
+++ b/packages/valory/skills/task_submission_abci/behaviours.py
@@ -1946,15 +1946,19 @@ class TransactionPreparationBehaviour(
 class PostTxSettlementBehaviour(TaskExecutionBaseBehaviour):
     """Runs once on-chain settlement of the round's tx confirms.
 
-    Reads ``synchronized_data.done_tasks`` for the just-settled cycle,
-    extracts every entry's pre-built ``predict_api_event`` payload (set on
-    the offchain path inside :py:meth:`task_execution.behaviours._build_predict_api_event`),
-    bundles them into one EIP-712-signed batch, and POSTs the batch to
-    the predict-api data lake at ``params.predict_api_events_url``. The predict-api
-    server is idempotent on ``request_id``, so multi-agent services don't
-    need a keeper pattern: each agent posts its own copy under its own
-    EOA signature; the per-EOA rate limiter on the server keeps the
-    co-owners in separate budgets.
+    Iterates ``synchronized_data.done_tasks`` for the just-settled cycle
+    and, for each entry, looks up its ``predict_api_event`` in
+    agent-local ``shared_state[PREDICT_API_EVENTS]`` (keyed by
+    ``str(request_id)``, populated on the offchain path inside
+    :py:meth:`task_execution.behaviours._build_predict_api_event`).
+    Each cache entry is a ``{event, written_at}`` wrapper so the sibling
+    TTL sweep in :py:meth:`TaskExecutionBaseBehaviour.remove_tasks_by_id`
+    can drop stale entries. Only the agent that executed a task
+    locally has its event; other agents skip and rely on the executor
+    to POST. The predict-api server is idempotent on ``request_id``,
+    so a single POST per task is sufficient — each agent posts its own
+    events under its own EOA signature; the per-EOA rate limiter on the
+    server keeps the co-owners in separate budgets.
 
     The behaviour is fail-soft by construction. A predict-api outage / 5xx
     / network drop does NOT stop the FSM — the settlement already landed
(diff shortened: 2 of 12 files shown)


DIvyaNautiyal07:
Superseded by upcoming PR — switching to kv-store persistence so events survive restarts (shared_state design lost them). Deleting this branch and starting fresh.

DIvyaNautiyal07:
Reopening — going with the simpler shared_state approach after all. Restart-loss risk is acknowledged; primary bug (RegistrationPayload > 1 MiB crash-loop) is what this fixes and the analytics gap on restart is a smaller price than the kv-store integration complexity.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -278,6 +282,19 @@ def remove_tasks_by_id(self, submitted_ids: List[str]) -> List[Dict[str, Any]]:
                 for task in snapshot
                 if str(task.get("request_id")) not in submitted_id_set
             ]
+        events_by_id = self.context.shared_state.get(PREDICT_API_EVENTS)
```
jmoreira-valory: 🟡 **F1 · MEDIUM — Event-cache prune inside remove_tasks_by_id also fires on the simulation-skip path, before the POST**

`remove_tasks_by_id` has two call sites with opposite semantics, and the new prune block treats them the same:

1. `handle_submitted_tasks` (line 393) - **post**-settlement, the events for those ids have already been POSTed by `PostTxSettlementBehaviour`. Pruning here is correct.
2. `remove_tasks([task])` (line 1500, `TransactionPreparationBehaviour.get_payload_content`) - **pre**-settlement, fired when a deliver-tx simulation fails and the task is skipped.

On path 2 the entry is popped from `shared_state[PREDICT_API_EVENTS]` while the task is *still present in* `synchronized_data.done_tasks` (consensus data is not mutated by `remove_tasks`). Later in the same period `_extract_offchain_events` iterates that same consensus list, looks the id up, misses, and drops the event with no log. The task is also gone from `shared_state[DONE_TASKS]`, so it is never re-pooled and the row is lost permanently.

Pre-PR this did not happen: the event rode inside the consensus `done_task`, so the skipped task's event was still extracted and POSTed. Reachable on legacy (non-marketplace) mechs with `use_offchain=true`, since `remaining_tasks` at line 1481 filters only on `not is_marketplace_mech` and therefore includes `is_offchain=True` tasks whose HTTP response was already delivered to the requester - exactly the rows the data lake is owed.

**Fix:** move the `PREDICT_API_EVENTS` prune (lines 285-296) out of `remove_tasks_by_id` and into `handle_submitted_tasks`, right after the `remove_tasks_by_id(submitted_ids)` call, so only the post-settlement prune touches the cache. Entries stranded by the simulation-skip path are then reclaimed by the TTL sweep, which is what it is for.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1773,40 +1775,18 @@ def _finalize_done_task(
...
         is_marketplace_delivery = bool(done_task.get("is_marketplace_mech"))
         if predict_api_mode_enabled and (is_offchain or is_marketplace_delivery):
-            done_task["predict_api_event"] = self._build_predict_api_event(
+            event = self._build_predict_api_event(
                 done_task=done_task,
                 cid=cid,
                 executing_task=cast(Dict[str, Any], executing_task),
                 response=response,
             )
+            self.context.shared_state.setdefault(PREDICT_API_EVENTS, {})[
```
jmoreira-valory: 🔵 **F2 · LOW — Durability regression is wider than the PR body claims - heap-only cache plus single-poster**

The PR body says the failure mode is "unchanged ... same window that already existed pre-PR between local completion and consensus". That understates it in two ways:

- **Window length.** Pre-PR the event became durable at `TaskPoolingRound` quorum (seconds after finalize) and was rebuilt from the Tendermint blockchain on replay, so it survived an agent restart. Post-PR it lives only in this process's heap until *this* agent reaches `PostTxSettlementBehaviour` - a full settlement cycle later (tx prep, Safe signing, on-chain confirmation). Any restart in that window loses the row. On `single_mech_mm_predict_2359`, which restarts every ~2 minutes, that is the common case rather than the tail.
- **Redundancy.** Pre-PR all four agents held the event and all four POSTed; one agent dying cost nothing. Post-PR only the executor holds it, so its death is total loss for that request_id.

The trade-off is defensible (the subsystem is explicitly fail-soft and the kv_store replay buffer is a known follow-up), but the claim in the description should be corrected so the operator rollout note is accurate. Also worth adding a `debug` log or a counter on the cache miss in `_extract_offchain_events` (line 2478) - today a miss from "another agent executed it" is indistinguishable from a miss from "we restarted and lost the cache", which makes the loss unmeasurable.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -278,6 +282,19 @@ def remove_tasks_by_id(self, submitted_ids: List[str]) -> List[Dict[str, Any]]:
...
+        if isinstance(events_by_id, dict):
+            for rid in submitted_id_set:
+                events_by_id.pop(rid, None)
+            cutoff = time.time() - PREDICT_API_EVENT_TTL_SECONDS
```
jmoreira-valory: 🔵 **F3 · LOW — TTL safety net cannot fire in the scenario its own test docstring describes**

The TTL sweep only runs *inside* `remove_tasks_by_id`, i.e. it requires a primary prune to have been triggered by some other task. Its test docstring says it is a "safety net for tasks that finalize locally but never route through a primary prune (e.g. FSM branch skips remove_tasks*)" - but in exactly that situation no prune ever runs, so nothing is ever swept.

Concretely, `handle_submitted_tasks` returns early when `check_last_tx_status()` is falsy (`FailedRound` path, or no `final_tx_hash`) or when `submitted_request_ids` is empty. An agent whose settlements keep failing accumulates 30-60 KB entries indefinitely - the same unbounded-growth shape this PR is trying to remove from consensus, just relocated to the heap.

**Fix:** run the TTL sweep somewhere that executes every cycle regardless of whether a prune happened - the top of `_do_predict_api_write_best_effort`, or on write in `_finalize_done_task`. The dict is small, so an O(n) scan per cycle is free.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -278,6 +282,19 @@ def remove_tasks_by_id(self, submitted_ids: List[str]) -> List[Dict[str, Any]]:
...
+            stale = [
+                rid
+                for rid, entry in events_by_id.items()
+                if not isinstance(entry, dict)
+                or float(entry.get("written_at", 0)) < cutoff
```
jmoreira-valory: 🔵 **F4 · LOW — Unguarded float() on written_at can raise on the FSM-critical prune path**

`float(entry.get("written_at", 0))` raises `ValueError` on a non-numeric string and `TypeError` on a list/dict. `remove_tasks_by_id` is called from `TaskPoolingBehaviour.handle_submitted_tasks`, so the exception escapes into `async_act` and wedges the period rather than degrading.

The surrounding code already defends against a malformed entry shape (`not isinstance(entry, dict)`, and the new test seeds a `"not-a-dict"` value), so the missing numeric guard is an inconsistency rather than a deliberate choice. Treat a bad `written_at` the same way as a bad entry - as stale, and drop it:

```suggestion
                or not isinstance(entry.get("written_at"), (int, float))
                or float(entry["written_at"]) < cutoff
```

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -87,6 +87,8 @@
 LAST_READ_ATTEMPT_TS = "last_read_attempt_ts"
 INFLIGHT_READ_TS = "inflight_read_ts"
 REQUEST_ID_TO_DELIVERY_RATE_INFO = "request_id_to_delivery_rate_info"
+PREDICT_API_EVENTS = "predict_api_events"
```
jmoreira-valory, line 90: ⚪ **F6 · NIT — PREDICT_API_EVENTS now has two definitions and one cross-skill import**

`PREDICT_API_EVENTS` is declared at `task_execution/behaviours.py:90` and again at `task_execution/handlers.py:91`, and `task_submission_abci/behaviours.py` imports the `behaviours.py` copy. Three sites for one shared-state key, two of which can drift silently - a typo in the handler copy would leave `setup()` initialising a key nobody reads, and the failure is invisible because `_extract_offchain_events` treats a miss as "another agent executed it".

The existing mirrored-constants block has a comment explaining the mirroring, so this matches local convention, but a single home (e.g. `task_execution/utils/`, already imported cross-skill via `utils.ipfs`) would be cheaper to keep honest - and would avoid importing the whole `task_execution.behaviours` module into `task_submission_abci` just for two string constants.

Separately, `test_egress_gate_off_with_diverged_ingress_emits_divergence_warning` seeds `{"r1": {"src": "off"}}`, which is not the `{"event": ..., "written_at": ...}` envelope the writer produces. It passes because the check is only a truthiness test, but the fixture encodes a shape that can never occur.

jmoreira-valory (changes requested):

**What this does.** The mech's per-task predict-api analytics payload (30-60 KB of raw tool output per task) used to be stuffed into the `done_task` dict, which rides Tendermint consensus replication. A batch of 15-20 tasks pushed one period's `done_tasks` past the 1 MiB `MAX_READ_IN_BYTES` limit and crash-looped the service. This PR keeps the payload in an agent-local `shared_state[PREDICT_API_EVENTS]` dict keyed by request_id instead, so consensus only carries the small fields, and the post-settlement behaviour looks each event up locally rather than reading it off the replicated `done_tasks`.

The core idea is right and the FSM ordering checks out: `TaskPooling -> TransactionSubmission -> PostTxSettlement -> ResetAndPause -> TaskPooling` (composition.py:51-54), so the POST in `PostTxSettlementBehaviour` always runs before the next period's prune in `handle_submitted_tasks`. Key normalisation is consistent too - the writer keys on `str(executing_task["requestId"])`, `done_task["request_id"]` is the same value (the `**executing_task` spread cannot clobber it, the source key is `requestId`), and `extract_request_ids` stringifies, so the prune pops the keys the writer created. `TaskPoolingRound` unions every participant's payloads rather than most-voting one list, so the "each agent posts only what it executed" split does cover the whole batch. CI is green (22/22) and the test delta tracks the behaviour change closely.

One blocker: `remove_tasks_by_id` is shared by a post-settlement caller and a pre-settlement one, and the new prune block does not distinguish them (F1). The rest are non-blocking.

Not re-raised, but worth noting for the rollout: this is the second size-reduction pass on `done_tasks` after #482, and the canary on the crash-looping service is still the only end-to-end evidence that 1 MiB is now actually bounded - the per-task residual is smaller but still unbounded in count.

**Findings** — 6 · 🟡 1 Medium · 🔵 3 Low · ⚪ 2 Nit

| F1 | 🟡 MEDIUM | [packages/valory/skills/task_submission_abci/behaviours.py:285](https://github.com/valory-xyz/mech/blob/256cf91e9e7653a76a7aa03bb3206f464b92720b/packages/valory/skills/task_submission_abci/behaviours.py#L285) | **Event-cache prune inside remove_tasks_by_id also fires on the simulation-skip path, before the POST** |
| F2 | 🔵 LOW | [packages/valory/skills/task_execution/behaviours.py:1787](https://github.com/valory-xyz/mech/blob/256cf91e9e7653a76a7aa03bb3206f464b92720b/packages/valory/skills/task_execution/behaviours.py#L1787) | **Durability regression is wider than the PR body claims - heap-only cache plus single-poster** |
| F3 | 🔵 LOW | [packages/valory/skills/task_submission_abci/behaviours.py:289](https://github.com/valory-xyz/mech/blob/256cf91e9e7653a76a7aa03bb3206f464b92720b/packages/valory/skills/task_submission_abci/behaviours.py#L289) | **TTL safety net cannot fire in the scenario its own test docstring describes** |
| F4 | 🔵 LOW | [packages/valory/skills/task_submission_abci/behaviours.py:294](https://github.com/valory-xyz/mech/blob/256cf91e9e7653a76a7aa03bb3206f464b92720b/packages/valory/skills/task_submission_abci/behaviours.py#L294) | **Unguarded float() on written_at can raise on the FSM-critical prune path** |
| F5 | ⚪ NIT | `packages/valory/skills/task_submission_abci/behaviours.py:1949` | **PostTxSettlementBehaviour docstrings still describe the old done_tasks-embedded event** |
| F6 | ⚪ NIT | [packages/valory/skills/task_execution/behaviours.py:90](https://github.com/valory-xyz/mech/blob/256cf91e9e7653a76a7aa03bb3206f464b92720b/packages/valory/skills/task_execution/behaviours.py#L90) | **PREDICT_API_EVENTS now has two definitions and one cross-skill import** |

---

- ⚪ **F5 · NIT — PostTxSettlementBehaviour docstrings still describe the old done_tasks-embedded event** Two docstrings survived the move and now describe behaviour that no longer exists:

- The class docstring (line 1949) says it "Reads `synchronized_data.done_tasks` for the just-settled cycle, extracts every entry's pre-built `predict_api_event` payload". It now reads the ids from `done_tasks` and the payloads from `shared_state[PREDICT_API_EVENTS]`.
- `_do_predict_api_write_best_effort` short-circuit 3 (line 2001) says "The round's done_tasks carry no offchain predict_api_event entries".

The class docstring also still asserts "each agent posts its own copy under its own EOA signature", which is now only true for the subset each agent executed - worth stating explicitly, since it is the load-bearing multi-agent change in this PR.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -278,6 +282,19 @@ def remove_tasks_by_id(self, submitted_ids: List[str]) -> List[Dict[str, Any]]:
                 for task in snapshot
                 if str(task.get("request_id")) not in submitted_id_set
             ]
+        events_by_id = self.context.shared_state.get(PREDICT_API_EVENTS)
+        if isinstance(events_by_id, dict):
+            for rid in submitted_id_set:
```
OjusWiZard: **Behaviour change worth a decision: this can delete an event before it is ever POSTed.**

`remove_tasks_by_id` has a second live caller besides `handle_submitted_tasks` — `TransactionPreparationBehaviour` calls `remove_tasks([task])` when a deliver-tx simulation fails. `TransactionPreparationRound` runs *before* `PostTxSettlementRound` in the **same period**, so this pop lands ahead of the read.

- Pre-PR: `_extract_offchain_events` read `task["predict_api_event"]` off `synchronized_data.done_tasks` — the consensus copy, which `remove_tasks` never touched (it only mutated `shared_state[DONE_TASKS]`). The event **was** posted.
- Post-PR: the event lives in the cache this line prunes. It is **not** posted.

Reachability is narrower than it first looks: `remaining_tasks` excludes marketplace mechs, and for a non-marketplace task an event only exists when `is_offchain=True` — so it needs legacy-mech + off-chain + sim failure + `use_offchain` on. That's the only other live `remove_tasks` call site; the marketplace and off-chain deliver paths handle sim failure without it.

I think the new behaviour is arguably the *better* one — no delivered-event for a task that was never delivered, and `_sweep_pending_undelivered` can still emit a request-only event later. But it's an undocumented semantic change, it isn't in the Changes list, and no test covers it. Either state it in the description as intended, or give the simulation-failure branch a variant that prunes `DONE_TASKS` only.

packages/valory/skills/task_submission_abci/behaviours.py
OjusWiZard: Two small things.

**`float()` is unguarded while the entry beside it is.** `isinstance(entry, dict)` protects the shape, but a non-numeric `written_at` still raises `ValueError` out of `remove_tasks_by_id` -> `handle_submitted_tasks` -> `TaskPooling.async_act`, which has no `try/except` around it. Not reachable today (only `_finalize_done_task` writes it, always `time.time()`), but the failure mode is nasty: one bad write and *every subsequent period's* prune raises.

**Three passes could be two.** A submitted id is evicted unconditionally, so that condition folds into the same filter:

```python
cutoff = time.time() - PREDICT_API_EVENT_TTL_SECONDS
to_evict = [
    rid
    for rid, entry in events_by_id.items()
    if rid in submitted_id_set
    or not isinstance(entry, dict)
    or float(entry.get("written_at", 0)) < cutoff
]
for rid in to_evict:
    events_by_id.pop(rid, None)
```

Behaviour-identical, keeps the list-then-pop shape, and `to_evict` is an honest name now that the list isn't only stale entries. Keep the `isinstance(events_by_id, dict)` guard — it's load-bearing for tests that don't run `ContractHandler.setup`.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2447,30 +2462,36 @@ def _post_predict_api_batch(  # noqa: C901
...
                 continue
-            events.append(event)
+            entry = events_by_id.get(str(req_id))
+            event = entry.get("event") if isinstance(entry, dict) else None
+            if isinstance(event, dict):
+                events.append(event)
+            else:
+                self.context.logger.debug(
+                    "no local predict_api_event for request_id=%s; "
+                    "assuming executed by another agent",
```
OjusWiZard: Good addition — but as written this log *assumes* rather than *checks*, and the code has what it needs to actually check.

`done_task` carries `task_executor_address`. A task another agent executed legitimately has no local entry; a task **this** agent executed must always have one. So the two cases are distinguishable:

```python
if task.get("task_executor_address") == self.context.agent_address:
    self.context.logger.warning(...)   # always a bug
else:
    self.context.logger.debug(...)     # normal, another agent posts it
```

As it stands, in a 4-agent service roughly 3 of every 4 done_tasks emit this line during normal operation, so a genuine ingress-side drop is buried in expected noise at DEBUG. `test_extract_offchain_events_logs_debug_when_local_miss` says the log *"separates the 'executed by another agent' case (normal) from a silent drop caused by an ingress-side bug"* — it doesn't yet; both paths log identically. The executor-address check is what would make that docstring true, and it's the check that would have surfaced the `remove_tasks` interaction I flagged above.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2447,30 +2462,36 @@ def _post_predict_api_batch(  # noqa: C901
...
-        set inside :py:meth:`task_execution.behaviours._build_predict_api_event`
-        is what disambiguates the two on the predict-api side. The filter
-        here keys on the presence of a ``predict_api_event`` field, which
-        :py:meth:`task_execution.behaviours._finalize_done_task` sets only
-        when ``use_offchain`` is on AND the task is either an off-chain
-        HTTP delivery (``is_offchain=True``) or an on-chain marketplace
-        delivery (``is_marketplace_delivery=True``). On-chain
-        non-marketplace tasks on a legacy mech contract stay out of the
-        lake by construction.
+        """Return the ``predict_api_event`` payload for every done_task that has one.
```
OjusWiZard: Per CLAUDE.md this wants to be a one-line summary. The multi-agent dedup rationale — only the executing agent holds the event, predict-api dedups on `request_id` — is exactly the "why" that belongs in the PR description, where it already is, and `git blame` -> this commit -> this PR reaches it. Keep the summary line and the `:return:`.

packages/valory/skills/task_submission_abci/tests/test_onchain_write_path.py
```diff
@@ -500,49 +501,74 @@ def test_request_only_event_falls_back_to_now_on_bad_timestamp() -> None:
...
     self_ = cast(
         PostTxSettlementBehaviour,
         SimpleNamespace(
             synchronized_data=SimpleNamespace(
                 done_tasks=[
-                    {"is_offchain": True, "predict_api_event": {"src": "off"}},
-                    {"is_offchain": False, "predict_api_event": {"src": "on"}},
-                    {"is_offchain": False, "predict_api_event": None},  # skipped
-                    {"is_offchain": True, "predict_api_event": "not-a-dict"},  # skipped
+                    {"request_id": "r-off"},
```
OjusWiZard, line 512: **This test can't fail the way production would break.** Both sides use literal `str` keys, so `events_by_id.get(req_id)` with the `str()` dropped would still pass — `str("r-off") == "r-off"`.

In production `request_id` is an **`int`**. Your own comment at `task_execution/behaviours.py:1797` says so ("`req_id` is `int` post ingress coercion"), and it's there because this exact str/int key-mismatch class already bit the `IN_MEMORY_REQUESTS` pop. `TestRemoveTasksById` has `test_int_shared_state_matches_str_id` guarding the same class on `DONE_TASKS`; the analogous one for `PREDICT_API_EVENTS` is still missing.

One case with `{"request_id": 42}` in `done_tasks` and `"42"` in the cache would pin the coercion that's actually load-bearing. Cheap, and it's the one failure mode in this hand-off that fails completely silently.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -87,6 +87,8 @@
...
+PREDICT_API_EVENT_TTL_SECONDS = 24 * 60 * 60
```
OjusWiZard, line 91: `256cf91e` explains *why there is a TTL* but not *why 24h*. Per CLAUDE.md a magic constant whose derivation lives nowhere earns a one-liner — but only if the derivation genuinely isn't recoverable. If 24h was picked to comfortably exceed the worst observed settlement latency, that sentence in the commit message or the PR body is the better home and no comment is needed here.

OjusWiZard (dismissed):
Approving. The design is right, the goal is achieved — the ~30-60 KB `raw_content` no longer rides Tendermint replication — and `d70562d2` already closed three of the things I was going to raise (the pop-loop test gap, the stale divergence-test docstring, and the duplicated `PREDICT_API_EVENTS` constant). Reviewed at head `d70562d2`; **831 tests pass** locally.

**Verified**

- **Multi-agent POST is correct.** `PostTxSettlementBehaviour` is the matching-round behaviour with no keeper gate, so every agent runs `_extract_offchain_events` and posts what it executed locally. Non-local tasks are skipped, not dropped.
- **Prune-before-POST ordering is safe on the main path.** `handle_submitted_tasks` prunes the *previous* period's `submitted_request_ids` at `TaskPoolingRound` — after that period's `PostTxSettlementRound` already posted.
- **Key discipline holds.** `requestId` is coerced to `int` on both ingress paths and survives the consensus JSON round-trip; `done_task["request_id"]` is copied from the same value, and the trailing `**executing_task` splat is camelCase so it cannot clobber it. Writer and reader stringify the same value.
- **No import cycle** from the new `handlers.py` -> `behaviours.py` import: `behaviours.py` never imports `handlers.py`, so the dependency stays one-directional.
- `task_submission_abci/skill.yaml` already declared the `task_execution` dependency and its hash matches the new `packages.json` entry. The `autonomy packages lock --check` failures on my machine are all third-party (stale local registry), none of this PR's own packages.

**One thing that looks alarming and isn't.** The unlocked `PREDICT_API_EVENTS` write/sweep is **not** a live race. The tool runs in a pebble `ProcessPool` — a separate *process* that cannot touch parent `shared_state` — and its future is polled from `async_act` with no `add_done_callback`. The only `ThreadPoolExecutor` is the RPC executor in `handlers.py`, whose callbacks only log. Everything here is on the single AEA agent-loop thread; `done_tasks_lock` predates the current executor. Keeping both dicts under one lock would be nice for consistency, but there is no `RuntimeError: dictionary changed size during iteration` to fix, and I'd rather not have that recorded as a bug that was "fixed".

**Nothing blocking.** One behaviour change I'd like a decision on (inline on `remove_tasks_by_id`), one test gap that matters given this repo's history (inline), and some smaller notes.

Two things for the description itself:

1. **"Failure mode is unchanged ... same window that already existed pre-PR" undercounts it.** Pre-PR the event entered consensus one round after local completion and *every* agent held a copy. Post-PR only the executor holds it, from local completion through the entire settlement pipeline (randomness -> select-keeper -> collect-signature -> finalize -> validate -> post-settlement). The window is materially longer and has no cross-agent redundancy. That is the deliberate trade in the "Multi-agent tradeoff" section — it just isn't the *same* window, and I'd rather it said so plainly for whoever reads this during an incident.
2. Minor leftover: `test_egress_gate_off_with_no_events_stays_silent` still says *"empty `done_tasks`"* in its docstring, but the gate it guards now reads `shared_state[PREDICT_API_EVENTS]`. Outside the diff so I couldn't pin it inline.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Confirmed and fixed in f0430069 / ea268e12. Moved the PREDICT_API_EVENTS prune out of remove_tasks_by_id and into handle_submitted_tasks so only the post-settlement caller touches the cache. The pre-settlement simulation-skip path leaves the entry alone and the TTL sweep on write in _finalize_done_task reclaims it. Added test_pre_settlement_remove_tasks_preserves_predict_api_cache to pin the invariant.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Agreed on both parts. On the description accuracy, will update the PR body to spell out that only the executor holds the row from local completion through the full settlement pipeline (randomness, select-keeper, collect-signature, finalize, validate, post-settlement), rather than paraphrasing it as the pre-PR window. On observability, _extract_offchain_events now branches on task_executor_address so a self-executed miss logs WARNING (real ingress-side drop) while other-agent misses stay at DEBUG (normal in a 4-agent service). Overlaps with Ojuswi's inline point on the same method.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Confirmed. Moved the TTL sweep to _finalize_done_task so it runs on every write regardless of settlement health. Bounds cache growth even for an agent whose settlements keep failing, since new writes never bypass the sweep. test_finalize_done_task_sweeps_stale_predict_api_events_on_write covers fresh vs stale vs malformed vs bad-written-at seeds.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Confirmed. Added an isinstance(written_at, (int, float)) guard so a malformed timestamp is dropped as stale rather than raising up into TaskPooling.async_act. The guard sits at the new TTL sweep site in _finalize_done_task; the bad-written-at seed in test_finalize_done_task_sweeps_stale_predict_api_events_on_write pins it.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Already addressed in d70562d2 (round 3, landed before this review). handlers.py now imports the single definition from behaviours.py, so PREDICT_API_EVENTS lives in one place. Not moving to task_execution/utils/ this round to keep the diff focused, but noted for a follow-up if this file grows more shared-state keys.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Confirmed and fixed the same way jmoreira called out separately. Moved the prune out of remove_tasks_by_id into handle_submitted_tasks so only the post-settlement path touches the cache. Pre-settlement simulation-skip leaves the entry alone; TTL sweep on write reclaims it if the request never gets a delivered event. Pinned with test_pre_settlement_remove_tasks_preserves_predict_api_cache. Also chose your framing that it's arguably the better behaviour (no delivered-event for something that was never delivered), and updated the PR description to state the semantic change explicitly.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Guard added. isinstance(written_at, (int, float)) sits alongside the existing isinstance(entry, dict) so a bad timestamp is dropped as stale rather than raising. Covered by the bad-written-at seed in test_finalize_done_task_sweeps_stale_predict_api_events_on_write.

On the three-passes-into-two fold, it no longer applies after the round-4 moves: the submitted-id prune is now a small loop inside handle_submitted_tasks, and the TTL sweep is inside _finalize_done_task. There's no shared filter to collapse anymore. Kept the isinstance(events_by_id, dict) guard at the prune site so the test-only path (no ContractHandler.setup) still works.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Great catch, fixed. task.get('task_executor_address') == self.context.agent_address now splits the log: a self-executed miss is a real ingress-side drop and logs WARNING, an other-agent miss is normal and stays at DEBUG. task_executor_address is set on every done_task at task_execution/behaviours.py:1116 so the field is reliable. Two new tests pin the two branches: test_extract_offchain_events_logs_warning_when_self_executor_has_no_cache_entry and test_extract_offchain_events_logs_debug_when_other_agent_executed.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Trimmed to a one-line summary plus :return:. The multi-agent dedup rationale sits in the class docstring on PostTxSettlementBehaviour and in the PR description, so a reader tracing the method has one line to skim and two clear places for the why.

packages/valory/skills/task_submission_abci/tests/test_onchain_write_path.py
DIvyaNautiyal07: Confirmed and fixed. Added test_extract_offchain_events_matches_int_request_id_against_str_key which seeds {'request_id': 42} in done_tasks against a '42' cache key. A regression that drops the str(req_id) cast fails there. Same class of failure that bit IN_MEMORY_REQUESTS, worth pinning explicitly.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: 24h was chosen as a comfortable multiple of the worst observed settlement round-trip on the crash-looping mechs (usually minutes, occasionally low tens of minutes when Gnosis is congested). No hard derivation, just a safe upper bound with plenty of headroom. Adding a one-line note in the PR description instead of a code comment since the constant lives in one place and git blame lands on the right commit; happy to inline a one-liner if you'd rather have it next to the constant.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2447,30 +2454,37 @@ def _post_predict_api_batch(  # noqa: C901
...
+            req_id = task.get("request_id")
+            if req_id is None:
+                continue
+            entry = events_by_id.get(str(req_id))
+            event = entry.get("event") if isinstance(entry, dict) else None
+            if isinstance(event, dict):
+                events.append(event)
                 continue
-            events.append(event)
+            if task.get("task_executor_address") == agent_address:
```
jmoreira-valory, line 2476: 🔵 **F7 · LOW — Cache-miss WARNING misfires on legacy non-marketplace on-chain tasks**

The new self-executor branch treats every cache miss on a locally-executed task as an ingress-side drop, but there is a third, entirely legitimate case: the ingress gate in `_finalize_done_task` builds an event only when `use_offchain and (is_offchain or is_marketplace_delivery)`. A task delivered on-chain through a mech whose `MechConfig.is_marketplace_mech` is `False` (the dataclass default, `task_execution/models.py:46`) never gets an event built - the old `_extract_offchain_events` docstring called this out explicitly as "on-chain non-marketplace legacy mech tasks stay out of the lake by construction".

So on any service running `use_offchain=True` with a legacy mech configured, every locally-executed on-chain task now logs `event dropped before POST` at WARNING on every settlement. That is the exact signal the branch was added to make trustworthy, and it fires loudest on a case that is working as designed.

Both discriminators are already on the consensus `done_task` (`is_offchain` from ingress, `is_marketplace_mech` set in `_finalize_done_task` right before the append), so the check is free:

```suggestion
            if task.get("task_executor_address") == agent_address and (
                task.get("is_offchain") or task.get("is_marketplace_mech")
            ):
```

The remaining `else` DEBUG then covers both "another agent executed it" and "no event was ever built for it", which is the right level for both.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1981,30 +1994,22 @@ def _do_predict_api_write_best_effort(self) -> Generator:
...
-            # here silently drops the entire batch. Warn loudly when we
-            # can see that has happened so the divergence is alertable
-            # rather than invisible.
-            if any(
-                isinstance(t, dict) and t.get("predict_api_event")
-                for t in cast(List[Dict[str, Any]], self.synchronized_data.done_tasks)
-            ):
+            # Divergence safety net: the ingress side wrote events into
+            # local shared_state but the egress side has the flag off.
+            if self.context.shared_state.get(PREDICT_API_EVENTS):
```
jmoreira-valory, line 2010: 🔵 **F8 · LOW — Divergence warning can fire permanently on a service whose flags are not diverged**

The divergence check moved from "this round's `done_tasks` carry events" to "the cache is non-empty", and the TTL sweep moved *inside* the ingress gate in `_finalize_done_task`. Those two changes interact: once `task_execution.use_offchain` is turned off, no write ever happens again, so the sweep never runs again either. Any entry left in the cache at that moment - a task whose settlement failed `check_last_tx_status`, or one in flight when the flag was flipped - is now unreachable by both prune paths.

On a service where an operator rolls the feature back (both flags off, which is the *correct* Phase 1 state), that residue makes this branch log "the ingress-side and egress-side use_offchain flags are diverging" on every single `PostTxSettlement` round, for the life of the process, pointing at a misconfiguration that does not exist. The 24h TTL is specifically what should have bounded this, and it can't fire.

Cheapest fix is to hoist the sweep out from under the ingress gate so it runs on any `_finalize_done_task` call rather than only on event-building ones; that restores the TTL as a real backstop and the warning drains with it. Alternatively, keep the warning scoped to the round by matching against `done_tasks` request_ids the way `_extract_offchain_events` does, so a stale entry with no corresponding settled task stays quiet.

jmoreira-valory (approved):

**What this does.** The mech used to stuff each task's 30-60 KB predict-api analytics payload into the `done_task` dict, which rides Tendermint consensus replication - a batch of 15-20 tasks blew past the 1 MiB `MAX_READ_IN_BYTES` limit and crash-looped the service. This PR keeps the payload in an agent-local `shared_state[PREDICT_API_EVENTS]` cache keyed by `str(request_id)` instead, so consensus carries only the small fields and the post-settlement behaviour looks each event up locally. Round 4 relocates the cache prune to the post-settlement caller only, moves the TTL sweep onto the write path, and splits the cache-miss log into a real-drop WARNING vs a normal-other-agent DEBUG.

All six findings from round 1 are addressed at HEAD, and the two blocking mechanics re-verify cleanly:

- **F1 (MEDIUM, the blocker)** - the prune now lives only in `handle_submitted_tasks` (behaviours.py:381), and `remove_tasks_by_id` no longer touches the cache, so the pre-settlement simulation-skip caller in `TransactionPreparationBehaviour` can no longer strand a row before the POST. `test_pre_settlement_remove_tasks_preserves_predict_api_cache` pins it from the other direction.
- **Ordering re-verified independently of the PR body.** `submitted_request_ids` is written in `PostTxSettlementRound.end_block` (rounds.py:356), not in `TaskPoolingRound`, so the ids the next period prunes are by construction ids the POST already ran on. The empty-`done_tasks` branch above it (delivery-rate settlements arriving at the same round) preserves a pending hand-off rather than clobbering it, so the prune can't be skipped by a non-task settlement either.
- **Key normalisation re-verified.** `extract_request_ids` (rounds.py:44) stringifies, the writer keys on `str(req_id)`, and `done_task["request_id"]` is the same `executing_task["requestId"]` value the writer used - so `events_by_id.pop(rid, None)` hits the keys it needs to. `test_extract_offchain_events_matches_int_request_id_against_str_key` now pins the read side against the int/str mismatch that previously bit `IN_MEMORY_REQUESTS`.
- **F3/F4** - the TTL sweep moved to `_finalize_done_task`, so it runs on every write regardless of settlement health, and the `isinstance(..., (int, float))` guard means a malformed `written_at` is dropped as stale instead of raising into `TaskPooling.async_act`. Both call sites of `_finalize_done_task` (the offchain path and the IPFS store callback) are on the agent's main loop, so the sweep's iterate-then-pop has no concurrent-mutation exposure.
- **F5/F6** - class docstring and short-circuit-3 docstring now describe the shared-state lookup, and `PREDICT_API_EVENTS` has one definition with `handlers.py` importing it.
- **F2** - the durability window is now stated plainly in the PR body (executor-only, from local completion through the whole settlement pipeline, no cross-agent redundancy), with the kv_store replay buffer named as the follow-up. That was the ask.

CI is green (22/22 at `ea268e12`). The two items below are non-blocking observability polish on the newly added logging - both are false-positive alarms, neither affects the data path.

**Findings** — 2 · 🔵 2 Low

| F7 | 🔵 LOW | [packages/valory/skills/task_submission_abci/behaviours.py:2476](https://github.com/valory-xyz/mech/blob/ea268e1214785267eafbf45aa344d54cf13eeb4e/packages/valory/skills/task_submission_abci/behaviours.py#L2476) | **Cache-miss WARNING misfires on legacy non-marketplace on-chain tasks** |
| F8 | 🔵 LOW | [packages/valory/skills/task_submission_abci/behaviours.py:2010](https://github.com/valory-xyz/mech/blob/ea268e1214785267eafbf45aa344d54cf13eeb4e/packages/valory/skills/task_submission_abci/behaviours.py#L2010) | **Divergence warning can fire permanently on a service whose flags are not diverged** |

packages/valory/skills/task_submission_abci/behaviours.py
OjusWiZard, line 2476: **Non-blocking, but this warns on tasks that never had an event by construction.**

The producer only builds an event when `use_offchain AND (is_offchain OR is_marketplace_mech)`. This warns on `task_executor_address == agent_address` alone. Those conditions don't match, so a legacy on-chain **non-marketplace** task (`is_offchain=False`, `is_marketplace_mech=False`) that this agent executed takes the WARNING branch even though nothing was ever cached for it. Ran it against the real method:

```
task = {"request_id": 7, "task_executor_address": ME,
        "is_offchain": False, "is_marketplace_mech": False}

WARNINGS : ['no local predict_api_event for request_id=7 but this agent
             executed it — event dropped before POST']
```

Nothing was dropped. `_do_predict_api_write_best_effort` already returned early unless `use_offchain` is on, so on a legacy non-marketplace mech with the flag on this is one false WARNING per self-executed task per settlement round — the same boot-noise problem `test_egress_gate_off_with_no_events_stays_silent` guards against elsewhere. The docstring removed in `d70562d2` said it outright: *"On-chain non-marketplace tasks on a legacy mech contract stay out of the lake by construction."*

Mirroring the producer's condition fixes it:

```python
expected = task.get("is_offchain") or task.get("is_marketplace_mech")
if expected and task.get("task_executor_address") == agent_address:
    self.context.logger.warning(...)
else:
    self.context.logger.debug(...)
```

Both keys are already on `done_task`. Heads-up that `test_extract_offchain_events_logs_warning_when_self_executor_has_no_cache_entry` passes a task with neither key, so it currently pins this exact shape as correct — it needs one of them added alongside the fix, or it'll go green on the bug.

Two trivia not worth their own round trip: the *why 24h* derivation is still unrecorded anywhere, and `isinstance(True, int)` is `True`, so a bool `written_at` would sweep rather than raise — harmless, just noting it.

OjusWiZard (approved):
Approving. Re-reviewed at `ea268e12`; all six comments from the last round are properly addressed, and I checked each against the code rather than the commit message:

- The prune moved out of `remove_tasks_by_id` into `handle_submitted_tasks`, so the pre-settlement simulation-skip caller no longer strands the row. That's the right fix, and the comment recording the ordering constraint earns its place under CLAUDE.md's "non-obvious ordering constraint" exception.
- `float()` is now guarded by an `isinstance(..., (int, float))` check.
- `task_executor_address` splits WARNING from DEBUG.
- The `_extract_offchain_events` docstring is down to a one-liner.
- `test_extract_offchain_events_matches_int_request_id_against_str_key` pins the `str()` coercion with a real `int` — that's the one that would have failed silently in production.
- The description's "same window" claim is gone.

**834 tests pass** locally; `black` and `pyflakes` are clean on the changed files.

Moving the TTL sweep to the write path is a better shape than what I suggested — since writes keep happening while settlement is stuck, it genuinely covers the unbounded-growth case. The only residue is an idle mech (or one with `use_offchain` flipped off) never sweeping leftover orphans, which is bounded and small because the post-settlement prune still drains delivered ids. Not worth changing.

One non-blocking thing the restructure introduced, inline. Nothing here blocks merge.

PR #486, opened 2026-09-02
fix(task_execution): coerce request_ids to bytes32 for batch status check

DIvyaNautiyal07:
## Summary
`_update_pending_tasks` calls `fetch_batch_request_id_status` with a list of `int` request_ids. The callee encodes them into a `bytes32[]` ABI slot via `eth_abi`. In `eth_abi>=6`, the strict `BytesEncoder` rejects an `int` at that slot with `EncodingTypeError`, which the ledger connection turns into a `code=500` on the contract-api dialogue. Marketplace mechs with a non-empty pending queue hit this every status-check tick (~10 min).

Fix: coerce each `requestId` to a 32-byte big-endian bytes value at the caller so the encoder receives what it expects.

## Why now

The mismatch has been in the code since [c2606d1a](https://github.com/valory-xyz/mech/commit/c2606d1a) (2025-10-28), the commit that both flipped the callee's ABI from `uint256[]` to `bytes32[]` and added the caller (which kept passing ints). Older `eth_abi` (4.x) silently coerced `int -> bytes32` via `int.to_bytes(32, "big")` so the crash never fired. It became visible after `eth_abi` bumped past 5.x, which reached the mech repo via the open-autonomy 0.21.23 / open-aea 2.2.7 bump on 2026-05-29 (commit `eab5e288`) and shipped to production over the following release cycles.

## Observable symptom

Marketplace mechs periodically emit a stack trace like:

```
INFO ... Received ContractApi message: Message(sender=valory/ledger:0.19.0,
  to=valory/task_execution:0.1.0, code=500, ...)
eth_abi.exceptions.EncodingTypeError:
  Value <uint256> of type <class 'int'>
  cannot be encoded by ExactLengthBytesEncoder
```

`task_execution` catches this as `"Contract API Message performative not recognized: error"` most of the time, so the process usually survives, but the marketplace status check is silently failing on every tick. Occasionally the error cascades and the agent crashes; production has seen restart-loops on affected services.

Impact: marketplace pending-queue state tracking has been silently broken on affected services since at least mid-2026-08 (Grafana retention floor). The queue can't verify whether other agents already delivered a request, so timeout / de-dup logic that depends on this check is running blind.

## Changes
- `packages/valory/skills/task_execution/behaviours.py:1010` — convert each `requestId` to `bytes` of length 32 before passing to `fetch_batch_request_id_status`.
- `packages/valory/skills/task_execution/tests/test_behaviours.py` — new test `test_update_pending_tasks_coerces_request_ids_to_bytes32` seeding a hash-derived `uint256` id and asserting the outgoing kwargs carry `bytes` of length 32. Verified as a mutation test: reverting the fix makes the test fail with `expected bytes, got int`.

## Test plan
- [x] Full skill suite passes locally (896 tests)
- [x] Lints clean (`black`, `isort`, `flake8`, `darglint`, `mypy`)
- [x] `autonomy packages lock --check` OK
- [ ] CI: full pipeline
- [ ] Post-merge: watch affected marketplace mechs (single_mech_mm_predict_2359, mech_mm_predict_clone) for cessation of `ExactLengthBytesEncoder` events

### DIFF
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -72,6 +72,7 @@ from packages.valory.skills.task_execution.utils.ipfs import (
     to_multihash,
 )
 from packages.valory.skills.task_execution.utils.local_cid import compute_cidv1
+from packages.valory.skills.task_execution.utils.request_id import to_request_id_bytes32
 from packages.valory.skills.task_execution.utils.task import AnyToolAsTask
 
 PENDING_TASKS = "pending_tasks"
@@ -278,21 +279,6 @@ def _iso_z(dt: datetime) -> str:
     if dt.tzinfo is None:
         raise ValueError("_iso_z requires a timezone-aware datetime; got a naive value")
     return dt.astimezone(timezone.utc).isoformat().replace("+00:00", "Z")
-
-
-def _to_bytes32(value: Any) -> bytes:
-    """Coerce a marketplace ``requestId`` to a 32-byte big-endian value.
-
-    :param value: the raw ``requestId`` (``int`` or ``bytes``/``bytearray``).
-    :return: a ``bytes`` value of length exactly 32.
-    :raises ValueError: if a ``bytes`` value longer than 32 bytes is passed.
-    """
-    if isinstance(value, (bytes, bytearray)):
-        raw = bytes(value)
-        if len(raw) > 32:
-            raise ValueError(f"requestId is {len(raw)} bytes, expected <= 32")
-        return raw.rjust(32, b"\x00")
-    return int(value).to_bytes(32, "big")
 
 
 # Cap on the JSON-serialised size of each ``raw_content`` blob attached to
@@ -1014,22 +1000,25 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         if self.last_status_check + STATUS_CHECK_INTERVAL > time.time():
             return
 
-        pending_tasks_count = len(self.pending_tasks)
         # no pending tasks to check
-        if pending_tasks_count == 0:
+        if len(self.pending_tasks) == 0:
             return
 
-        self.context.logger.info(
-            f"Checking status change for {pending_tasks_count} pending tasks..."
-        )
         pending_tasks_request_ids = [
-            _to_bytes32(t["requestId"])
+            to_request_id_bytes32(t["requestId"])
             for t in self.pending_tasks
             if not t.get("is_offchain")
         ]
+        # All-offchain queue: nothing to send but stamp the throttle so
+        # ``act()`` doesn't re-enter every tick.
         if not pending_tasks_request_ids:
+            self.last_status_check_time = time.time()
             return
 
+        self.context.logger.info(
+            f"Checking status change for {len(pending_tasks_request_ids)} "
+            "pending tasks..."
+        )
         contract_api_msg, _ = self.context.contract_dialogues.create(
             performative=ContractApiMessage.Performative.GET_STATE,
             contract_address=self.params.mech_marketplace_address,
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -280,6 +280,21 @@ def _iso_z(dt: datetime) -> str:
     return dt.astimezone(timezone.utc).isoformat().replace("+00:00", "Z")
 
 
+def _to_bytes32(value: Any) -> bytes:
+    """Coerce a marketplace ``requestId`` to a 32-byte big-endian value.
+
+    Handles the two shapes that coexist in the pending-tasks queue: fresh
+    marketplace ingest stores the id as ``bytes`` (from the on-chain
+    ``bytes32`` decode), and execution-pop rewrites it to ``int``.
+
+    :param value: the raw ``requestId`` (``int`` or ``bytes``/``bytearray``).
+    :return: a ``bytes`` value of length exactly 32.
+    """
+    if isinstance(value, (bytes, bytearray)):
+        return bytes(value).rjust(32, b"\x00")[-32:]
+    return int(value).to_bytes(32, "big")
+
+
 # Cap on the JSON-serialised size of each ``raw_content`` blob attached to
 # a predict-api event. The blob rides Tendermint consensus replication into
 # ``synchronized_data`` on every agent and the field is requester-influenced
@@ -1009,9 +1024,13 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         )
         # ``fetch_batch_request_id_status`` encodes each id into a
         # ``bytes32[]`` ABI slot, which the strict ``BytesEncoder`` rejects
-        # if it receives an int. Coerce to 32-byte big-endian here.
+        # if it receives an int. Fresh marketplace ingest stores
+        # ``requestId`` as ``bytes`` (see ``mech_marketplace/contract.py``);
+        # execution-pop converts it to ``int`` (see
+        # ``_extract_request_data_from_task``). Both shapes may coexist in
+        # the queue, so coerce either into a 32-byte big-endian value.
         pending_tasks_request_ids = [
-            int(t["requestId"]).to_bytes(32, "big") for t in self.pending_tasks
+            _to_bytes32(t["requestId"]) for t in self.pending_tasks
         ]
 
         contract_api_msg, _ = self.context.contract_dialogues.create(
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -283,15 +283,15 @@ def _iso_z(dt: datetime) -> str:
 def _to_bytes32(value: Any) -> bytes:
     """Coerce a marketplace ``requestId`` to a 32-byte big-endian value.
 
-    Handles the two shapes that coexist in the pending-tasks queue: fresh
-    marketplace ingest stores the id as ``bytes`` (from the on-chain
-    ``bytes32`` decode), and execution-pop rewrites it to ``int``.
-
     :param value: the raw ``requestId`` (``int`` or ``bytes``/``bytearray``).
     :return: a ``bytes`` value of length exactly 32.
+    :raises ValueError: if a ``bytes`` value longer than 32 bytes is passed.
     """
     if isinstance(value, (bytes, bytearray)):
-        return bytes(value).rjust(32, b"\x00")[-32:]
+        raw = bytes(value)
+        if len(raw) > 32:
+            raise ValueError(f"requestId is {len(raw)} bytes, expected <= 32")
+        return raw.rjust(32, b"\x00")
     return int(value).to_bytes(32, "big")
 
 
@@ -1022,16 +1022,13 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         self.context.logger.info(
             f"Checking status change for {pending_tasks_count} pending tasks..."
         )
-        # ``fetch_batch_request_id_status`` encodes each id into a
-        # ``bytes32[]`` ABI slot, which the strict ``BytesEncoder`` rejects
-        # if it receives an int. Fresh marketplace ingest stores
-        # ``requestId`` as ``bytes`` (see ``mech_marketplace/contract.py``);
-        # execution-pop converts it to ``int`` (see
-        # ``_extract_request_data_from_task``). Both shapes may coexist in
-        # the queue, so coerce either into a 32-byte big-endian value.
         pending_tasks_request_ids = [
-            _to_bytes32(t["requestId"]) for t in self.pending_tasks
+            _to_bytes32(t["requestId"])
+            for t in self.pending_tasks
+            if not t.get("is_offchain")
         ]
+        if not pending_tasks_request_ids:
+            return
 
         contract_api_msg, _ = self.context.contract_dialogues.create(
             performative=ContractApiMessage.Performative.GET_STATE,
--- a/packages/valory/skills/task_execution/behaviours.py
+++ b/packages/valory/skills/task_execution/behaviours.py
@@ -1007,7 +1007,12 @@ class TaskExecutionBehaviour(SimpleBehaviour):
         self.context.logger.info(
             f"Checking status change for {pending_tasks_count} pending tasks..."
         )
-        pending_tasks_request_ids = [t["requestId"] for t in self.pending_tasks]
+        # ``fetch_batch_request_id_status`` encodes each id into a
+        # ``bytes32[]`` ABI slot, which the strict ``BytesEncoder`` rejects
+        # if it receives an int. Coerce to 32-byte big-endian here.
+        pending_tasks_request_ids = [
+            int(t["requestId"]).to_bytes(32, "big") for t in self.pending_tasks
+        ]
 
         contract_api_msg, _ = self.context.contract_dialogues.create(
             performative=ContractApiMessage.Performative.GET_STATE,
--- a/packages/valory/skills/task_execution/handlers.py
+++ b/packages/valory/skills/task_execution/handlers.py
@@ -61,7 +61,10 @@ from packages.valory.protocols.ipfs import IpfsMessage
 from packages.valory.protocols.kv_store.message import KvStoreMessage
 from packages.valory.protocols.ledger_api import LedgerApiMessage
 from packages.valory.skills.abstract_round_abci.handlers import AbstractResponseHandler
-from packages.valory.skills.task_execution.behaviours import PREDICT_API_EVENTS
+from packages.valory.skills.task_execution.behaviours import (
+    PREDICT_API_EVENTS,
+    _to_bytes32,
+)
 from packages.valory.skills.task_execution.dialogues import HttpDialogue
 from packages.valory.skills.task_execution.models import Params
 from packages.valory.skills.task_execution.utils import preimage as preimage_buffer
@@ -770,12 +773,16 @@ class ContractHandler(BaseHandler):
         :param body: the on-chain status response body.
         """
         before = len(self.pending_tasks)
-        on_chain_ids = body[BodyKey.REQUEST_IDS.value]
+        # ``on_chain_ids`` is ``List[bytes]`` from ``codec.decode(["bytes32[]"], ...)``;
+        # ``req["requestId"]`` may be ``int`` on tasks re-queued after execution.
+        # Normalise both to 32-byte bytes so a retried task doesn't get pruned
+        # against a set of a different type.
+        on_chain_ids = {_to_bytes32(rid) for rid in body[BodyKey.REQUEST_IDS.value]}
         self.context.shared_state[PENDING_TASKS] = [
             req
             for req in self.pending_tasks
             if req.get(RequestKey.IS_OFFCHAIN.value)
-            or req[RequestKey.REQUEST_ID_CAMEL.value] in on_chain_ids
+            or _to_bytes32(req[RequestKey.REQUEST_ID_CAMEL.value]) in on_chain_ids
         ]
         after = len(self.pending_tasks)
         self.context.logger.info(
--- a/packages/valory/skills/task_execution/handlers.py
+++ b/packages/valory/skills/task_execution/handlers.py
@@ -61,10 +61,7 @@ from packages.valory.protocols.ipfs import IpfsMessage
 from packages.valory.protocols.kv_store.message import KvStoreMessage
 from packages.valory.protocols.ledger_api import LedgerApiMessage
 from packages.valory.skills.abstract_round_abci.handlers import AbstractResponseHandler
-from packages.valory.skills.task_execution.behaviours import (
-    PREDICT_API_EVENTS,
-    _to_bytes32,
-)
+from packages.valory.skills.task_execution.behaviours import PREDICT_API_EVENTS
 from packages.valory.skills.task_execution.dialogues import HttpDialogue
 from packages.valory.skills.task_execution.models import Params
 from packages.valory.skills.task_execution.utils import preimage as preimage_buffer
@@ -79,6 +76,7 @@ from packages.valory.skills.task_execution.utils.local_cid import compute_cidv1
 from packages.valory.skills.task_execution.utils.request_id import (
     compute_request_id,
     recover_eoa_signer,
+    to_request_id_bytes32,
 )
 
 PENDING_TASKS = "pending_tasks"
@@ -685,8 +683,11 @@ class ContractHandler(BaseHandler):
         ):
             # handle the undelivered requests response from data and wait_for_timeout_tasks
             self._handle_get_undelivered_reqs(body)
-        if body.get(BodyKey.REQUEST_IDS.value):
+        if BodyKey.REQUEST_IDS.value in body:
             # handle the request id status check response
+            # ``fetch_batch_request_id_status`` returns only the undelivered
+            # subset, so an empty list is a valid "all delivered" answer that
+            # must still prune the queue.
             self._update_pending_list(body)
         if body.get(BodyKey.MECH_TYPE.value):
             # handle the mech type response
@@ -773,16 +774,15 @@ class ContractHandler(BaseHandler):
         :param body: the on-chain status response body.
         """
         before = len(self.pending_tasks)
-        # ``on_chain_ids`` is ``List[bytes]`` from ``codec.decode(["bytes32[]"], ...)``;
-        # ``req["requestId"]`` may be ``int`` on tasks re-queued after execution.
-        # Normalise both to 32-byte bytes so a retried task doesn't get pruned
-        # against a set of a different type.
-        on_chain_ids = {_to_bytes32(rid) for rid in body[BodyKey.REQUEST_IDS.value]}
+        on_chain_ids = {
+            to_request_id_bytes32(rid) for rid in body[BodyKey.REQUEST_IDS.value]
+        }
         self.context.shared_state[PENDING_TASKS] = [
             req
             for req in self.pending_tasks
             if req.get(RequestKey.IS_OFFCHAIN.value)
-            or _to_bytes32(req[RequestKey.REQUEST_ID_CAMEL.value]) in on_chain_ids
+            or to_request_id_bytes32(req[RequestKey.REQUEST_ID_CAMEL.value])
+            in on_chain_ids
         ]
         after = len(self.pending_tasks)
         self.context.logger.info(
--- a/packages/valory/skills/task_execution/tests/test_behaviours.py
+++ b/packages/valory/skills/task_execution/tests/test_behaviours.py
@@ -1607,6 +1607,63 @@ def test_update_pending_tasks_no_pending_tasks(
     behaviour._update_pending_tasks()  # should not raise
 
 
+def test_update_pending_tasks_coerces_request_ids_to_bytes32(
+    behaviour: Any,
+    params_stub: Any,
+    shared_state: Dict[str, Any],
+    monkeypatch: Any,
+) -> None:
+    """``request_ids`` reach ``fetch_batch_request_id_status`` as 32-byte bytes.
+
+    The callee encodes them into a ``bytes32[]`` ABI slot. Newer ``eth_abi``
+    versions reject an ``int`` there with ``EncodingTypeError``, so the
+    caller must coerce first. This test seeds a marketplace-hash-derived
+    ``uint256`` id (~1e77) that would fail encoding without the coercion,
+    and asserts the value on the outgoing kwargs is ``bytes`` of length 32.
+
+    :param behaviour: TaskExecutionBehaviour fixture.
+    :param params_stub: params fixture.
+    :param shared_state: shared_state fixture used to seed ``PENDING_TASKS``.
+    :param monkeypatch: pytest monkeypatch used to freeze ``time.time`` and
+        replace ``contract_dialogues.create`` with a capturing stub.
+    """
+    params_stub.use_mech_marketplace = True
+    params_stub.in_flight_req = False
+    behaviour.last_status_check_time = 0.0
+    monkeypatch.setattr(time, "time", lambda: beh_mod.STATUS_CHECK_INTERVAL + 1.0)
+    huge_id = (
+        100705699761465541306241857262966577157937941808748966974775453031896335917267
+    )
+    small_id = 42
+    shared_state[beh_mod.PENDING_TASKS] = [
+        {"requestId": huge_id},
+        {"requestId": small_id},
+    ]
+    captured: Dict[str, Any] = {}
+
+    def _capture_create(*a: Any, **k: Any) -> Any:
+        """Capture the kwargs passed to ``contract_dialogues.create``."""
+        captured.update(k)
+        return SimpleNamespace(), SimpleNamespace(
+            dialogue_label=SimpleNamespace(dialogue_reference=("nonce-x", ""))
+        )
+
+    monkeypatch.setattr(behaviour.context.contract_dialogues, "create", _capture_create)
+    behaviour._update_pending_tasks()
+    kwargs = captured.get("kwargs")
+    assert kwargs is not None, "contract_dialogues.create was not called"
+    body = kwargs.body if hasattr(kwargs, "body") else dict(kwargs)
+    request_ids = body["request_ids"]
+    assert len(request_ids) == 2
+    for rid in request_ids:
+        assert isinstance(rid, bytes), f"expected bytes, got {type(rid).__name__}"
+        assert len(rid) == 32, f"expected 32 bytes, got {len(rid)}"
+    assert (
+        int.from_bytes(request_ids[0], "big") == huge_id
+    ), "round-trip must preserve the uint256 value"
+    assert int.from_bytes(request_ids[1], "big") == small_id
+
+
 # ---------------------------------------------------------------------------
 # send_message
 # ---------------------------------------------------------------------------
--- a/packages/valory/skills/task_execution/tests/test_behaviours.py
+++ b/packages/valory/skills/task_execution/tests/test_behaviours.py
@@ -1607,38 +1607,56 @@ def test_update_pending_tasks_no_pending_tasks(
     behaviour._update_pending_tasks()  # should not raise
 
 
+_HUGE_UINT256 = (
+    100705699761465541306241857262966577157937941808748966974775453031896335917267
+)
+_BYTES32_HIGH_BIT = b"\xaa" * 32
+
+
+@pytest.mark.parametrize(
+    "seeded, expected_int",
+    [
+        pytest.param(_HUGE_UINT256, _HUGE_UINT256, id="int-hash-derived-uint256"),
+        pytest.param(42, 42, id="int-small-legacy"),
+        pytest.param(
+            _BYTES32_HIGH_BIT,
+            int.from_bytes(_BYTES32_HIGH_BIT, "big"),
+            id="bytes32-fresh-marketplace-ingest",
+        ),
+        pytest.param(b"\x00" * 32, 0, id="bytes32-zero"),
+    ],
+)
 def test_update_pending_tasks_coerces_request_ids_to_bytes32(
     behaviour: Any,
     params_stub: Any,
     shared_state: Dict[str, Any],
     monkeypatch: Any,
+    seeded: Any,
+    expected_int: int,
 ) -> None:
     """``request_ids`` reach ``fetch_batch_request_id_status`` as 32-byte bytes.
 
     The callee encodes them into a ``bytes32[]`` ABI slot. Newer ``eth_abi``
     versions reject an ``int`` there with ``EncodingTypeError``, so the
-    caller must coerce first. This test seeds a marketplace-hash-derived
-    ``uint256`` id (~1e77) that would fail encoding without the coercion,
-    and asserts the value on the outgoing kwargs is ``bytes`` of length 32.
+    caller must coerce first. Fresh marketplace ingest stores the id as
+    ``bytes`` (from the on-chain ``bytes32[]`` decode in
+    ``mech_marketplace/contract.py``); execution-pop rewrites it to
+    ``int``. Both shapes may coexist in the queue at status-check time, so
+    the coercion has to accept either.
 
     :param behaviour: TaskExecutionBehaviour fixture.
     :param params_stub: params fixture.
     :param shared_state: shared_state fixture used to seed ``PENDING_TASKS``.
     :param monkeypatch: pytest monkeypatch used to freeze ``time.time`` and
         replace ``contract_dialogues.create`` with a capturing stub.
+    :param seeded: value stashed on ``t["requestId"]`` (``int`` or ``bytes``).
+    :param expected_int: uint256 the round-trip must recover.
     """
     params_stub.use_mech_marketplace = True
     params_stub.in_flight_req = False
     behaviour.last_status_check_time = 0.0
     monkeypatch.setattr(time, "time", lambda: beh_mod.STATUS_CHECK_INTERVAL + 1.0)
-    huge_id = (
-        100705699761465541306241857262966577157937941808748966974775453031896335917267
-    )
-    small_id = 42
-    shared_state[beh_mod.PENDING_TASKS] = [
-        {"requestId": huge_id},
-        {"requestId": small_id},
-    ]
+    shared_state[beh_mod.PENDING_TASKS] = [{"requestId": seeded}]
     captured: Dict[str, Any] = {}
 
     def _capture_create(*a: Any, **k: Any) -> Any:
@@ -1654,14 +1672,63 @@ def test_update_pending_tasks_coerces_request_ids_to_bytes32(
     assert kwargs is not None, "contract_dialogues.create was not called"
     body = kwargs.body if hasattr(kwargs, "body") else dict(kwargs)
     request_ids = body["request_ids"]
-    assert len(request_ids) == 2
-    for rid in request_ids:
-        assert isinstance(rid, bytes), f"expected bytes, got {type(rid).__name__}"
-        assert len(rid) == 32, f"expected 32 bytes, got {len(rid)}"
+    assert len(request_ids) == 1
+    rid = request_ids[0]
+    assert isinstance(rid, bytes), f"expected bytes, got {type(rid).__name__}"
+    assert len(rid) == 32, f"expected 32 bytes, got {len(rid)}"
     assert (
-        int.from_bytes(request_ids[0], "big") == huge_id
+        int.from_bytes(rid, "big") == expected_int
     ), "round-trip must preserve the uint256 value"
-    assert int.from_bytes(request_ids[1], "big") == small_id
+
+
+def test_update_pending_tasks_coerces_mixed_int_and_bytes(
+    behaviour: Any,
+    params_stub: Any,
+    shared_state: Dict[str, Any],
+    monkeypatch: Any,
+) -> None:
+    """A queue holding both int and bytes ids in the same batch coerces cleanly.
+
+    The realistic production shape: fresh marketplace ingest (``bytes``)
+    plus a re-enqueued task after execution-pop (``int``) can sit
+    side-by-side in ``PENDING_TASKS`` when a status-check tick fires.
+
+    :param behaviour: TaskExecutionBehaviour fixture.
+    :param params_stub: params fixture.
+    :param shared_state: shared_state fixture used to seed ``PENDING_TASKS``.
+    :param monkeypatch: pytest monkeypatch used to freeze ``time.time`` and
+        replace ``contract_dialogues.create`` with a capturing stub.
+    """
+    params_stub.use_mech_marketplace = True
+    params_stub.in_flight_req = False
+    behaviour.last_status_check_time = 0.0
+    monkeypatch.setattr(time, "time", lambda: beh_mod.STATUS_CHECK_INTERVAL + 1.0)
+    shared_state[beh_mod.PENDING_TASKS] = [
+        {"requestId": _HUGE_UINT256},
+        {"requestId": _BYTES32_HIGH_BIT},
+        {"requestId": 42},
+    ]
+    captured: Dict[str, Any] = {}
+
+    def _capture_create(*a: Any, **k: Any) -> Any:
+        """Capture the kwargs passed to ``contract_dialogues.create``."""
+        captured.update(k)
+        return SimpleNamespace(), SimpleNamespace(
+            dialogue_label=SimpleNamespace(dialogue_reference=("nonce-x", ""))
+        )
+
+    monkeypatch.setattr(behaviour.context.contract_dialogues, "create", _capture_create)
+    behaviour._update_pending_tasks()
+    kwargs = captured.get("kwargs")
+    assert kwargs is not None
+    body = kwargs.body if hasattr(kwargs, "body") else dict(kwargs)
+    request_ids = body["request_ids"]
+    assert len(request_ids) == 3
+    for rid in request_ids:
+        assert isinstance(rid, bytes) and len(rid) == 32
+    assert int.from_bytes(request_ids[0], "big") == _HUGE_UINT256
+    assert request_ids[1] == _BYTES32_HIGH_BIT
+    assert int.from_bytes(request_ids[2], "big") == 42
 
 
 # ---------------------------------------------------------------------------
--- a/packages/valory/skills/task_execution/tests/test_behaviours.py
+++ b/packages/valory/skills/task_execution/tests/test_behaviours.py
@@ -32,6 +32,9 @@ import pytest
 from prometheus_client import REGISTRY
 
 import packages.valory.skills.task_execution.behaviours as beh_mod
+from packages.valory.skills.task_execution.utils.request_id import (
+    to_request_id_bytes32,
+)
 
 
 def _dispatch_target(cb: Any) -> Any:
@@ -1683,11 +1686,6 @@ def test_update_pending_tasks_carves_out_offchain_tasks(
 ) -> None:
     """Off-chain tasks are filtered out of the on-chain batch status check.
 
-    Off-chain request ids do not exist on chain until settlement, so
-    querying the marketplace for their status is a wasted call and
-    (until F4) enlarged the outgoing ``bytes32[]``. Also verify the
-    all-off-chain case short-circuits without opening a dialogue.
-
     :param behaviour: TaskExecutionBehaviour fixture.
     :param params_stub: params fixture.
     :param shared_state: shared_state fixture used to seed ``PENDING_TASKS``.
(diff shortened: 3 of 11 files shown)


packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1007,7 +1007,12 @@ def _update_pending_tasks(self) -> None:
         self.context.logger.info(
             f"Checking status change for {pending_tasks_count} pending tasks..."
         )
-        pending_tasks_request_ids = [t["requestId"] for t in self.pending_tasks]
+        # ``fetch_batch_request_id_status`` encodes each id into a
+        # ``bytes32[]`` ABI slot, which the strict ``BytesEncoder`` rejects
+        # if it receives an int. Coerce to 32-byte big-endian here.
+        pending_tasks_request_ids = [
+            int(t["requestId"]).to_bytes(32, "big") for t in self.pending_tasks
```
jmoreira-valory: 🟠 **F1 · HIGH — int() on a bytes32 requestId raises ValueError on every status tick**

On-chain marketplace request ids arrive as **`bytes`**, not `int`. `MarketplaceRequest.requestIds` is `bytes32[]` in the ABI, `get_marketplace_undelivered_reqs` puts the decoded value straight into `"requestId"`, and `handlers.filter_requests` appends the dict unchanged. The `bytes -> int` normalisation in this same file (line 920-923, `# convert request id to int if it's bytes`) runs only when the task is **popped** in `_execute_task` - so while a task sits in `pending_tasks`, its id is `bytes`.

`int()` on a bytes object parses it as an ASCII decimal literal, so `int(b'\x8a\xf3...')` raises `ValueError: invalid literal for int() with base 10`, not a conversion. That happens inside `_update_pending_tasks()`, which `act()` calls with no `try`/`except`; `Behaviour.act_wrapper` re-raises it as `AEAActException`, and the mech agent sets no `skill_exception_policy` (open-aea default is `propagate`), so the agent stops. Net effect: on the exact population this fix targets (marketplace mech, non-empty on-chain pending queue), a logged contract-api 500 every ~10 min is replaced by a deterministic agent crash every ~10 min.

`fetch_batch_request_id_status(request_ids: List[bytes])` already wants `bytes`, so on-chain ids should pass through untouched; only `int` ids (off-chain, and tasks re-queued by `_handle_timeout_task`, which are `int` by then) need converting:

```suggestion
            t["requestId"]
            if isinstance(t["requestId"], bytes)
            else int(t["requestId"]).to_bytes(32, "big")
            for t in self.pending_tasks
```

Worth updating the comment above too - the ids are not uniformly ints coming in.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1607,6 +1607,63 @@ def test_update_pending_tasks_no_pending_tasks(
...
+    """
+    params_stub.use_mech_marketplace = True
+    params_stub.in_flight_req = False
+    behaviour.last_status_check_time = 0.0
+    monkeypatch.setattr(time, "time", lambda: beh_mod.STATUS_CHECK_INTERVAL + 1.0)
+    huge_id = (
+        100705699761465541306241857262966577157937941808748966974775453031896335917267
+    )
+    small_id = 42
+    shared_state[beh_mod.PENDING_TASKS] = [
```
jmoreira-valory: 🟡 **F2 · MEDIUM — New test seeds only int ids, so it cannot catch the on-chain shape**

Both seeded tasks (`huge_id`, `small_id`) are `int`, which is the off-chain / re-queued shape. The on-chain marketplace shape - the one that actually breaks - is `bytes`, and the file already has a precedent for it at line 1470 (`test_execute_task_bytes_request_id_converted` seeds `req_id_int.to_bytes(32, byteorder="big")` into `PENDING_TASKS`).

Because the test never exercises a `bytes` id, it passes against a fix that raises `ValueError` in production. The mutation check quoted in the PR body (revert -> `expected bytes, got int`) is real but only covers half the input domain. Please add a third entry so the test is a genuine gate on the finding above:

```python
    onchain_id = huge_id.to_bytes(32, "big")
    shared_state[beh_mod.PENDING_TASKS] = [
        {"requestId": onchain_id},   # on-chain: bytes32 from the event
        {"requestId": huge_id},      # re-queued after timeout: int
        {"requestId": small_id},     # off-chain: int
    ]
```

and assert the bytes entry round-trips to the same 32 bytes it went in as.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1007,7 +1007,12 @@ def _update_pending_tasks(self) -> None:
         self.context.logger.info(
             f"Checking status change for {pending_tasks_count} pending tasks..."
         )
-        pending_tasks_request_ids = [t["requestId"] for t in self.pending_tasks]
+        # ``fetch_batch_request_id_status`` encodes each id into a
+        # ``bytes32[]`` ABI slot, which the strict ``BytesEncoder`` rejects
+        # if it receives an int. Coerce to 32-byte big-endian here.
+        pending_tasks_request_ids = [
```
jmoreira-valory, line 1007: 🟡 **F3 · MEDIUM — Re-enabling the check drops re-queued timed-out tasks (int vs bytes32)**

This PR turns a code path back on that has been dead in production, so the consumer of the response matters. `handlers.py:773-779` (`_update_pending_list`) rewrites `pending_tasks` to the ids the contract returned:

```python
on_chain_ids = body[BodyKey.REQUEST_IDS.value]   # tuple of bytes32 (codec.decode(["bytes32[]"]))
... if req.get("is_offchain") or req["requestId"] in on_chain_ids
```

Fresh on-chain tasks hold `bytes` and match. But a task re-queued by `_handle_timeout_task` (`self.pending_tasks.append(executing_task)`, line 1399) holds an `int` id, because `_execute_task` converted it at pop time - and it is not `is_offchain`. `int in (bytes, ...)` is always `False`, so on the first successful status check every previously-timed-out on-chain task is silently pruned from the queue and never delivered, while the requester keeps waiting for the on-chain timeout. That is a paid request lost per occurrence.

Fix by normalising both sides to one type. Cheapest version: keep the outgoing ids as `bytes` (finding above) and compare on `bytes` in `_update_pending_list`, e.g. normalise each `req["requestId"]` through the same helper before the membership test. A shared `_to_bytes32(request_id)` helper used by the behaviour and the handler would also stop this drifting again.

jmoreira-valory, line 1007: 🔵 **F4 · LOW — Off-chain pending tasks are sent to the on-chain batch status check**

`self.pending_tasks` includes `is_offchain=True` entries, whose ids do not exist on chain until settlement - which is exactly why `_update_pending_list` carves them out of the response filtering. Sending them here just enlarges the `bytes32[]` constructor payload of the throwaway batch contract and asks the marketplace for statuses that cannot exist.

Since the comprehension is being touched anyway, mirroring the handler's carve-out keeps the two ends consistent:

```python
        pending_tasks_request_ids = [
            _to_bytes32(t["requestId"])
            for t in self.pending_tasks
            if not t.get("is_offchain")
        ]
```

Note the early-return above counts `len(self.pending_tasks)`, so with an all-off-chain queue this would send an empty `request_ids` list - guard on the filtered list being non-empty before creating the dialogue.

jmoreira-valory (changes requested):

**What this does.** The marketplace status check in `task_execution` sends a list of pending request ids to the `fetch_batch_request_id_status` contract callable, which ABI-encodes them into a `bytes32[]` slot. Newer `eth_abi` refuses to encode an `int` there, so the call has been returning a contract-api 500 on every ~10-minute tick and the pending-queue reconciliation has been silently dead. This PR coerces each id to 32 big-endian bytes at the call site and adds a unit test asserting the outgoing kwargs carry `bytes` of length 32.

The diagnosis is right and well-evidenced, and the fix is in the correct place. The problem is the coercion assumes `pending_tasks` holds `int` request ids, and for the on-chain marketplace path it does not.

What I verified at HEAD (not taken from the PR body):

- `MechMarketplace.json` ABI: `MarketplaceRequest(priorityMech, requester, numRequests, requestIds: bytes32[], requestDatas)` -> web3 decodes each id to `bytes`.
- `mech_marketplace/contract.py::get_marketplace_undelivered_reqs` stores that value verbatim as `"requestId": rid` in the dicts that become `pending_tasks`; `handlers.filter_requests` appends them without normalising.
- `behaviours.py:920-923` converts `bytes -> int` only when a task is **popped** in `_execute_task`, i.e. after it leaves `pending_tasks`.
- The repo's own `test_execute_task_bytes_request_id_converted` (test_behaviours.py:1470) seeds `PENDING_TASKS` with a `bytes` id, documenting the shape.
- `fetch_batch_request_id_status(request_ids: List[bytes])` already expects `bytes`, so on-chain ids need no conversion at all - only off-chain / re-queued (`int`) ones do.

CI is green (22/22 check-runs at `45a5a26`) and the CID/fingerprint cascade in `packages.json`, `aea-config.yaml`, `service.yaml` and the three `skill.yaml`s is consistent, so `packages lock` hygiene is fine.

**Findings** — 4 · 🟠 1 High · 🟡 2 Medium · 🔵 1 Low

| F1 | 🟠 HIGH | [packages/valory/skills/task_execution/behaviours.py:1014](https://github.com/valory-xyz/mech/blob/45a5a26e8b7dd33fddc809113079693b1f43795d/packages/valory/skills/task_execution/behaviours.py#L1014) | **int() on a bytes32 requestId raises ValueError on every status tick** |
| F2 | 🟡 MEDIUM | [packages/valory/skills/task_execution/tests/test_behaviours.py:1638](https://github.com/valory-xyz/mech/blob/45a5a26e8b7dd33fddc809113079693b1f43795d/packages/valory/skills/task_execution/tests/test_behaviours.py#L1638) | **New test seeds only int ids, so it cannot catch the on-chain shape** |
| F3 | 🟡 MEDIUM | [packages/valory/skills/task_execution/behaviours.py:1013](https://github.com/valory-xyz/mech/blob/45a5a26e8b7dd33fddc809113079693b1f43795d/packages/valory/skills/task_execution/behaviours.py#L1013) | **Re-enabling the check drops re-queued timed-out tasks (int vs bytes32)** |
| F4 | 🔵 LOW | [packages/valory/skills/task_execution/behaviours.py:1013](https://github.com/valory-xyz/mech/blob/45a5a26e8b7dd33fddc809113079693b1f43795d/packages/valory/skills/task_execution/behaviours.py#L1013) | **Off-chain pending tasks are sent to the on-chain batch status check** |

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -280,6 +280,21 @@ def _iso_z(dt: datetime) -> str:
     return dt.astimezone(timezone.utc).isoformat().replace("+00:00", "Z")
 
 
+def _to_bytes32(value: Any) -> bytes:
+    """Coerce a marketplace ``requestId`` to a 32-byte big-endian value.
+
+    Handles the two shapes that coexist in the pending-tasks queue: fresh
+    marketplace ingest stores the id as ``bytes`` (from the on-chain
+    ``bytes32`` decode), and execution-pop rewrites it to ``int``.
```
OjusWiZard: Per the repo comment bar, this paragraph is recoverable from `git blame` -> this commit -> the PR description, which says the same thing at greater length. Suggest keeping the one-line summary plus the darglint-required `:param:`/`:return:` and dropping these three lines.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -280,6 +280,21 @@ def _iso_z(dt: datetime) -> str:
...
+
+    :param value: the raw ``requestId`` (``int`` or ``bytes``/``bytearray``).
+    :return: a ``bytes`` value of length exactly 32.
+    """
+    if isinstance(value, (bytes, bytearray)):
+        return bytes(value).rjust(32, b"\x00")[-32:]
```
OjusWiZard: `rjust(32, b"\x00")[-32:]` is a no-op for a 32-byte input, but for anything **longer** it silently drops the leading bytes and returns a wrong id. Unreachable today (every `bytes` id is an exactly-32-byte ABI decode), but it turns a future shape error into silent corruption rather than a loud failure.

Suggest raising instead:

```python
if isinstance(value, (bytes, bytearray)):
    raw = bytes(value)
    if len(raw) > 32:
        raise ValueError(f"requestId is {len(raw)} bytes, expected <= 32")
    return raw.rjust(32, b"\x00")
```

Minor: the `bytearray` arm looks unreachable from current callers — web3 decodes `bytes32` to `HexBytes`, a `bytes` subclass.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1007,7 +1022,16 @@ def _update_pending_tasks(self) -> None:
...
             f"Checking status change for {pending_tasks_count} pending tasks..."
         )
-        pending_tasks_request_ids = [t["requestId"] for t in self.pending_tasks]
+        # ``fetch_batch_request_id_status`` encodes each id into a
+        # ``bytes32[]`` ABI slot, which the strict ``BytesEncoder`` rejects
+        # if it receives an int. Fresh marketplace ingest stores
+        # ``requestId`` as ``bytes`` (see ``mech_marketplace/contract.py``);
+        # execution-pop converts it to ``int`` (see
+        # ``_extract_request_data_from_task``). Both shapes may coexist in
+        # the queue, so coerce either into a 32-byte big-endian value.
```
OjusWiZard: This block restates the commit message and the PR description almost verbatim, and duplicates the `_to_bytes32` docstring three lines below the helper. `git blame` on line 1032 resolves straight to this commit and PR, so the chain already answers it. Suggest deleting the whole block — the call `_to_bytes32(t["requestId"])` is self-describing.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1607,6 +1607,130 @@ def test_update_pending_tasks_no_pending_tasks(
...
+    the coercion has to accept either.
+
+    :param behaviour: TaskExecutionBehaviour fixture.
+    :param params_stub: params fixture.
+    :param shared_state: shared_state fixture used to seed ``PENDING_TASKS``.
+    :param monkeypatch: pytest monkeypatch used to freeze ``time.time`` and
+        replace ``contract_dialogues.create`` with a capturing stub.
+    :param seeded: value stashed on ``t["requestId"]`` (``int`` or ``bytes``).
+    :param expected_int: uint256 the round-trip must recover.
+    """
```
OjusWiZard, line 1654: The test name already carries the claim, and the rationale here is a third copy of what the PR description and the `behaviours.py` comment both say. Suggest trimming to a one-line summary.

One genuine coverage gap while you are here: nothing pins `_to_bytes32` on bytes shorter than 32, bytes longer than 32, or a negative / `>= 2**256` int. The oversized-bytes case is the one worth a test, since it currently truncates silently.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1607,6 +1607,130 @@ def test_update_pending_tasks_no_pending_tasks(
...
+        captured.update(k)
+        return SimpleNamespace(), SimpleNamespace(
+            dialogue_label=SimpleNamespace(dialogue_reference=("nonce-x", ""))
+        )
+
+    monkeypatch.setattr(behaviour.context.contract_dialogues, "create", _capture_create)
+    behaviour._update_pending_tasks()
+    kwargs = captured.get("kwargs")
+    assert kwargs is not None, "contract_dialogues.create was not called"
+    body = kwargs.body if hasattr(kwargs, "body") else dict(kwargs)
```
OjusWiZard: The `hasattr` fallback is dead — `ContractApiMessage.Kwargs` defines `body` as a property over `__slots__ = ("_body",)`, so it is always present, and the call site always constructs a real `Kwargs`. `body = kwargs.body` is enough (same on line 1724).

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1607,6 +1607,130 @@ def test_update_pending_tasks_no_pending_tasks(
...
+    The realistic production shape: fresh marketplace ingest (``bytes``)
+    plus a re-enqueued task after execution-pop (``int``) can sit
+    side-by-side in ``PENDING_TASKS`` when a status-check tick fires.
+
+    :param behaviour: TaskExecutionBehaviour fixture.
+    :param params_stub: params fixture.
+    :param shared_state: shared_state fixture used to seed ``PENDING_TASKS``.
+    :param monkeypatch: pytest monkeypatch used to freeze ``time.time`` and
+        replace ``contract_dialogues.create`` with a capturing stub.
+    """
```
OjusWiZard: This test looks subsumed by the parametrized one above it. The coercion is a per-element list comprehension with no cross-element state, so exercising the three shapes together adds only `len == 3` and order preservation — both plain comprehension semantics. It also copy-pastes `_capture_create` verbatim.

If you want to keep the mixed-queue case explicit, a single extra `pytest.param` with a list-valued `seeded` would cover it without the duplicate stub. Not blocking either way.

OjusWiZard (dismissed):
### Approving — the fix is correct. One blocking-severity follow-up it exposes.

Verified locally: full `task_execution` / `task_submission_abci` / `mech_abci` suites green (900 passed), and the new tests are genuine mutation tests — reverting the coercion fails them with `expected bytes, got int`. The root-cause analysis in the description checks out: fresh marketplace ingest stores `requestId` as `bytes` (`mech_marketplace/contract.py:299`, from the `bytes32[]` event decode) and execution-pop rewrites it to `int` (`behaviours.py:936-938`), so both shapes really can coexist in the queue.

---

## :rotating_light: This PR un-blocks a type-mismatched response handler (not in this diff)

`_to_bytes32` is applied only when *building* the request. The response to that same call lands in `ContractHandler._update_pending_list`, which filters on the **raw, uncoerced** id:

```python
# packages/valory/skills/task_execution/handlers.py:778
or req[RequestKey.REQUEST_ID_CAMEL.value] in on_chain_ids
```

`on_chain_ids` is `body["request_ids"]` — the `codec.decode(["bytes32[]"], ...)` output at `mech_marketplace/contract.py:438`, i.e. a list of `bytes`. But a task that has been popped and re-enqueued on timeout carries `requestId` as `int` (`behaviours.py:936-938` then `:1419`). `int in List[bytes]` is unconditionally `False`, so a retried task is pruned from `pending_tasks` on the next status-check tick — right after the contract confirmed it is still undelivered. The `is_offchain` retention guard at line 777 does not cover it.

Reproduced against the real response shape:

```
SURVIVORS: ['bytes']
AssertionError: int-shaped id was dropped: [{'requestId': b'\xaa'*32, 'is_offchain': False}]
assert 1 == 2
```

Before this PR the batch call returned `code=500` whenever any id was `int`, so `_update_pending_list` never ran and nothing was evicted. This PR is correct to fix the encode side — but it converts a loud no-op into a silent eviction, logged only as `Pending list updated via status check. N -> M`.

**Observable effect is a churn loop, not permanent loss.** `LedgerApiHandler.handle` resets `from_block = block_number - from_block_range` (50 000, `skill.yaml:102`) on every block read (`handlers.py:857-859`), so a dropped request is re-ingested from events as `bytes`, popped to `int`, times out, is requeued, and is dropped again — with `enqueued_at_local` and the local retry counter reset each cycle, so `timeout_limit` may never terminate it. Permanent loss only once the request ages past the ~50k-block (~3 day on Gnosis) lookback.

**Fix:** normalize both sides at `handlers.py:778`. `handlers.py` already imports from `behaviours.py` (line 64) and `behaviours.py` does not import `handlers`, so reusing `_to_bytes32` introduces no cycle.

Happy for this to land as a follow-up PR rather than growing this one — the line is not in this diff and the fix here is independently correct. Flagging it because merging this without it changes production behaviour on affected marketplace mechs.

## The existing test for that handler masks it

`tests/test_handlers.py:5952` uses `body = {REQUEST_IDS: [200]}` against `{"requestId": 200}` — `int` on both sides. It passes only because both are ints, never exercises the real `bytes32[]` -> `bytes` shape, and would pass whether or not the bug above exists. The follow-up should add a case seeding `on_chain_ids` as `bytes` against a mixed int/bytes queue.

## Checked and ruled out

- **No unbounded-int crash.** `int(value).to_bytes(32, "big")` raises `OverflowError` on a negative or `>= 2**256` value, and `_update_pending_tasks` is called from `act()` with the AEA default `propagate` exception policy. But off-chain `request_id` is already bounded by `_reject_unless_uint256_decimal` (`handlers.py:2131`, `MAX_REQUEST_ID = 2**256-1`) and the regex forbids a sign, and on-chain ids are always exactly-32-byte ABI decodes. No reachable trigger.
- **No serialization problem.** `ContractApiMessage.Kwargs` uses `DictProtobufStructSerializer`, which raises `UnicodeDecodeError` on non-UTF-8 bytes — but only when an envelope is actually encoded, which does not happen for the in-process ledger connection. `task_submission_abci/behaviours.py:1291` already passes `data=bytes.fromhex(...)` through the same path in production.

Inline notes below are non-blocking.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Confirmed and fixed. The fix went through two rounds: cdecd796 added the _to_bytes32 helper accepting int, bytes, or bytearray; bef6f7e1 hardens it to raise on len(bytes) > 32 rather than silently truncate (per Ojuswi's nit). The comprehension at behaviours.py:1013 now goes through the helper. Test coverage was widened to seed both shapes (parametrized case gained bytes32-fresh-marketplace-ingest and bytes32-zero variants), and mutation-verified: reverting to the int-only line makes the bytes cases fail with the exact ValueError from production.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Fixed. The parametrized case now includes both bytes32-fresh-marketplace-ingest (b'\xaa'*32) and bytes32-zero shapes alongside the int cases, plus a mixed-int-and-bytes variant covering the realistic queue where fresh ingest (bytes) and re-enqueued (int) sit side by side. Also added test_to_bytes32_raises_on_oversized_bytes to pin the new raise-on-oversize behaviour.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Confirmed and fixed via shared helper. handlers.py now imports _to_bytes32 from behaviours (no cycle — behaviours doesn't import handlers) and normalises both sides at the membership test: on_chain_ids becomes a set of 32-byte bytes, and each req[requestId] is also coerced through _to_bytes32 before the in check. Same helper on both ends, so the drift you flagged can't re-open silently.

New test test_update_pending_list_retains_retried_int_task_matched_by_bytes_response in test_handlers.py seeds the exact production shape (retried task with int requestId, response body with the same id as bytes32) and asserts survival. Mutation-verified: reverting the fix drops the task with survivors == [] and the assert fails.

DIvyaNautiyal07: Fixed. The comprehension now carves out is_offchain=True entries, mirroring the response-side carve-out in _update_pending_list, and the empty-list guard short-circuits an all-off-chain queue before a dialogue is created (avoiding an empty bytes32[] batch request):

```python
pending_tasks_request_ids = [
    _to_bytes32(t["requestId"])
    for t in self.pending_tasks
    if not t.get("is_offchain")
]
if not pending_tasks_request_ids:
    return
```

Covered by test_update_pending_tasks_carves_out_offchain_tasks: one mixed-shape queue that emits only the on-chain id, plus a follow-up all-off-chain case asserting the dialogue never fires.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Trimmed. The _to_bytes32 docstring is now the one-line summary + :param:/:return: + :raises: (the new raise-on-oversize).

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Good catch. Changed to raise ValueError on len(bytes) > 32 rather than silent tail-truncation via [-32:]. Unreachable today but converts a future shape drift into a loud failure. Pinned by test_to_bytes32_raises_on_oversized_bytes.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Removed. The call site is now just `_to_bytes32(t["requestId"])`; git blame lands on this commit + PR for anyone tracing.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Trimmed test docstring to one-line summary. Added test_to_bytes32_raises_on_oversized_bytes for the oversized-bytes case per your suggestion.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Removed. Both test bodies now use `kwargs.body["request_ids"]` directly.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Done. Folded the mixed-shape case into the parametrized case as one extra pytest.param with a list-valued `seeded` (and `expected_ints` for the round-trip check), dropping the duplicate stub. The new test_update_pending_tasks_carves_out_offchain_tasks is separate because it exercises a different code branch (F4 carve-out + empty-list guard).

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -280,6 +280,21 @@ def _iso_z(dt: datetime) -> str:
...
+
+    :param value: the raw ``requestId`` (``int`` or ``bytes``/``bytearray``).
+    :return: a ``bytes`` value of length exactly 32.
+    :raises ValueError: if a ``bytes`` value longer than 32 bytes is passed.
+    """
+    if isinstance(value, (bytes, bytearray)):
+        raw = bytes(value)
+        if len(raw) > 32:
+            raise ValueError(f"requestId is {len(raw)} bytes, expected <= 32")
+        return raw.rjust(32, b"\x00")
```
jmoreira-valory: 🔵 **F6 · LOW — Two 32-byte coercers with opposite padding now live in the same skill**

`_to_bytes32` left-pads a short `bytes` (`raw.rjust(32, b"\x00")`), while `utils/request_id.py:61 `_encode_bytes32`` right-pads the same input (`value + b"\x00" * (32 - len(value))`). Both are correct for their own job — a request id read as a big-endian uint256 vs. ABI `bytes32` padding — but neither name says which convention it implements, and they now sit in the same skill under near-synonymous names. A future maintainer reaching for the wrong one gets a silently different request id, not an error.

The int branch is also a duplicate: `utils/request_id.py:54 `_encode_uint256`` is already `value.to_bytes(32, "big")` with a range guard (which `_to_bytes32` lacks — a negative or `>= 2**256` id raises a bare `OverflowError` here instead of the module's own `ValueError`).

Related: `handlers.py:64-67` now imports the underscore-private `_to_bytes32` from `behaviours`, pulling the whole behaviour module for a four-line pure function. `utils/request_id.py` is already the declared home for exactly this ("Pure functions. No I/O, no ledger calls") and `handlers.py:79` already imports from it.

Suggest moving the helper there under a name that states the convention (`to_request_id_bytes32`), delegating to the two existing encoders, and importing it from `utils.request_id` on both sides. Non-blocking, but it closes the ambiguity while the context is fresh.

packages/valory/skills/task_execution/tests/test_handlers.py
```diff
@@ -5944,17 +5944,50 @@ def test_update_pending_list_preserves_offchain_tasks(
...
     ch._update_pending_list(body)
     survivors = [
         t["requestId"] for t in handler_context.shared_state[hmod.PENDING_TASKS]
     ]
-    assert survivors == [100, 200]
+    assert survivors == [100, onchain_rid]
+
+
+def test_update_pending_list_retains_retried_int_task_matched_by_bytes_response(
+    handler_context: Any, monkeypatch: Any
```
jmoreira-valory: ⚪ **F7 · NIT — New handler test takes a monkeypatch fixture it never uses**

`test_update_pending_list_retains_retried_int_task_matched_by_bytes_response` declares `monkeypatch: Any` and documents it in the docstring, but the body only builds a `ContractHandler` and seeds `shared_state` — nothing is patched. Drop the parameter and its `:param monkeypatch:` line so darglint stays happy.

jmoreira-valory (changes requested):
### Review round 2 — ⚠️ Request changes

**What this does.** The marketplace status check in `task_execution` batches the pending request ids into a `bytes32[]` ABI slot, and newer `eth_abi` refuses to encode an `int` there, so the call had been returning a contract-api 500 every ~10 minutes and the pending-queue reconciliation was silently dead. This round adds a shared `_to_bytes32` helper that accepts both the `bytes` shape (fresh marketplace ingest) and the `int` shape (a task re-queued after a timeout), uses it on both the request side and the response-side membership test in `_update_pending_list`, and stops sending off-chain-only tasks to the on-chain batch call. Tests were widened to cover both shapes plus the mixed queue.

**All four of my round-1 findings landed, and landed properly** — I re-verified each against HEAD rather than against the reply text:

- **F1 (HIGH)** — `_to_bytes32` (`behaviours.py:283`) now takes the `bytes`/`bytearray` branch, so a fresh-ingest `bytes32` id no longer hits `int()`.
- **F2 (MEDIUM)** — the parametrized case covers `bytes32-fresh-marketplace-ingest`, `bytes32-zero` and `mixed-int-and-bytes`, not just ints.
- **F3 (MEDIUM)** — `_update_pending_list` normalises *both* sides of the membership test through the same helper (`handlers.py:780,785`), so a retried int-shaped id can't be pruned against a set of bytes. The new `test_update_pending_list_retains_retried_int_task_matched_by_bytes_response` pins exactly that.
- **F4 (LOW)** — off-chain entries are carved out of the outgoing `bytes32[]`, mirroring the response-side carve-out.

Also good: the `rjust`→raise change on oversized bytes is the right call (loud drift beats silent truncation); CI is fully green (22/22, including `lock_check` on all three OSes, so the task_execution → mech_abci / task_submission_abci / agent / service CID cascade checks out); and the tests read as genuine mutation tests rather than assertions written around the implementation.

The one blocker below is a side effect of *where* the new empty-list guard was placed, not of the fix itself — a one-hunk move.

**Findings** — 3 · 🟡 1 Medium · 🔵 1 Low · ⚪ 1 Nit

| F5 | 🟡 MEDIUM | `packages/valory/skills/task_execution/behaviours.py:1022` | **All-off-chain pending queue re-logs the status check on every behaviour tick** |
| F6 | 🔵 LOW | [packages/valory/skills/task_execution/behaviours.py:294](https://github.com/valory-xyz/mech/blob/bef6f7e1425bd84cf2d19a9f7473cfffe6f30667/packages/valory/skills/task_execution/behaviours.py#L294) | **Two 32-byte coercers with opposite padding now live in the same skill** |
| F7 | ⚪ NIT | [packages/valory/skills/task_execution/tests/test_handlers.py:5962](https://github.com/valory-xyz/mech/blob/bef6f7e1425bd84cf2d19a9f7473cfffe6f30667/packages/valory/skills/task_execution/tests/test_handlers.py#L5962) | **New handler test takes a monkeypatch fixture it never uses** |

---

- 🟡 **F5 · MEDIUM — All-off-chain pending queue re-logs the status check on every behaviour tick** The new `if not pending_tasks_request_ids: return` (line 1030) sits **after** the `logger.info` and **before** `self.last_status_check_time = time.time()` (line 1050). So the 10-minute throttle never re-arms on that path.

`TaskExecutionBehaviour` is a `SimpleBehaviour` with no `tick_interval` override in `skill.yaml`, so it inherits open-aea's 0.001s default and `act()` — which calls `_update_pending_tasks()` unconditionally — fires hundreds of times per second. Whenever the pending queue is non-empty but **entirely** `is_offchain=True`, every one of those ticks passes the stale-timestamp guard, emits

```
Checking status change for N pending tasks...
```

and then returns having checked nothing.

That state is routine, not exotic: `_execute_task` pops one task at a time, so during a tool run (an LLM call, tens of seconds) the rest of an off-chain backlog sits in `pending_tasks`, and `_update_pending_tasks` still runs because `in_flight_req` is `False` while the executor future is pending. Before this PR the same state took the dialogue path and stamped the timestamp, so the line fired once per interval. This is a fresh INFO-level flood on exactly the log surface the PR's own test plan depends on ("watch affected marketplace mechs for cessation of `ExactLengthBytesEncoder` events").

Second, smaller problem in the same three lines: the message reports `pending_tasks_count` (the whole queue) while only the on-chain subset is now checked, so on a mixed queue the number is wrong.

Both go away by moving the log below the guard and counting the filtered list — no timestamp change, so a newly-arriving on-chain task still gets checked promptly rather than waiting out a freshly-stamped interval:

```python
        pending_tasks_request_ids = [
            _to_bytes32(t["requestId"])
            for t in self.pending_tasks
            if not t.get("is_offchain")
        ]
        if not pending_tasks_request_ids:
            return

        self.context.logger.info(
            f"Checking status change for {len(pending_tasks_request_ids)} "
            "pending tasks..."
        )
```

Worth a test too: the existing `test_update_pending_tasks_carves_out_offchain_tasks` already drives the all-off-chain branch twice, so asserting the logger was not called on the second pass is nearly free.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1007,7 +1022,13 @@ def _update_pending_tasks(self) -> None:
...
             f"Checking status change for {pending_tasks_count} pending tasks..."
         )
-        pending_tasks_request_ids = [t["requestId"] for t in self.pending_tasks]
+        pending_tasks_request_ids = [
+            _to_bytes32(t["requestId"])
+            for t in self.pending_tasks
+            if not t.get("is_offchain")
+        ]
+        if not pending_tasks_request_ids:
+            return
```
OjusWiZard, line 1016: **This early return bypasses the 10-minute throttle.**

It sits *after* the `Checking status change for N pending tasks...` log at line 1022, but *before* `self.last_status_check_time = time.time()` at the end of the method. So when the queue is non-empty but entirely off-chain — the exact state this carve-out adds support for — `last_status_check_time` never advances, the interval gate at line 1014 stays open, and `act()` re-enters on every tick.

AEA's default act period is `0.05s` (`aea_builder.py:321`, and there is no `period` override in `aea-config.yaml`), so that is ~20x/second: roughly 1.7M log lines/day instead of one per 600s, for as long as no on-chain task is pending.

Note this is **not** equivalent to the pre-existing `pending_tasks_count == 0` return at line 1019 — that one returns before the log line, so it is silent.

Fix:

```python
if not pending_tasks_request_ids:
    self.last_status_check_time = time.time()
    return
```

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1607,6 +1607,137 @@ def test_update_pending_tasks_no_pending_tasks(
...
+    captured.clear()
+    shared_state[beh_mod.PENDING_TASKS] = [
+        {"requestId": 1, "is_offchain": True},
+        {"requestId": 2, "is_offchain": True},
+    ]
+    behaviour.last_status_check_time = 0.0
+    behaviour._update_pending_tasks()
+    assert (
+        captured == {}
+    ), "contract_dialogues.create must not fire for all-offchain queue"
```
OjusWiZard: **This second assertion passes vacuously — it does not test the carve-out.**

The first `_update_pending_tasks()` call above sets `params.in_flight_req = True` (`behaviours.py:1052`). This second call therefore returns immediately at the `if self.params.in_flight_req: return` guard and never reaches the carve-out at all, so `captured == {}` is true for the wrong reason.

Probed directly:

```
AFTER 1st CALL: in_flight_req = True
>>> reached carve-out? NO - short-circuited on in_flight_req
```

And mutation-tested by deleting the `if not pending_tasks_request_ids: return` guard from `behaviours.py` entirely:

```
MUTATION: removed `if not pending_tasks_request_ids: return`
1 passed, 174 deselected
RESULT: TEST STILL PASSES -> does not detect the mutation
```

So that early return currently has zero coverage. One line fixes it — reset the flag before the second call:

```python
captured.clear()
params_stub.in_flight_req = False
shared_state[beh_mod.PENDING_TASKS] = [...]
```

The underlying behaviour is correct — I confirmed an all-off-chain queue opens no dialogue once `in_flight_req` is reset. Only the test is unsound.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -280,6 +280,21 @@ def _iso_z(dt: datetime) -> str:
...
+    """Coerce a marketplace ``requestId`` to a 32-byte big-endian value.
+
+    :param value: the raw ``requestId`` (``int`` or ``bytes``/``bytearray``).
+    :return: a ``bytes`` value of length exactly 32.
+    :raises ValueError: if a ``bytes`` value longer than 32 bytes is passed.
+    """
+    if isinstance(value, (bytes, bytearray)):
+        raw = bytes(value)
+        if len(raw) > 32:
+            raise ValueError(f"requestId is {len(raw)} bytes, expected <= 32")
```
OjusWiZard: Good change — much better than silently truncating.

One thing to be aware of, non-blocking: this `raise` now sits in a call chain with no exception handling at any layer. `_to_bytes32` is called from `_update_pending_list`, and `ContractHandler` extends the plain `Handler` base, not `AbstractResponseHandler` (which does `try/except` around its callbacks). An escape propagates through `handle_wrapper` into the agent loop's `_execution_control` and halts the whole agent, not just this skill.

Not reachable today — every `bytes` id is an exactly-32-byte ABI decode and every `int` id is bounded — so this is a latent hazard rather than a bug. But it means one malformed queue entry becomes a dead agent instead of a skipped task. A `try/except Exception` around the body dispatch in `ContractHandler.handle` that logs and leaves `pending_tasks` untouched would match how the IPFS/CID paths in the same file already guard.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -770,12 +773,16 @@ def _update_pending_list(self, body: Dict[str, List]) -> None:
         :param body: the on-chain status response body.
         """
         before = len(self.pending_tasks)
-        on_chain_ids = body[BodyKey.REQUEST_IDS.value]
+        # ``on_chain_ids`` is ``List[bytes]`` from ``codec.decode(["bytes32[]"], ...)``;
+        # ``req["requestId"]`` may be ``int`` on tasks re-queued after execution.
+        # Normalise both to 32-byte bytes so a retried task doesn't get pruned
+        # against a set of a different type.
```
OjusWiZard: This reintroduces the kind of comment that was trimmed last round. The new test docstring in `test_handlers.py` already explains this int/bytes mismatch in full, and `git blame` on line 780 resolves to this commit and PR, which explain it again. Two of those three are enough — suggest deleting the comment and keeping the test as the explanation.

packages/valory/skills/task_execution/tests/test_handlers.py
```diff
@@ -5944,17 +5944,50 @@ def test_update_pending_list_preserves_offchain_tasks(
...
+) -> None:
+    """A retried task carrying ``requestId`` as ``int`` is retained when the on-chain response returns the same id as ``bytes32``.
```
OjusWiZard: The summary line runs well past one line, and the paragraph below it is a full narrative reconstruction of the data flow (ingest -> rewrite -> re-enqueue -> compare) — the same material the commit message and PR description carry.

Suggest keeping the summary short and dropping the paragraph:

```python
"""A retried ``int`` requestId is retained against a ``bytes32`` response.

:param handler_context: pytest fixture, mech HTTP handler test context.
:param monkeypatch: pytest fixture, per-test monkeypatch helper.
"""
```

Also: `monkeypatch` is in the signature but unused in the body — drop the parameter and its `:param:` line.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -1607,6 +1607,137 @@ def test_update_pending_tasks_no_pending_tasks(
...
+    behaviour: Any,
+    params_stub: Any,
+    shared_state: Dict[str, Any],
+    monkeypatch: Any,
+) -> None:
+    """Off-chain tasks are filtered out of the on-chain batch status check.
+
+    Off-chain request ids do not exist on chain until settlement, so
+    querying the marketplace for their status is a wasted call and
+    (until F4) enlarged the outgoing ``bytes32[]``. Also verify the
```
OjusWiZard: `(until F4)` is an internal review-tracking reference — it will mean nothing to anyone reading this in six months, and it is already stale if F4 landed. Suggest `... is a wasted call and enlarges the outgoing ``bytes32[]``.` and dropping the marker.

Same docstring: the rationale paragraph is what the commit message is for; the one-line summary above it already states the behaviour under test.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -61,7 +61,10 @@
 from packages.valory.protocols.kv_store.message import KvStoreMessage
 from packages.valory.protocols.ledger_api import LedgerApiMessage
 from packages.valory.skills.abstract_round_abci.handlers import AbstractResponseHandler
-from packages.valory.skills.task_execution.behaviours import PREDICT_API_EVENTS
+from packages.valory.skills.task_execution.behaviours import (
+    PREDICT_API_EVENTS,
+    _to_bytes32,
```
OjusWiZard: Non-blocking: `_to_bytes32` is now shared API across two modules but is still named module-private and reached into from outside. No import cycle (behaviours does not import handlers), so this is hygiene only — but `task_execution/utils/` already exists and is imported here for exactly this kind of shared helper (`preimage`). Worth relocating and dropping the underscore.

OjusWiZard (dismissed):
### Approving again on `bef6f7e1`. Everything from the last round is fixed and verified.

Both sides of the membership test are normalised (`handlers.py:780/785`) with a real regression test; `test_handlers.py` now uses the true `bytes32` response shape; `_to_bytes32` raises instead of truncating, with a test; the comment bloat is trimmed; the `hasattr` fallback is gone and the mixed-shape case is folded into the parametrize. Verified locally: **903 tests pass**, and `black` / `isort` / `flake8` / `darglint` / `mypy` / `pylint` are all clean.

I confirmed the batch contract's semantics directly from the bytecode in `BATCH_REQUESTID_STATUS_DATA` rather than from its docstring: selector `0x45d07664` is called per id, the result is compared against `1` and `2` of a 4-member enum, and only matching ids are copied into the output array. So it returns the **still-undelivered subset** of what you send. That matters for the first item below.

Two things below are new regressions in the off-chain carve-out (which was not part of the previous round's requests) — both are inline. Two more are out-of-diff and are here in the body.

---

## Not in this diff: the empty-response case still silently no-ops

`packages/valory/skills/task_execution/handlers.py:688`

```python
if body.get(BodyKey.REQUEST_IDS.value):
    self._update_pending_list(body)
```

An empty list is falsy. Given the contract returns only the undelivered subset, a batch in which **every** pending on-chain task has been delivered comes back as `[]` -> the guard is falsy -> `_update_pending_list` is skipped -> nothing is pruned. The delivered tasks stay queued, get popped by `_execute_task` (`behaviours.py:925-930`, which has no already-delivered check), and are re-executed and re-delivered — burning a full tool-execution cycle and a consensus round on work the chain has already settled.

This is pre-existing and untouched by this PR, but it is the same family as the int/bytes bug you just fixed: both are cases where the status check silently does nothing. `fetch_batch_request_id_status` is the only callable that feeds this body key (`has_priority_passed`'s copy is `.pop()`-consumed inside `get_undelivered_reqs` and never surfaces), so the fix is unambiguous:

```python
if BodyKey.REQUEST_IDS.value in body:
```

No test currently covers the all-delivered case. Fine as a follow-up — flagging it because the same reasoning that motivated this PR applies.

## Not in this diff: `pending_tasks_count` now overstates what is checked

`packages/valory/skills/task_execution/behaviours.py:1017` still counts the whole queue, but only the on-chain subset is sent after the carve-out. A queue of 3 off-chain + 2 on-chain logs `"Checking status change for 5 pending tasks..."` while 2 ids go out. Cosmetic, but it compounds the throttle issue flagged inline at line 1030.

## Checked and ruled out

`autonomy packages lock --check` fails in my local checkout, but every mismatch is a **third-party** package (`ledger`, `abstract_round_abci`, `http_server`, `p2p_libp2p_client`, `registration_abci`, `reset_pause_abci`, `termination_abci`, `transaction_settlement_abci`). None of this PR's first-party packages are mismatched, and the declared hashes agree between `packages.json` and `aea-config.yaml`. That is a stale local registry, not a defect here — CI syncs from IPFS before checking.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: F6 addressed in 02320274. Moved the helper to utils/request_id.py as `to_request_id_bytes32`, delegating to `_encode_uint256` for the int path so the uint256 range guard is shared (a negative or out-of-range int now raises `ValueError` with the module's own message rather than a bare `OverflowError`). The bytes path is strict: exactly 32 bytes or raise, so it doesn't silently pad or truncate — the ambiguity vs `_encode_bytes32`'s ABI-padding convention no longer applies. Both behaviours.py and handlers.py import the public name from utils, dropping the underscore-private cross-module reach. Coverage: `test_to_request_id_bytes32_raises_on_wrong_length_bytes` and `test_to_request_id_bytes32_raises_on_out_of_range_int`.

packages/valory/skills/task_execution/tests/test_handlers.py
DIvyaNautiyal07: Fixed. Dropped the unused monkeypatch parameter and trimmed the docstring to a one-line summary + :param: for the handler_context fixture.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Fixed exactly as suggested. Filter first, early-return with `self.last_status_check_time = time.time()` stamped, then log with the filtered count. `test_update_pending_tasks_carves_out_offchain_tasks` was also strengthened to assert the stamp — mutation-verified by removing the stamp line, the assertion fires with `0.0 == 601.0`.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Caught. Reset `params_stub.in_flight_req = False` before the second call so the carve-out branch actually runs, and added the throttle-stamp assertion so removing the fix now fails the test (verified via mutation).

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Good latent point. The raise is now unreachable in practice because `to_request_id_bytes32` accepts only exactly-32-byte bytes or in-range uint256, and both call sites in the queue only produce those shapes (on-chain ABI decodes are exactly 32; execution-pop is `int.to_bytes(32, 'big')`). Skipping the try/except around `ContractHandler.handle` in this PR to keep it scoped, but noted for a follow-up if this ever grows callers that don't guarantee the shape.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Deleted. The test docstring in test_handlers.py carries the same info + git blame + PR description; the comment was the third copy.

packages/valory/skills/task_execution/tests/test_handlers.py
DIvyaNautiyal07: Trimmed to a one-line summary + :param: for handler_context. Dropped the unused monkeypatch parameter and its docstring line.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Dropped the `(until F4)` marker and trimmed the rationale paragraph.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Done — moved to utils/request_id.py as `to_request_id_bytes32` (public), dropping the underscore-private cross-module reach on both sides. Same helper used from behaviours.py and handlers.py, no cycle.

DIvyaNautiyal07:
Round 3 pushed at `02320274`. Both round-2 out-of-diff items are addressed:

**Empty-response guard (handlers.py:688)** — switched from `body.get(BodyKey.REQUEST_IDS.value)` to `BodyKey.REQUEST_IDS.value in body`, so an empty-list response (the legitimate "every id was delivered" answer per the batch contract semantics you traced from the bytecode) now runs the prune instead of silently skipping. The old test `test_contract_handler_doesnot_updates_pending_list_based_on_undelivered_request_status` was pinning the buggy behaviour — renamed to `test_contract_handler_prunes_pending_list_on_empty_undelivered_request_status` and the assertion inverted (`len(pending_tasks) == 0`). Mutation-verified: reverting to `.get(...)` fails the assertion with survivors `[100, 200]`.

**`pending_tasks_count` overstating** — folded into the F5 fix: the log now uses `len(pending_tasks_request_ids)` (the filtered count) after the carve-out, so a mixed 3-off/2-on queue logs "2 pending" instead of "5 pending".

Two more code changes not directly requested but adjacent:
- `to_request_id_bytes32` also raises `ValueError` on an int outside the uint256 range (delegates to `_encode_uint256` which has the guard already) — closes the theoretical unbounded-int / bytes-drift path you called out as unreachable-today, now unreachable-by-construction.
- All-off-chain skip now stamps `last_status_check_time` before returning (per your inline note), so `act()` doesn't re-enter every ~50ms and log-flood.

904 skill tests pass, all lints green, `autonomy packages lock --check` OK, packages pushed to IPFS.

OjusWiZard (approved):

jmoreira-valory (approved):
### Review round 3 — ✅ Approve

**What this does.** The mech's marketplace status check batches its pending request ids into a `bytes32[]` ABI slot, and newer `eth_abi` refuses to encode an `int` there, so the call had been returning a contract-api 500 every ~10 minutes and the pending-queue reconciliation was silently dead. This round finishes the cleanup: the coercer moves out of `behaviours.py` into `utils/request_id.py` as `to_request_id_bytes32` (sharing the uint256 range guard with `_encode_uint256`, and strict on the bytes path), the all-off-chain carve-out now stamps the throttle so it cannot log-flood, and the response gate flips from a truthiness check to a presence check so an empty "everything was delivered" answer actually prunes the queue.

Clearing my round-2 `CHANGES_REQUESTED`. All three blockers are fixed and verified at `02320274`:

- **F1 (HIGH)** - `to_request_id_bytes32` accepts both the fresh-ingest `bytes32` and the execution-pop `int` shape; no `int()` on a `bytes` id anywhere on the path.
- **F2 (MEDIUM)** - the parametrized test now seeds `bytes32` shapes (`b"\xaa" * 32`, all-zero) and a mixed int/bytes queue alongside the int cases.
- **F3 (MEDIUM)** - both sides of the membership test in `_update_pending_list` go through the same helper (`handlers.py:779` and `handlers.py:784`), pinned by `test_update_pending_list_retains_retried_int_task_matched_by_bytes_response`.
- **F4 / F6 / F7 (LOW/NIT)** - off-chain carve-out landed, the two opposite-padding coercers now sit side by side in `utils/request_id.py` where the naming distinction is visible, unused `monkeypatch` dropped.

Verified against source at HEAD rather than relying on the reply threads:

- `last_status_check` is a property over `last_status_check_time` (`behaviours.py:604-606`), so the new all-off-chain stamp at `behaviours.py:1015` really does advance the interval gate at `behaviours.py:1000`. @OjusWiZard's throttle-bypass finding is genuinely closed, and the second phase of `test_update_pending_tasks_carves_out_offchain_tasks` now resets `in_flight_req`, so it no longer passes vacuously.
- The `body.get(...)` -> `in body` flip is safe. I checked every callable this `ContractHandler` can receive a response for: `get_marketplace_undelivered_reqs` returns `{data, wait_for_timeout_tasks, timed_out_requests}`, `get_undelivered_reqs` returns `{data}`, `get_mech_type` / `get_mech_types` return `{mech_type}` / `{mech_types}`. Only `fetch_batch_request_id_status` returns a top-level `request_ids`, so the presence check cannot cross-fire on another response body.
- The "empty list means all delivered" claim holds at the contract level, not just in the docstring. Walking the `BatchRequestIdStatusData` creation-code path: it calls selector `0x45d07664` per id and appends the id only when the returned status enum is `1` or `2`, so the `bytes32[]` it returns is the undelivered subset and an empty array is a legitimate all-delivered answer. The new `test_contract_handler_prunes_pending_list_on_empty_undelivered_request_status` uses `()` for the empty case, which is the real `codec.decode` shape.
- No stale `_to_bytes32` references remain in either module or in the test files at HEAD, and the CID cascade is consistent: `task_execution` `bafybeifsbxo5...requ5q` matches across `packages.json`, `aea-config.yaml`, `mech_abci/skill.yaml` and `task_submission_abci/skill.yaml`, and `utils/request_id.py`'s fingerprint is bumped alongside the two modules.

CI is green at HEAD (22/22 check-runs successful, none failing or skipped).

The single item below is non-blocking and pre-dates this round's delta. I am recording it because this PR is what brings the code path back to life in production, so it stops being theoretical the day this merges - a follow-up issue is a perfectly good home for it rather than another round here.

**Findings** — 1 · 🔵 1 Low

| F8 | 🔵 LOW | `packages/valory/skills/task_execution/handlers.py:780` | **Task re-queued during the status-check round trip is pruned and never re-ingested** |

---

- 🔵 **F8 · LOW — Task re-queued during the status-check round trip is pruned and never re-ingested** This is the residual half of my round-2 F3. The shape mismatch is fixed; what remains is that `_update_pending_list` prunes against *the response*, not against *the set that was queried*, so anything that enters `pending_tasks` between the query going out and the reply landing is treated as delivered.

The window is reachable. `_update_pending_tasks` sets `in_flight_req = True` at `behaviours.py:1038` and returns. While that flag is set `_execute_task` does **not** short-circuit - `behaviours.py:869-883` still runs the deadline branch and can call `_handle_timeout_task()`, which re-appends the executing task at `behaviours.py:1405`. That id was never in the outgoing `request_ids`, so it cannot come back in the response, so this comprehension drops it.

Failure scenario: `pending_tasks = [A, B]`. Tick N pops `A` for execution and queries `[B]`. Mid-round-trip `A`'s handling deadline expires and `_handle_timeout_task` re-appends `A`. The reply `[B]` arrives and `A` is pruned. `from_block` has already advanced past `A`'s block (`handlers.py:740`), so the next marketplace poll will not re-ingest it: `A` is still undelivered on chain but gone from the queue permanently, and the requester's payment is stranded.

LOW because the window is one `eth_call` round trip against a 10-minute check interval, so it needs a deadline expiry to land inside roughly a second. But it is a silent, unrecoverable drop when it does, and this PR is what makes the path live again.

The cheap fix is to prune only against what was actually asked. Stash the outgoing ids when the query is built in `_update_pending_tasks`, then leave un-queried tasks alone here:

```python
queried = set(self.params.last_status_check_ids or ())
self.context.shared_state[PENDING_TASKS] = [
    req
    for req in self.pending_tasks
    if req.get(RequestKey.IS_OFFCHAIN.value)
    or (rid := to_request_id_bytes32(req[RequestKey.REQUEST_ID_CAMEL.value]))
    not in queried
    or rid in on_chain_ids
]
```

A task the status check never asked about is then retained instead of being inferred as delivered.

PR #487, opened 2026-09-08
feat: add Robinhood Chain (4663) ledger wiring

DIvyaNautiyal07:
## Summary

Adds Robinhood Chain (mainnet chain ID `4663`) to the agent's chain-aware ledger wiring, mirroring what #459 did for Optimism. This is the agent-side prerequisite for deploying a mech on Robinhood once the Mech Marketplace contracts are live there. No behaviour changes for existing chains.

## What changed

- `packages/valory/agents/mech/aea-config.yaml`: new `robinhood` entry under the `ledger` connection (`chain_id: 4663`, `poa_chain: false`, `eip1559`) and a `robinhood_ledger_rpc` param on the `task_execution` override.
- `packages/valory/services/mech/service.yaml`: for all four agent indexes, `robinhood_ledger_rpc` driven by `ROBINHOOD_LEDGER_RPC_<N>`, and the matching `ledger_apis.robinhood` block driven by `ROBINHOOD_LEDGER_RPC_<N>`, `ROBINHOOD_LEDGER_CHAIN_ID`, `ROBINHOOD_LEDGER_IS_POA_CHAIN`, `ROBINHOOD_LEDGER_PRICING`.
- `packages/valory/skills/task_execution/models.py`: `robinhood_ledger_rpc` Params field.
- `packages/valory/skills/task_execution/handlers.py`: `ChainId.ROBINHOOD = 4663`. Without it `_get_ledger_settings` returns `Unsupported chain: robinhood` even with the RPC configured.
- `packages/valory/skills/task_execution/tests/test_handlers.py`: the supported-chains table test now also covers `optimism` and `robinhood`.
- `autonomy packages lock`; packages pushed to the IPFS registry.

## Deployment notes

- A Robinhood deployment sets `DEFAULT_CHAIN_ID=robinhood`, `ROBINHOOD_LEDGER_RPC_<N>`, and `ROBINHOOD_LEDGER_CHAIN_ID=4663`. Gas token is ETH.
- Keep `USE_ACN=false` on Robinhood: the `service_registry` package's expected-address map only knows chain IDs 1, 5, 100, 137 and 31337, and it is consulted on the ACN startup path.
- `mech-predict` (what production mechs run) needs to pick these packages up separately before a Robinhood mech can be deployed from `mech-deployments`.

## Verification

Run locally from `.github/workflows/common_checks.yaml`: `uv lock --check`, `check-hash`, `check-packages`, `check-doc-hashes`, `check-dependencies`, `check-third-party-hashes`, copyright, `black-check`, `isort-check`, `flake8`, `mypy`, `pylint`, `darglint`, `check-abci-docstrings`, `check-abciapp-specs`, `check-handlers`, `bandit`, `gitleaks`, and the `task_execution` handler and model test modules (199 passed).


OjusWiZard (approved):
Approving. Small and correct. I verified the chain facts against the live network rather than the docs, and checked the two Python lines against the code path that consumes them.

### Verified

| claim | result |
|---|---|
| `ChainId.ROBINHOOD = 4663` | matches the live chain (`eth_chainId` → `0x1237`), and the enum's existing uppercase-name convention |
| `poa_chain: false` | correct — 4663's `extraData` is 32 bytes, so no POA middleware is needed |
| `eip1559` | correct — blocks carry `baseFeePerGas`. Worth knowing operationally: `eth_maxPriorityFeePerGas` returns `0` on Nitro, which is expected rather than a misconfiguration |
| param default `""` | matches all four siblings (`gnosis`, `polygon`, `base`, `optimism`) — consistent, and fails closed |
| env-var naming | consistent across all four agent indexes: per-agent `ROBINHOOD_LEDGER_RPC_<N>`, shared `CHAIN_ID` / `IS_POA_CHAIN` / `PRICING`, and `robinhood_ledger_rpc` matches the `getattr(params, f"{chain}_ledger_rpc")` lookup exactly |
| lock / hashes | CI green across 22 jobs, including `lock_check` on Linux, macOS and Windows |

**The reasoning in the description is precisely right**, which is worth saying because it is the kind of claim that is usually approximate. `_get_ledger_settings` checks the RPC *first* and returns `Missing RPC config for chain: …`; only once that passes does it reach `ChainId[chain.upper()]`. So with the RPC configured you would indeed land on the `KeyError` and get `Unsupported chain: robinhood`. Both additions are genuinely required, and for different reasons.

### One advisory — the new test does not exercise the `models.py` change

`test_get_ledger_settings_supported_chains` sets the attribute dynamically on the mock:

```python
setattr(handler_context.params, f"{chain}_ledger_rpc", rpc_attr)
```

so it passes whether or not `models.py` declares `robinhood_ledger_rpc`. It proves the enum member and the handler logic; it does not prove the Params wiring that makes the RPC reachable at runtime. `test_models.py` has no `*_ledger_rpc` assertions either, so that line is untested for **every** chain, not just this one.

Low severity — the field mirrors four siblings exactly, and a missing one fails closed with a clear message. Not worth blocking on. But "the supported-chains table test now also covers `optimism` and `robinhood`" is true of the handler path only, and it would be easy to read as end-to-end.

### One deployment note worth promoting out of the PR body

**`USE_ACN=false` on Robinhood.** The `service_registry` package's expected-address map only knows chain IDs 1, 5, 100, 137 and 31337, and it is consulted on the ACN startup path. No CI check catches that — it is a runtime foot-gun, and it should land in the deployment runbook rather than living only in a merged PR description.

Same for the `mech-predict` dependency: production mechs need to pick these packages up separately before a Robinhood mech can be deployed from `mech-deployments`. Both are correctly stated here; they just need a home outside this PR.

packages/valory/skills/task_execution/skill.yaml
```diff
@@ -109,6 +109,7 @@ models:
       mech_to_max_delivery_rate: {}
       num_agents: 4
       optimism_ledger_rpc: http://localhost:8545
+      robinhood_ledger_rpc: http://localhost:8545
```
jmoreira-valory, line 112: ⚪ **F1 · NIT — New param breaks the alphabetical ordering of models.params.args**

The `params.args` block is otherwise strictly alphabetical (`num_agents` -> `optimism_ledger_rpc` -> `polling_interval` -> `polygon_ledger_rpc` -> `serious_slash_unit_amount`). `robinhood_ledger_rpc` was inserted directly after `optimism_ledger_rpc`, which puts it before `polling_interval` and `polygon_ledger_rpc`.

Purely cosmetic today -- no linter enforces it and CI is green -- but this file's hash feeds the `task_execution` CID, which cascades into `mech_abci`, `task_submission_abci`, the agent and the service. If any tooling ever re-serialises this config with sorted keys, that no-op reorder churns five CIDs. It also keeps the diff clean against the equivalent block in `mech-predict`, which has to pick these packages up separately.

Move the line two positions down so it sits between `polygon_ledger_rpc` and `serious_slash_unit_amount`:

```yaml
      optimism_ledger_rpc: http://localhost:8545
      polling_interval: 30.0
      polygon_ledger_rpc: http://localhost:8545
      robinhood_ledger_rpc: http://localhost:8545
      serious_slash_unit_amount: 8000000000000000
```

jmoreira-valory (approved):
### Review round 1 — ✅ Approve

**What this does.** This teaches the mech agent about a fifth EVM chain, Robinhood Chain (id 4663), so a mech can eventually be deployed there. It is purely additive plumbing: a `robinhood` entry in the ledger connection's `ledger_apis` (agent + all four service agent indexes), a `robinhood_ledger_rpc` param threaded through skill.yaml / aea-config.yaml / service.yaml, and the two Python lines the runtime actually reads -- `ChainId.ROBINHOOD = 4663` and the matching `Params` field. Nothing changes for the existing chains; the mech only reaches this wiring when `DEFAULT_CHAIN_ID=robinhood`.

I verified the wiring against source at HEAD rather than the PR body, and it holds up end to end:

- `_get_ledger_settings` (`handlers.py:3829-3853`) resolves the RPC via `getattr(params, f"{chain}_ledger_rpc")` and the id via `ChainId[chain.upper()]`. Both halves are supplied, and the ordering claim in the description is right -- the RPC check runs first, so with the RPC configured you land on the `KeyError` branch.
- `task_submission_abci/behaviours.py` passes `chain_id=self.params.default_chain_id` (the **alias**) into every `get_contract_api_response`, which is exactly what the new `ledger_apis.robinhood` key is keyed on. So the on-chain path is covered too, not just the offchain HTTP path.
- 3-layer config parity is complete: `task_execution/skill.yaml` default, agent `aea-config.yaml` override, and service.yaml for all four agent indexes -- every new value in `${ENV:type:default}` form, never a bare literal.
- CID cascade is complete and consistent (`packages.json`, agent `aea-config.yaml`, `mech_abci/skill.yaml`, `task_submission_abci/skill.yaml`), and CI is 22/22 green including `lock_check` on Linux, macOS and Windows.
- The runtime-footgun defaults are all safe as shipped: `share_tm_config_on_startup` (`USE_ACN`), `use_termination` and `use_slashing` are each `${...:bool:false}`, so the `service_registry` expected-address map the description warns about (chain ids 1/5/100/137/31337 only) is not on the startup path unless an operator opts in. And a forgotten RPC degrades to `Missing RPC config for chain: robinhood` rather than crashing.
- No hardcoded chain-alias table exists anywhere else in the repo -- `mech_events_chain_id` deliberately carries the EIP-155 integer separately, so there is no second place to keep in sync.

Two non-blocking notes I am deliberately **not** filing as findings:

- **`CLAUDE.md:7` is now stale** -- "Runs on Gnosis chain (and Polygon, Base, Optimism)". A one-word addition keeps the repo's own agent-context file honest. The `README.md` supported-chains table is correctly left alone: it lists live marketplace deployments, and Robinhood has none yet.
- @OjusWiZard already covered the test-depth point (the parametrized test `setattr`s the attribute onto the fixture, so it proves the enum but not the `Params` wiring) and the `USE_ACN` / `mech-predict` runbook items, so I am not re-raising any of those.

**Findings** — 1 · ⚪ 1 Nit

| F1 | ⚪ NIT | [packages/valory/skills/task_execution/skill.yaml:112](https://github.com/valory-xyz/mech/blob/d46f87307ea71b05506763786cb9dc1cbdbb204e/packages/valory/skills/task_execution/skill.yaml#L112) | **New param breaks the alphabetical ordering of models.params.args** |

PR #488, opened 2026-09-09
fix(task_execution): real timestamps and tool latency in predict-api events

DIvyaNautiyal07:
## Summary

Every live row written by the mechs to the predict-api data lake has `requested_at == executed_at == delivered_at` (100% of `mech_offchain` rows, 99.2% of `mech_onchain` rows since Aug 25) and `execution_latency_ms = NULL`, so request-to-result latency cannot be derived from the lake. Three causes in `_build_predict_api_event` (`packages/valory/skills/task_execution/behaviours.py`):

1. `executed_at` is stamped as an ISO string in `_handle_done_task`, but the builder only accepted a Unix number and substituted the build time otherwise, so `executed_at` always equalled `delivered_at`.
2. `requested_at` looked for a `datetime` / `requested_at` field in the request body; off-chain bodies carry none, so it fell back to `executed_at`.
3. `execution_latency_ms` was hard-coded to `None` although the mech computes it into the response metadata a few lines earlier.

## Changes

- `requested_at` is now the mech's receipt time, the `enqueued_at_local` stamp set by the handler on every queue path: on-chain requests for this mech in `filter_requests` (already there), timed-out requests from other mechs when they are loaded from the contract or classified in `filter_requests` (new, so step-in deliveries get a real value too), and off-chain requests in `_enqueue_offchain_request` (new; the key is reserved, so a client-supplied value is dropped and logged). A body timestamp is used only for tasks enqueued before the stamp existed, then `executed_at`, and that last fallback logs a warning.
- `executed_at` keeps the ISO value from the response; Unix numbers are still accepted.
- `execution_latency_ms` is taken from the response metadata when it is a finite non-negative number; anything else is logged and written as NULL.
- New `_parse_timestamp` helper used for all three: returns `None` instead of raising for anything unusable (wrong type, NaN/inf, out-of-range Unix values, malformed ISO strings, and ISO strings whose UTC conversion overflows), so a bad value can never take down the round. `_parse_timestamp_field` wraps it and logs a warning when a value is present but unusable, with the same presence rule (`None` or `""` count as absent) for all three fields.
- `_sweep_pending_undelivered` in `task_submission_abci` now skips off-chain tasks. They carry the stamp too since this change, and the sweep must not turn them into `mech_onchain` request-only rows or drop them from the queue.

## Tests

- `test_parse_timestamp` (16 cases incl. the overflow strings).
- `test_build_predict_api_event_timestamp_precedence` (11 cases labelled by fallback tier, each asserting the exact warnings emitted: enqueue stamp, body value, `executed_at`, build time; overflow and out-of-range Unix values on both sides).
- `test_build_predict_api_event_execution_latency_ms_from_metadata` (10 cases incl. NaN and inf, each asserting whether a warning fires).
- `test_handlers.py`: `test_filter_requests_stamps_enqueued_at_local`, `test_filter_requests_stamps_enqueued_at_local_on_timed_out`, `test_contract_handler_handle_get_undelivered_reqs_stamps_timed_out` (on-chain and step-in stamps bracketed by `time.time()`), `test_filter_requests_keeps_existing_enqueued_at_local` and `test_contract_handler_handle_get_undelivered_reqs_keeps_existing_stamp` (a request re-fed on a later poll keeps its first stamp), `test_signed_requests_stamps_receipt_time_on_enqueued_task` (off-chain stamp is the handler's clock; a body value is dropped and logged).
- `test_onchain_write_path.py`: `test_sweep_skips_offchain_tasks`.
- All of the above fail on the base commit except the cases that guard behaviour which already existed: `test_filter_requests_stamps_enqueued_at_local` (the priority-branch stamp), the `executed_at_unix_out_of_range_uses_now` precedence case, the `missing_key` / `no_metadata` latency cases, and the two `keeps_existing` tests, which guard against a stamp being overwritten rather than missing.
- `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` green; `autonomy packages lock --check` green; packages pushed.
- Note: `test_verify_offchain_signature_wall_clock_deadline_bounds_eip1271_call` fails locally on `main` as well (takes 63 s, environment-dependent), unrelated to this change.


Adamantios (dismissed):

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1933,87 +1966,54 @@ def _build_predict_api_event(
...
-            try:
-                requested_at_iso = _iso_z(
-                    datetime.fromisoformat(requested_at_raw.replace("Z", "+00:00"))
-                )
-            except ValueError:
+        else:
+            executed_at_iso = _iso_z(executed_at_dt)
+        # ``requested_at`` is the time this mech received the request: the
+        # ``enqueued_at_local`` stamp the handler puts on the task on both
+        # the off-chain and the on-chain path. Tasks enqueued before that
```
OjusWiZard: Not true for step-in tasks. `filter_requests` only stamps `enqueued_at_local` in the priority-mech branch (`handlers.py:817`). Timed-out requests from other mechs go onto `unprocessed_timed_out_tasks` unstamped (`handlers.py:723` and `:827`) and are executed from `timed_out_tasks` (`behaviours.py:947`). On-chain bodies rarely carry `datetime`/`requested_at`, so every step-in delivery still ends up with `requested_at == executed_at`, which is the bug this PR fixes.

Could we stamp these where the handler queues them? Separately, a missing stamp logs nothing, because the warnings only fire on a present-but-unusable value. A warning when `requested_at` falls back to `executed_at` would make this visible.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -2791,6 +2792,12 @@ def _enqueue_offchain_request(
             RequestKey.IS_OFFCHAIN.value: True,
             RequestKey.REQUEST_DELIVERY_RATE.value: request_delivery_rate,
             **{k: v for k, v in data.items() if k not in reserved_keys},
+            # Receipt time, same key the on-chain path stamps in
+            # ``filter_requests``; it becomes the predict-api ``requested_at``.
+            # Must stay last in this literal: a later key wins on duplicates,
+            # so the stamp overrides any client-supplied value even if the
+            # reserved-keys filter above ever stops covering it.
+            "enqueued_at_local": time.time(),
```
OjusWiZard, line 2799: Stamping off-chain tasks changes `_sweep_pending_undelivered` (`task_submission_abci/behaviours.py:2561`). It sweeps any pending task with a numeric `enqueued_at_local` and has no `is_offchain` check. Its docstring says the off-chain path doesn't produce request-only events "by construction", but that only held because these tasks had no stamp.

The sweep is off by default. With it on, an off-chain task pending past `mech_events_sweep_max_age_seconds` is skipped with a warning every round, since signed bodies have no `priorityMech`/`requester`. Neither key is reserved, though: a client that sends both gets a `mech_onchain` request-only row with a `requester` it chose, and its task is dropped from `pending_tasks` without running.

Suggest skipping `is_offchain` tasks in the sweep, plus a sweep test with an off-chain task in the pending list.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -2791,6 +2792,12 @@ def _enqueue_offchain_request(
             RequestKey.IS_OFFCHAIN.value: True,
             RequestKey.REQUEST_DELIVERY_RATE.value: request_delivery_rate,
             **{k: v for k, v in data.items() if k not in reserved_keys},
+            # Receipt time, same key the on-chain path stamps in
```
OjusWiZard: `enqueued_at_local` is already in `reserved_keys`, so keeping it last is only a backup. I'd drop this comment or cut it to one line; the reasoning fits in the commit message.

packages/valory/skills/task_execution/tests/test_handlers.py
```diff
@@ -2245,6 +2245,51 @@ def test_signed_requests_accepts_zero_delivery_rate(
     assert len(handler_context.shared_state["pending_tasks"]) == 1
 
 
+def test_signed_requests_stamps_receipt_time_on_enqueued_task(
```
OjusWiZard, line 2360: This covers the off-chain stamp, but nothing asserts the on-chain one. `filter_requests` sets `enqueued_at_local` at `handlers.py:817`, and the existing `test_filter_requests_*` tests only check list lengths. If that `setdefault` were removed, on-chain `requested_at` would quietly fall back to `executed_at` again, with no warning and no failing test. Worth adding `test_filter_requests_stamps_enqueued_at_local` with the same before/after `time.time()` bracket, and one for the step-in path if that gets stamped.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1933,87 +1966,54 @@ def _build_predict_api_event(
...
-        # the FSM consensus channel and a buggy tool returning a wild
-        # value would otherwise crash the round; fall back to ``now_iso``.
-        executed_at_unix = response_data.get("executed_at")
-        if isinstance(executed_at_unix, (int, float)):
-            try:
-                executed_at_iso = _iso_z(
-                    datetime.fromtimestamp(executed_at_unix, tz=timezone.utc)
-                )
-            except (ValueError, OverflowError, OSError):
+        # ``executed_at`` is stamped as an ISO string in ``_handle_done_task``
```
OjusWiZard: Nit: this block and the one at 1986 restate the fallback chain the code already shows, and the PR description covers the why. I'd drop both.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -1933,87 +1966,54 @@ def _build_predict_api_event(
...
+        # when the tool result arrives; a Unix number is accepted as well.
+        # Anything unusable falls back to ``now_iso`` rather than crashing
+        # the round, since the value rides the FSM consensus channel.
+        executed_at_raw = response_data.get("executed_at")
```
OjusWiZard: Nit: parse, warn if unusable, fall back is written out three times (`executed_at`, `enqueued_at_local`, body `requested_at`). A small helper returning `Optional[datetime]` would shorten this and remove a mismatch: the body check skips `""` but the other two only skip `None`. The mismatch is harmless today.

OjusWiZard (commented):
The core fix looks right: both paths stamp `enqueued_at_local` as `time.time()`, a client can't override it, and all three timestamps go through `_iso_z`. Inline comments cover two gaps around where the stamp is and isn't set (step-in tasks, the pending sweep), a missing on-chain stamp test, and some comment/docstring trimming.

The test list in the PR description is out of date: names like `test_build_predict_api_event_preserves_iso_executed_at`, `..._malformed_iso_executed_at_falls_back` and `..._requested_at_falls_back_to_enqueue_time` no longer exist; they're now cases of `test_build_predict_api_event_timestamp_precedence`.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -253,6 +254,38 @@ def _discard_outstanding_nonce(
         target.pop(sender_key, None)
 
 
+def _parse_timestamp(value: Any) -> Optional[datetime]:
```
OjusWiZard, line 257: Nit: I'd cut this docstring to the summary plus `:param:`/`:return:`. The list of `None` cases restates the body. If the Python 3.10 trailing-`Z` workaround needs a note, one line at the `.replace` is enough.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -3729,6 +3729,26 @@ def _predict_api_event_setup(
     return done_task, executing_task
 
 
+def _capture_warnings(behaviour: Any, monkeypatch: Any) -> List[str]:
```
OjusWiZard, line 3732: Nit: the new test docstrings have explanatory paragraphs beyond the summary and `:param:` lines: here, `test_build_predict_api_event_timestamp_precedence`, `test_build_predict_api_event_execution_latency_ms_from_metadata`, and `test_signed_requests_stamps_receipt_time_on_enqueued_task`. One-line summaries would do.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -3884,6 +3904,252 @@ def test_build_predict_api_event_unix_executed_at_out_of_range_falls_back(
...
+
+
+def test_build_predict_api_event_warns_on_unusable_enqueue_stamp(
+    behaviour: Any,
+    params_stub: Any,
+    shared_state: Dict[str, Any],
+    monkeypatch: Any,
+) -> None:
+    """An unusable ``enqueued_at_local`` is logged, not silently skipped."""
+    # Symmetric with the body-timestamp warning: without this line a bad
```
OjusWiZard: Nit: this comment restates the docstring and the assertions below; I'd drop it.

packages/valory/skills/task_execution/tests/test_behaviours.py
```diff
@@ -3884,6 +3904,252 @@ def test_build_predict_api_event_unix_executed_at_out_of_range_falls_back(
...
+        "iso_z",
+        "iso_utc_offset",
+        "iso_non_utc_offset",
+        "iso_naive_as_utc",
+        "none",
+        "empty",
+        "bool",
+        "nan",
+        "inf",
+        "unix_out_of_range",
```
OjusWiZard, line 3880: Nit: now that parsing is centralised, `test_build_predict_api_event_unix_requested_at_out_of_range_falls_back` and `test_build_predict_api_event_unix_executed_at_out_of_range_falls_back` (unchanged, above) overlap with this case and the overflow cases in `test_build_predict_api_event_timestamp_precedence`. They could become precedence cases.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Good catch. Step-in requests are now stamped where the handler queues them: in the timed-out branch of filter_requests, and when timed-out requests are loaded from the contract in _handle_get_undelivered_reqs. There is also a warning now whenever requested_at falls back to executed_at, so a missing stamp shows up in the logs. Fixed in 7fc8fd04.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Agreed. The sweep now skips is_offchain tasks. test_sweep_skips_offchain_tasks puts a stale off-chain task with priorityMech and requester set next to a stale on-chain one, and checks that only the on-chain one is swept and both stay in the queue.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Cut to one line.

packages/valory/skills/task_execution/tests/test_handlers.py
DIvyaNautiyal07: Added test_filter_requests_stamps_enqueued_at_local, plus two for the step-in path: test_filter_requests_stamps_enqueued_at_local_on_timed_out and test_contract_handler_handle_get_undelivered_reqs_stamps_timed_out. All three use the before/after time.time() bracket.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Dropped both.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Added _parse_timestamp_field: it parses, warns when the value is present but unusable, and returns Optional[datetime]. All three fields go through it, so None and "" count as absent the same way everywhere.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Trimmed to the summary plus :param/:return, with a one-line note at the .replace.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Trimmed all four to a one-line summary plus the :param lines.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: That test is gone. It is now the bad_stamp_falls_back_to_body row of the precedence table, which also asserts the warning.

packages/valory/skills/task_execution/tests/test_behaviours.py
DIvyaNautiyal07: Folded both in as body_unix_out_of_range_uses_executed_at and executed_at_unix_out_of_range_uses_now. I also added a row where both the stamp and the body value are bad, so the order of the fallbacks and their warnings is pinned.

DIvyaNautiyal07:
Thanks for the review. All ten threads are addressed in 7fc8fd04, and the test list in the PR description is updated to match.

packages/valory/skills/task_execution/tests/test_handlers.py
DIvyaNautiyal07: Also added two tests in 2c38e71d that feed an already-stamped request back through filter_requests and _handle_get_undelivered_reqs and check the first stamp is kept. The handler sees the same timed-out request dicts on every poll, so overwriting the stamp instead of using setdefault would push requested_at to the last poll, and both tests fail if that happens.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -821,6 +823,7 @@ def filter_requests(self, reqs: List[Dict[str, Any]]) -> None:
                 self.pending_tasks.append(req)
 
             elif status == TIMED_OUT_STATUS:
+                req.setdefault("enqueued_at_local", time.time())
```
OjusWiZard, line 826: Question, not blocking: a request for another mech is first seen in the wait-for-timeout branch just below, which doesn't stamp. It only gets `enqueued_at_local` here, on the poll where it shows up as timed out. So a step-in row's `requested_at` lands about one response timeout after this mech first saw the request, and latency computed from the lake leaves out that wait.

If "when this mech could step in" is the intended meaning, this is fine. If `requested_at` should be first sight, a `req.setdefault("enqueued_at_local", time.time())` in the wait branch would do it, since the contract hands the same dicts back on each poll.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -720,9 +720,11 @@ def _handle_get_undelivered_reqs(self, body: Dict[str, Any]) -> None:
         # Reset lists.
         self.context.shared_state[INFLIGHT_READ_TS] = None
         self.wait_for_timeout_tasks.clear()
-        self.unprocessed_timed_out_tasks = body.get(
-            BodyKey.TIMED_OUT_REQUESTS.value, []
-        )
+        timed_out_reqs = body.get(BodyKey.TIMED_OUT_REQUESTS.value, [])
+        now = time.time()
+        for req in timed_out_reqs:
+            req.setdefault("enqueued_at_local", now)
```
OjusWiZard, line 726: Nit: every dict that reaches this loop came through the timed-out branch of `filter_requests` on an earlier poll (the contract returns the `timeout_tasks` it was given, as the same objects), so this `setdefault` never changes anything in production. Fine to keep as a backup. If you drop it, `test_contract_handler_handle_get_undelivered_reqs_stamps_timed_out` and `..._keeps_existing_stamp` go with it.

packages/valory/skills/task_execution/tests/test_handlers.py
```diff
@@ -1545,6 +1545,57 @@ def test_contract_handler_handle_get_undelivered_reqs_empty(
     assert len(ch.pending_tasks) == 0
 
 
+def test_contract_handler_handle_get_undelivered_reqs_stamps_timed_out(
```
OjusWiZard, line 1548: Nit: this test and `..._keeps_existing_stamp` differ only in whether the request already carries a stamp. Likewise, `test_filter_requests_stamps_enqueued_at_local` and `..._on_timed_out` differ only in priority mech and status. Both pairs could be parametrized the way `test_filter_requests_keeps_existing_enqueued_at_local` is.

OjusWiZard (approved):
The fixes check out. Step-in requests are stamped, the sweep skips off-chain tasks, and the fallback chain goes through one helper. I also traced the stamp through the contract round trip: `get_marketplace_undelivered_reqs` updates and returns the same dicts it's given, so `setdefault` does keep the first stamp across polls. Inline: one question that doesn't block this, and two optional nits.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Good point. This only affects on-chain step-in rows, since off-chain requests go straight to one mech and are never stepped in. For step-in, "when this mech could step in" is fine for us, so I'll leave it as is. It does mean step-in latency in the lake starts from when the request timed out, not from when it was sent.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Makes sense. It's harmless, so I'll keep it as a backup.

packages/valory/skills/task_execution/tests/test_handlers.py
DIvyaNautiyal07: Fair point, but I'll leave these as they are for this PR.

Adamantios (approved):

PR #490, opened 2026-09-14
fix: add olas-predict-r1-14b to TOKEN_PRICES (deliveries fail with 'Model not supported')

jmoreira-valory:
## Why

`TokenCounterCallback.__call__` rejects any model that is not a literal key in `TOKEN_PRICES`:

```python
if model not in list(TokenCounterCallback.TOKEN_PRICES.keys()):
    raise ValueError(f"Model {model} not supported.")
```

It raises **after** the LLM call has already completed, so a tool serving an unknown model burns the spend and then fails the delivery:

```
Failed to generate prediction after retries; last error: Model olas-predict-r1-14b not supported.
```

Not an edge case — every delivery of that tool.

`olas-predict-r1-14b` is the model behind the mech-predict tool `superforcaster_full_search_olas_predict_r1_14b` (valory-xyz/mech-predict#474). That tool cannot be promoted to production until this entry exists **and a `mech` release carries it**.

## The change

```python
"olas-predict-r1-14b": {"input": 0.0002, "output": 0.0002},
```

Everything else in the diff is the lock cascade: `benchmarks.py` lives in `task_execution`, so the hash change propagates through `task_execution` -> `task_submission_abci` -> `mech_abci` -> agent -> service -> `packages.json`. Same file set as #453 (`claude-fable-5`), plus one test.

## On the price

The model is **self-hosted**, so this is not a provider list price — there is no per-token metering behind it. The server costs a flat **~$0.90/hour**, which makes $/1k tokens a pure function of traffic. Measured against the running endpoint (1,159 input + 394 output = 1,553 tokens per forecasting request; 1,052 tok/s at 8 concurrent, still scaling):

| Requests/hour | Tokens/hour | $/1k tokens |
|---|---|---|
| 10 | 15.5k | $0.058 |
| 100 | 155k | $0.0058 |
| 1,000 | 1.55M | $0.00058 |
| Saturated (measured) | 3.79M | $0.00024 |

`$0.0002` is the well-utilised end of that range, and independently matches the hosted market rate for a 14B model. Two things worth knowing:

- **It understates cost at low traffic.** Break-even against GPT-4.1 is ~165 requests/hour — the same request costs $0.0055 on `gpt-4.1-2025-04-14`. Below that a dedicated server is *more* expensive per delivery than the API.
- **A flat rate slightly misattributes.** On a self-hosted server almost all GPU time is decode, not prefill, so the real cost sits in output tokens. Every other open-weights entry in this table is flat, so this follows suit.

## Verification

Exercised through the real executor path — `ComponentPackageLoader` -> `exec()` -> `run()` -> the real `TokenCounterCallback` — against local packages and the live endpoint. No FSM, no chain, no IPFS; that is the entire blast radius of a `TOKEN_PRICES` change.

```
component.yaml default_model = olas-predict-r1-14b
delivery    : {"p_yes": 0.2, "p_no": 0.8, "confidence": 0.6, "info_utility": 0.5}
actual_model: olas-predict-r1-14b
cost_dict   : input 1231, output 403, total_cost 0.000327
```

Negative control, same run with the key removed:

```
Attempt 1 failed with error: Model olas-predict-r1-14b not supported.
...
delivery: {"p_yes": null, "p_no": null, ..., "error": "Failed to generate
           prediction after retries: Model olas-predict-r1-14b not supported."}
```

So the failure this PR prevents is confirmed, and confirmed to be prevented. 969 tests pass; third-party hashes verified unchanged against `origin/main` key-for-key.

## Two notes for the reviewer

**The model was renamed mid-review.** The endpoint renamed this checkpoint from `qwen-14b-sft` to `olas-predict-r1-14b` on 2026-09-14; the old name now 404s. Same checkpoint (`root: checkpoints_sft_warm_start/best`) — a rename, not a new model. `TOKEN_PRICES` keys on the exact served string, so the entry follows. This is why the branch name still says `qwen-14b-sft`. `qwen-14b-base` and `qwen-14b-fine-tuned` were **not** renamed, so nothing else pointed at that server is affected.

**Watch out when verifying locally.** mech-predict *vendors* `task_execution`. Putting both repos on `sys.path` silently resolves `packages.valory.skills.task_execution` to mech-predict's released copy — the one without this change — so a test can reproduce the outage while appearing to test the fix. Assert `benchmarks.__file__` points where you expect. It also means step 2 of the propagation (re-pinning the vendored skills in mech-predict after the release) is load-bearing, not bookkeeping.

## Relationship to #472

#472 deprecates `TokenCounterCallback` but keeps `TOKEN_PRICES` functional and still enforcing, so this entry is still required and the two do not conflict. Once pricing moves into the tool packages, this entry goes away with the table.


jmoreira-valory:
## Update: there is a real cost basis, so zero is wrong

Since opening this I was given the serving cost: **~$0.90/hour** to keep the inference server up, with the caveat that the effective per-request price depends on traffic. That caveat turns out to be the whole story, so here is the arithmetic rather than a number picked by feel.

### Measured, against the running endpoint

A forecasting request through `superforcaster_full_search_olas_predict_r1_14b`:

| | |
|---|---|
| Prompt tokens | 1,159 |
| Completion tokens | 394 (reasoning model, `<think>` block included) |
| **Total** | **1,553 tokens/request** |
| Latency, single request | ~11 s |
| Throughput, 4 concurrent | 545 tok/s |
| Throughput, 8 concurrent | 1,052 tok/s (still scaling — not saturated) |

### The conversion

`$0.90/h` is a **fixed** cost, so `$/1k tokens = 900 ÷ tokens-per-hour`. It spans three orders of magnitude:

| Requests/hour | Tokens/hour | $/1k tokens | $/request |
|---|---|---|---|
| 1 | 1.6k | $0.58 | $0.90 |
| 10 | 15.5k | $0.058 | $0.090 |
| 100 | 155k | $0.0058 | $0.0090 |
| 1,000 | 1.55M | $0.00058 | $0.00090 |
| Saturated (measured, 8 concurrent) | 3.79M | $0.00024 | $0.00037 |

### Two things worth knowing before picking

**Break-even against GPT-4.1 is ~165 requests/hour.** The same request on `gpt-4.1-2025-04-14` costs $0.0055 (1.159 x $0.002 + 0.394 x $0.008). Below ~165 req/h the dedicated server is *more* expensive per delivery than the API; at 10 req/h it is roughly 16x more expensive. Whether this tool is cheap or expensive is a traffic question, not a model question.

**At a well-utilised ~1,000 req/h the derived cost is $0.00058/1k** — which is within rounding of the table's one existing open-weights entry, `nousresearch/nous-hermes-2-mixtral-8x7b-sft` at $0.00054 flat. Two independent routes landing on the same figure is the best argument available for it.

One caveat on the split: on a self-hosted server almost all GPU time is decode, not prefill, so the real cost sits in **output** tokens. A flat rate over-charges input and under-charges output. Every other open-model entry in this table is flat, so I have followed that, but a split is easy if preferred.

### What I need

Which basis should the entry carry? Anything on the table above is a one-line change and I will push it with the reasoning in the comment:

- `0.00054` flat — matches the existing open-weights convention and our ~1k req/h figure. My default if nobody objects.
- `0.0058` flat — priced at ~100 req/h, closer to realistic early-life traffic for a newly promoted tool.
- A split, e.g. `{"input": 0.0001, "output": 0.002}` — physically closer to where the GPU time goes.

Holding the branch at zero until this is settled; zero was a placeholder to keep the lookup from rejecting the model, and it should not merge as the final value.

jmoreira-valory:
## Verified through the real executor path, not just a unit test

Ran the change through `ComponentPackageLoader` -> `exec()` -> `run()` -> the real `TokenCounterCallback`, against the local packages and the live inference endpoint. No FSM, no chain, no IPFS — the entire blast radius of a `TOKEN_PRICES` change.

**With the entry present:**

```
component.yaml default_model = qwen-14b-sft
delivery    : {"p_yes": 0.2, "p_no": 0.8, "confidence": 0.6, "info_utility": 0.5}
actual_model: qwen-14b-sft
cost_dict   : {'input_tokens': 1231, 'output_tokens': 403, 'total_tokens': 1634, ...}
```

**Negative control — same run with the key removed:**

```
Attempt 1 failed with error: Model qwen-14b-sft not supported.
Attempt 2 failed with error: Model qwen-14b-sft not supported.
Attempt 3 failed with error: Model qwen-14b-sft not supported.
delivery: {"p_yes": null, "p_no": null, ..., "error": "Failed to generate prediction
           after retries: Model qwen-14b-sft not supported."}
```

So the failure this PR prevents is confirmed, and it is confirmed to be prevented.

Two things worth recording from building that harness.

**The vendored copy is a real trap.** My first attempt put both this repo and the mech-predict worktree on `sys.path`. mech-predict *vendors* `task_execution`, so `packages.valory.skills.task_execution` silently resolved to its released copy — the one without this change — and the test reproduced the outage while appearing to test the fix. Anyone verifying a `task_execution` change from a mech-predict checkout should assert `benchmarks.__file__` points where they think it does. It also means step 2 of the propagation (re-pinning the 3 vendored skills in mech-predict) is not optional bookkeeping: until it happens, mech-predict genuinely runs the old table.

**`total_cost` is 0.00 while the price is zero.** The canonical pass criterion for this kind of change is `cost_dict['total_cost'] > 0`, which cannot hold at zero pricing. That is one more reason not to merge the placeholder — the tokens are counted correctly (1,231 in / 403 out), but the cost line stays flat. Pending the basis decision in the comment above.

packages/valory/skills/task_execution/utils/benchmarks.py
```diff
@@ -56,6 +56,13 @@ class TokenCounterCallback:
         "google/gemini-2.5-flash:online": {"input": 0.0003, "output": 0.0025},
         "anthropic/claude-haiku-4.5:online": {"input": 0.001, "output": 0.005},
         "anthropic/claude-sonnet-4:online": {"input": 0.003, "output": 0.015},
+        # Self-hosted on Valory infrastructure, so there is no provider list
+        # price to copy and no per-token metering behind it. Priced at zero for
+        # now: the entry exists because __call__ raises for any model that is
+        # not a literal key, which would fail every delivery AFTER the LLM call
+        # has already been paid for. TODO: revisit once the serving cost is
+        # known and can be expressed per 1k tokens.
+        "qwen-14b-sft": {"input": 0.0, "output": 0.0},
```
Adamantios:
```suggestion
        "qwen-14b-sft": {"input": 0.0002, "output": 0.0002},
```

jmoreira-valory: Applied, with one deviation I could not avoid: `291969c5`.

```python
"olas-predict-r1-14b": {"input": 0.0002, "output": 0.0002},
```

The price is exactly as you suggested. The **key** had to change — the endpoint renamed this checkpoint between your review and now, and the old name is gone:

```
NotFoundError: Error code: 404 - The model `qwen-14b-sft` does not exist.
```

`/v1/models` now serves `olas-predict-r1-14b` (and `olas-predict-r1-14b-calibrated`) off the same `root: checkpoints_sft_warm_start/best`, so it is a rename, not a new checkpoint. `qwen-14b-base` and `qwen-14b-fine-tuned` were not renamed, so nothing else pointed at this server is affected. Since `TOKEN_PRICES` keys on the exact served string, applying the suggestion verbatim would have priced a model that no longer exists.

On the number itself: it sits between the two bounds in the arithmetic above — the server costs ~$0.90/hour flat, which works out at $0.00024/1k at the throughput I measured under load and $0.00058/1k at ~1k requests/hour. $0.0002 is the well-utilised end of that, and it matches the hosted market rate for a 14B model, so it is defensible from both directions. Worth keeping in mind that it understates the cost at low traffic: below roughly 165 requests/hour a dedicated server is more expensive per delivery than calling GPT-4.1.

The placeholder zero is gone and the test now asserts the accrued cost instead of its absence:

```python
assert cb.cost_dict["total_cost"] == pytest.approx(0.0002 * 1.5)
```

That also fixes something I flagged earlier: the canonical pass criterion for this class of change is `cost_dict['total_cost'] > 0`, which was unmeetable while the price was zero. Re-ran the executor path against the live endpoint:

```
component.yaml default_model = olas-predict-r1-14b
delivery    : {"p_yes": 0.2, "p_no": 0.8, "confidence": 0.6, "info_utility": 0.5}
actual_model: olas-predict-r1-14b
cost_dict   : input 1231, output 403, total_cost 0.000327
```

Negative control still surfaces `Model olas-predict-r1-14b not supported.` 969 tests pass, third-party hashes unchanged.

OjusWiZard (approved):
Approving — the change is correct and nothing here blocks merge.

I traced the entry through to where it actually lands: `TOKEN_PRICES` → `token_to_cost` → `calculate_cost` → `cost_dict["total_cost"]` → `get_cost_for_done_task`, which does `math.ceil(total_cost * 100)`. At realistic Predict-tool token counts that settles at 1–3 credits — neither rounding to zero nor overshooting, which is the thing a stand-in price could plausibly have got wrong. The dict shape matches the other 25 entries, so `token_to_cost` indexes it identically.

The test earns its place. `__call__` raises *before* the `try` block, so the `ValueError` is uncaught — remove the entry and `test_self_hosted_model_is_supported_and_priced` fails for the real reason. I'd also keep the hard-coded `0.0002 * 1.5` rather than deriving it from `TOKEN_PRICES[...]` the way the sibling tests do: those are tautological and would pass at any price, whereas this one pins the price as a fact. Deliberate, not sloppy.

On the hash cascade: the five new first-party hashes propagate to every file that references them, with no orphans and no third-party entries touched. `lock_check` passing on all three platforms is the authoritative check, but the cross-references are self-consistent too.

One non-blocking note, and it is about the design rather than this fix: **the real fragility is the coupling.** `TOKEN_PRICES` keys on the exact served string, and that string already moved once mid-review. `default_model` lives in an external tools package's `component.yaml`, so nothing in this repo or in CI can catch the next rename — it surfaces in production, after the spend, on every delivery. The live-endpoint verification in 291969c5 is good evidence for today, but it is a point-in-time check against a server someone else can rename tomorrow. Not something to solve here; worth naming because #472 moving pricing into the tool packages is what actually closes it, and until then this class of PR recurs.

Adamantios (approved):

jmoreira-valory:
Thanks — and thanks for tracing it to the settlement rather than stopping at the dict.

I re-ran your path to pin the number. `get_cost_for_done_task` does `math.ceil(total_cost * 100)` as you say, but it lands on **exactly 1 credit at every reachable token count**, not 1–3:

| | in | out | total_cost | credits |
|---|---|---|---|---|
| measured delivery | 1,231 | 403 | $0.000327 | 1 |
| typical | 1,159 | 394 | $0.000311 | 1 |
| full evidence block | 3,373 | 469 | $0.000768 | 1 |
| arithmetic worst case | 4,096 | 2,048 | $0.001229 | 1 |

A second credit needs `total_cost > $0.01`, i.e. ~50,000 tokens at this price. The model's context window is 8,192 shared between prompt and completion, so that is unreachable by construction — the last row above is already past what the tool can actually emit.

Your conclusion holds, and the precise version is slightly stronger: **for this tool, any price in (0, $0.01 per delivery] bills identically at 1 credit.** So the choice of $0.0002 is not billing-sensitive; it matters for cost *accounting*, not for what the requester pays. It also retroactively confirms the placeholder zero was wrong in a way I had not spotted — `math.ceil(0.0 * 100)` is 0, so a zero price would have billed **0 credits**, not 1.

Good catch on the test rationale, and you read the intent correctly — the sibling tests deriving the expected value from `TOKEN_PRICES[...]` pass at any price, which makes them assertions about arithmetic rather than about the entry.

## On the coupling

This is the right thing to be uneasy about, and I agree it is the real fragility. The string moved once during this PR precisely because nothing anywhere is watching it.

One correction to "nothing in this repo or in CI can catch the next rename": nothing can catch a rename on the **server**, true — that is an external system and only a live call would see it. But the *other* half of the coupling is checkable today. mech-predict **vendors** `task_execution`, so a test there can import `TokenCounterCallback.TOKEN_PRICES` and assert every tool's `component.yaml` `default_model` is a key in it:

```python
component = yaml.safe_load((TOOL_DIR / "component.yaml").read_text())
assert component["params"]["default_model"] in TokenCounterCallback.TOKEN_PRICES
```

That would have caught this PR's rename at CI time in the repo that owns the tool, and it catches the general case of a new tool shipping an unpriced `default_model` — which is the same failure that produced #453. It does not catch a server-side rename, but it converts "surfaces in production after the spend" into "fails a unit test" for every cause except that one.

Worth a follow-up on mech-predict rather than here — happy to open it. And agreed that #472's direction is what actually closes the class; until pricing lives with the tool, this shape of PR recurs.

PR #491, opened 2026-09-14
fix: price x-ai/grok-4.3:online, used by resolve_market_jury

jmoreira-valory:
## The bug

`resolve_market_jury` fans out to four OpenRouter voters. Three are keys in `TOKEN_PRICES`; `VOTER_MODEL_GROK = "x-ai/grok-4.3:online"` is not. `TokenCounterCallback.__call__` raises for an unknown model, so that voter's usage never reaches `cost_dict` and **the mech under-bills for every jury delivery**.

Confirmed live, not inferred:

- `VOTER_MODEL_GROK = "x-ai/grok-4.3:online"` on `origin/main`.
- The CID deployed as `resolve-market-jury-v1` (`bafybeigdv6zbb…`) matches `packages.json` on main **exactly**, so the running code is the code that references it.

## Severity — it does NOT fail deliveries

Worth stating plainly, because this looks like the usual "unpriced model kills every delivery" shape and it is not. The jury wraps its callback:

```python
except Exception as e:  # pylint: disable=broad-except
    print(f"  Warning: counter_callback failed for {model}: {e}")
```

The `ValueError` is caught, the completion returns, the delivery settles — with one voter's cost silently missing. The symptom is under-accounting in the logs, not a failed request, which is arguably harder to notice since nothing surfaces on-chain. Nothing needs rolling back.

The only other `counter_callback` call site is the `delivery_rate == 0` branch, which passes `max_cost=` / `models_calls=` — neither implemented by `TokenCounterCallback`. That is a separate pre-existing gap, untouched here.

## The price — verified against OpenRouter, and corrected

```python
"x-ai/grok-4.3:online": {"input": 0.00125, "output": 0.0025},
```

`x-ai/grok-4.3` lists at **$1.25 / $2.50 per MTok**, checked against `https://openrouter.ai/api/v1/models`.

I had this wrong in the first commit and it is worth recording why, because the trap is reusable:

| | |
|---|---|
| First attempt | `{"input": 0.0002, "output": 0.0005}` — **6.25x and 5x too low** |
| Anchored on | `x-ai/grok-4.1-fast:online`, whose **base model is no longer in OpenRouter's catalogue** — a stale entry, the worst thing in the table to copy |
| Justified by | "under-pricing self-corrects via `call_cost`" — true mechanically, but not this table's convention |

The other three `:online` entries each match OpenRouter's listed rate **exactly**:

| Entry | Table | OpenRouter | |
|---|---|---|---|
| `openai/gpt-4.1:online` | 0.002 / 0.008 | 0.002 / 0.008 | match |
| `google/gemini-2.5-flash:online` | 0.0003 / 0.0025 | 0.0003 / 0.0025 | match |
| `anthropic/claude-haiku-4.5:online` | 0.001 / 0.005 | 0.001 / 0.005 | match |
| `x-ai/grok-4.1-fast:online` | 0.0002 / 0.0005 | — **base model not in catalogue** | stale |

So exact is the convention, and this entry now follows it. `call_cost` still carries the `:online` web-search fee into `extra_cost` — that is a per-search fee, not a token cost, and not a licence to under-price the tokens.

## On the `:online` ids

All four are valid. `:online` is an OpenRouter **request-time variant**, not a catalogue id, so it does not appear in `/api/v1/models` — but every base model resolves (`openai/gpt-4.1`, `x-ai/grok-4.3`, `google/gemini-2.5-flash`, `anthropic/claude-haiku-4.5`). Flagging it because "not found in the model list" is an easy false alarm here.

## Test

Asserts the four current jury voter models are keys in the table — a snapshot of the jury set, nothing more.

**It does not lock the bug.** Raised in review and correct: the check never calls `__call__`, and it hardcodes the four strings, so renaming `VOTER_MODEL_GROK` in mech-predict leaves it green. Verified against `origin/main`: the membership check passes there, while a `__call__`-based version raises `ValueError: Model x-ai/grok-4.3:online not supported.` — i.e. only the latter would have caught this.

The stronger assertion follows in a separate PR (test-only; this one is green with two approvals and the mech release carrying the price is blocking mech-predict work).

**Neither version can catch a voter rename.** `VOTER_MODEL_GROK` lives in mech-predict and nothing in this repo can see it. That needs a guard on the mech-predict side, walking each tool's declared models against the vendored `TOKEN_PRICES` — queued there, and the check that would have caught the May bump.

## How it was found, and what was left alone

Sweeping every `*MODEL*` string constant across all 29 mech-predict customs tools against `TOKEN_PRICES`. This was the **only** hit. Two things surfaced and were deliberately not touched:

- `x-ai/grok-4.1-fast:online` is stale — its base model is gone from OpenRouter. Nothing in-tree uses it, and removing a key could break a requester passing it explicitly, so it stays. Worth a follow-up decision.
- `TOKEN_PRICES` has `claude-4-sonnet-20250514`, while Anthropic's real id is `claude-sonnet-4-20250514` (words transposed). Unused today, so latent — but the next tool declaring the correct name hits #453 again. `resolve_market_jury`'s own `component.yaml` declares that correct name as `default_model`, which is unpriced but harmless since the tool ignores `default_model`.

969 tests pass; third-party hashes verified unchanged against `origin/main`.

## Relationship to #490

Independent. Both add a key to the same dict, so whichever merges second needs a trivial rebase.


jmoreira-valory:
## Verification of the model and the rate, against OpenRouter's API

Re-checked from `https://openrouter.ai/api/v1/models` rather than from the table or from memory. Full record for `x-ai/grok-4.3`:

```json
{
  "name": "SpaceXAI: Grok 4.3",
  "context_length": 1000000,
  "pricing": {
    "prompt": "0.00000125",
    "completion": "0.0000025",
    "web_search": "0.005",
    "input_cache_read": "0.0000002",
    "overrides": [
      { "min_prompt_tokens": 200000,
        "prompt": "0.0000025", "completion": "0.000005",
        "input_cache_read": "0.0000004" }
    ]
  }
}
```

| Claim | Verdict |
|---|---|
| The model exists on OpenRouter | **yes** — `x-ai/grok-4.3`, 1M context |
| The entry's rate is correct | **yes** — `0.00000125 × 1000 = $0.00125/1k`, `0.0000025 × 1000 = $0.0025/1k`, exactly what this PR sets |
| The tool references it | **yes** — `VOTER_MODEL_GROK` at `resolve_market_jury.py:61` on `origin/main`, used at `:304` |
| It is the deployed code | **yes** — the CID served as `resolve-market-jury-v1` matches `packages.json` on main exactly |
| It was genuinely unpriced | **yes** — the table had `x-ai/grok-4.1-fast:online` and no 4.3 |

Three details the record should carry:

**`:online` is not a catalogue id.** Zero of the 445 models OpenRouter lists carry a `:online` suffix — it is a request-time variant that enables the web plugin. So "not found in `/api/v1/models`" is an easy false alarm for all four of the jury's voter ids; every base model resolves.

**`web_search: 0.005`** confirms the reasoning in the diff comment: the `:online` fee is **per search**, not per token. It cannot be expressed in this table and correctly arrives through `call_cost` into `extra_cost`.

**There is a long-context override** — above 200,000 prompt tokens the rate doubles to $2.50/$5.00 per MTok. It does not apply here (jury prompts are a few thousand tokens) and `TOKEN_PRICES` has no way to express a tiered rate. Worth naming so nobody re-derives it: if a tool ever sends this model a 200k+ prompt, the token cost is under-counted by 2x, and only `call_cost` would correct it.

## Rebased onto #490

`13541a8f` merged while this was open. Both PRs add a key to the same dict, so `benchmarks.py` auto-merged with both entries. Two real conflicts, resolved in `03ede7e7`: both branches had inserted a test at the same anchor (**kept both** — they assert different things), and the generated CID files were taken from main and regenerated. Third-party hashes verified identical to `origin/main` key-for-key. 970 tests pass.

dagacha (approved):

packages/valory/skills/task_execution/tests/utils/test_benchmarks.py
```diff
@@ -126,6 +126,22 @@ def test_self_hosted_model_is_supported_and_priced(self) -> None:
...
+        instead, and the mech under-bills for that request.
+        """
+        jury_models = (
+            "openai/gpt-4.1:online",
+            "x-ai/grok-4.3:online",
+            "google/gemini-2.5-flash:online",
+            "anthropic/claude-haiku-4.5:online",
+        )
+        missing = [m for m in jury_models if m not in TokenCounterCallback.TOKEN_PRICES]
+        assert not missing, f"unpriced jury voter models: {missing}"
```
dagacha, line 143: This only checks four hardcoded strings are keys. It never calls `__call__`, and it cannot see `VOTER_MODEL_GROK` in mech-predict.

The PR text says the next voter swap fails CI. It does not: rename the jury model and this stays green. Neighbor `test_self_hosted_model_is_supported_and_priced` is the pattern that actually would have failed on main:

```python
cb = TokenCounterCallback()
cb("x-ai/grok-4.3:online", _dummy_counter, input_tokens=1000, output_tokens=500)
assert cb.cost_dict["total_cost"] == pytest.approx(0.00125 + 0.0025 * 0.5)
```

Keep the four-key membership check if you want a snapshot of the current jury set; it is not a substitute for the call.

dagacha (changes requested):
Reviewed with **Omp** (LLM: `xai-oauth/grok-4.6`, effort: Medium).

Price for `x-ai/grok-4.3:online` is correct against OpenRouter (`prompt: 0.00000125`, `completion: 0.0000025` per token → `$0.00125` / `$0.0025` per 1k, matching `PRICE_NUM_TOKENS = 1000`). Hash cascade and CI are fine.

One change required: the new test does not exercise `__call__`, so it does not lock the under-billing bug this PR claims to prevent. See inline.

dagacha (approved):

DIvyaNautiyal07 (approved):

packages/valory/skills/task_execution/utils/benchmarks.py
```diff
@@ -53,6 +53,11 @@ class TokenCounterCallback:
         },
         "openai/gpt-4.1:online": {"input": 0.002, "output": 0.008},
         "x-ai/grok-4.1-fast:online": {"input": 0.0002, "output": 0.0005},
+        # Used by resolve_market_jury's VOTER_MODEL_GROK. OpenRouter's listed
+        # rate for x-ai/grok-4.3 ($1.25/$2.50 per MTok), matching how the other
+        # `:online` entries here are priced. The `:online` web-search fee is not
+        # a token cost; it arrives via `call_cost` and lands in `extra_cost`.
```
OjusWiZard, line 59: Per the comment convention: default to no comments — rationale belongs in the commit message and PR body, which already cover every clause here (the price source, the `:online` pricing convention, and the `call_cost` -> `extra_cost` split). `git blame` -> `cbe5590c` -> this PR reconstructs all of it, so this block only duplicates it and will rot the next time the price moves.

Worth distinguishing from the `olas-predict-r1-14b` comment a few lines below: that one documents a number that is *not* a provider list price — a derivation that lives nowhere else, which earns its line. This one documents a published list price, which doesn't.

Suggest dropping the four comment lines and letting the entry stand bare, like the other `:online` entries.

packages/valory/skills/task_execution/tests/utils/test_benchmarks.py
```diff
@@ -126,6 +126,22 @@ def test_self_hosted_model_is_supported_and_priced(self) -> None:
         assert cb.cost_dict["total_tokens"] == 1500
         assert cb.cost_dict["total_cost"] == pytest.approx(0.0002 * 1.5)
 
+    def test_every_jury_voter_model_is_priced(self) -> None:
+        """resolve_market_jury's voter models must all be keys in the table.
+
+        The tool swallows the callback's ValueError, so an unpriced voter does
+        not fail the delivery -- its cost is silently dropped from cost_dict
+        instead, and the mech under-bills for that request.
+        """
```
OjusWiZard, line 135: Docstring convention here is a one-line summary, not a paragraph of rationale. The "why" (swallowed `ValueError` -> silent under-billing) is already stated at length in the PR description.

```suggestion
        """resolve_market_jury's voter models must all be keys in the table."""
```

packages/valory/skills/task_execution/tests/utils/test_benchmarks.py
```diff
@@ -126,6 +126,22 @@ def test_self_hosted_model_is_supported_and_priced(self) -> None:
...
+        jury_models = (
+            "openai/gpt-4.1:online",
+            "x-ai/grok-4.3:online",
+            "google/gemini-2.5-flash:online",
+            "anthropic/claude-haiku-4.5:online",
+        )
```
OjusWiZard, line 141: These four strings are a hand-copied snapshot of `VOTER_MODEL_*` from `resolve_market_jury`, which lives in mech-predict. I grepped this repo: `resolve_market_jury` and `VOTER_MODEL` appear nowhere except this test and the new comment in `benchmarks.py`.

So the PR description's claim — "the next voter swap fails CI instead of silently dropping a cost" — doesn't hold. If mech-predict swaps a voter, this tuple is untouched and the test stays green. What it does catch is someone deleting a key from `TOKEN_PRICES` *in this repo*, which is still worth having; it just isn't the cross-repo guard the description promises.

Hardcoding is unavoidable on this side, so the fix is to the claim rather than the code: either soften the description, or put the real guard in mech-predict where the constants actually live. If you want the coupling recorded, a tracking issue beats a version pointer in a comment — the pointer goes stale the moment mech-predict moves.

packages/valory/skills/task_execution/tests/utils/test_benchmarks.py
OjusWiZard, line 143: `in TOKEN_PRICES` is a key-presence check, so nothing in the suite asserts `0.00125` / `0.0025`. I confirmed by mutation: reverting the entry to the original `0.0002` / `0.0005` leaves all 13 tests in this file green. It would equally pass for an entry missing `"output"`.

Given this PR's own history — `8a0a2a0c` shipped a 6.25x/5x underpricing that `cbe5590c` then corrected — a wrong price is precisely the bug class left uncovered. The sibling `test_self_hosted_model_is_supported_and_priced` just above already has the stronger shape:

```python
cb("x-ai/grok-4.3:online", _dummy_counter, input_tokens=1000, output_tokens=500)
assert cb.cost_dict["total_cost"] == pytest.approx(0.00125 + 0.0025 * 0.5)
```

Worth adding alongside the membership check rather than replacing it.

OjusWiZard (approved):
Approving — the fix is correct and independently verified.

**What I checked, rather than taking from the description:**

- **The price is right.** Queried OpenRouter's live `/api/v1/models`: `x-ai/grok-4.3` is `0.00000125` / `0.0000025` per token, i.e. **0.00125 / 0.0025 per 1K** — exactly what this adds. The other three `:online` entries also match their base models (`gpt-4.1` 0.002/0.008, `gemini-2.5-flash` 0.0003/0.0025, `claude-haiku-4.5` 0.001/0.005), so "exact is the convention" holds.
- **The stale-entry note holds.** `x-ai/grok-4.1-fast` is genuinely absent from the catalogue (only 4.6, 4.5, 4.3, 4.3:batch and 4.20 are listed). Agreed it should stay for now.
- **Hashes are consistent.** `autonomy packages lock --check` passes in a clean worktree at HEAD. (Heads-up for anyone reproducing locally: running it in a dirty working directory reports spurious mismatches for unrelated packages, from stray `__pycache__` dirs left by test runs.)
- 13/13 tests pass locally; all 22 CI checks green.

**Non-blocking, four inline notes:** two are comment/docstring convention, one is that the test doesn't deliver the cross-repo protection the description claims, and one is that the price values themselves are asserted nowhere.

One thing worth a follow-up issue rather than this PR: `cost_dict` flows unvalidated into the delivery payload (`behaviours.py:1209`, `:2172`), so a voter whose cost was dropped is indistinguishable from a cheap call — no metric, no reconciliation, just a `logging.error` lost in the noise. That's the reason this class of bug is invisible in the first place.

packages/valory/skills/task_execution/tests/utils/test_benchmarks.py
jmoreira-valory: You are right, and I verified it rather than taking it on faith. Loading `origin/main`'s `benchmarks.py` and running your version:

```
ValueError: Model x-ai/grok-4.3:online not supported.
```

Mine passes against main. So the membership check does not lock the bug this PR exists to fix — your `__call__` version does.

**The PR text was the worse half of this.** It claimed "the next voter swap fails CI instead of silently dropping a cost". That is false for the reason you give: the test hardcodes the four strings, so renaming `VOTER_MODEL_GROK` leaves it green. I have corrected the description — editing it touches no commits, so it does not disturb the approvals.

**One thing worth being explicit about, because it limits both versions:** neither test can catch a voter *rename*. `VOTER_MODEL_GROK` lives in mech-predict; nothing in this repo can see it. Your version is strictly better — it locks the behaviour and would have failed on main — but the cross-repo drift that caused this needs a guard on the mech-predict side, walking each tool's declared models against the vendored `TOKEN_PRICES`. That is queued there; it is the check that would have caught the May bump.

**On sequencing:** taking the test change as a follow-up rather than pushing here, since this PR is test-only-green with two approvals and mech-predict work is blocked behind the release that carries it. The price — the part that actually fixes the under-billing — is unchanged and independently verified. Follow-up will carry your exact assertion.

Thanks for catching it; the membership check reading as coverage it does not provide is precisely the failure mode this PR is about.

jmoreira-valory: Agreed, and verified: your version raises `ValueError: Model x-ai/grok-4.3:online not supported.` against `origin/main` while the membership check passes there. The follow-up adds your exact assertion alongside the membership check (OjusWiZard independently asked for the same thing, and for keeping both).

Deferred to an immediate follow-up PR rather than pushed here: this PR is the price fix that a `mech` release is blocking mech-predict work on, and everything raised is test/comment-level with no effect on the shipped value. Resolving so the merge can proceed — not because it is dismissed. The follow-up is being opened now and I will link it here.

packages/valory/skills/task_execution/utils/benchmarks.py
jmoreira-valory: Taking this, and the distinction you draw is the right one: a published list price needs no in-line rationale because `git blame` reaches the PR, whereas the `olas-predict-r1-14b` comment documents a *derived* number that exists nowhere else. The four lines come out in the follow-up.

Noting for the record that this is the one thread of the five touching a non-test file — removing the comment changes the file content and so triggers the full CID cascade (task_execution -> task_submission_abci -> mech_abci -> agent -> service -> packages.json). That is the specific reason it is batched into the follow-up rather than pushed here: the cascade then happens once, in a PR already relocking, instead of twice.

Deferred to an immediate follow-up PR rather than pushed here: this PR is the price fix that a `mech` release is blocking mech-predict work on, and everything raised is test/comment-level with no effect on the shipped value. Resolving so the merge can proceed — not because it is dismissed. The follow-up is being opened now and I will link it here.

packages/valory/skills/task_execution/tests/utils/test_benchmarks.py
jmoreira-valory: Taking the suggestion verbatim in the follow-up.

Deferred to an immediate follow-up PR rather than pushed here: this PR is the price fix that a `mech` release is blocking mech-predict work on, and everything raised is test/comment-level with no effect on the shipped value. Resolving so the merge can proceed — not because it is dismissed. The follow-up is being opened now and I will link it here.

packages/valory/skills/task_execution/tests/utils/test_benchmarks.py
jmoreira-valory: You and dagacha reached this independently and you are both right. I have already corrected the PR description — editing it touches no commits, so it disturbed nothing — and it now states plainly that the tuple is a snapshot of the current jury set, that renaming `VOTER_MODEL_GROK` leaves the test green, and that no cross-repo guard is possible from this repo.

Agreed that the real guard belongs in mech-predict, walking each tool's declared models against the vendored `TOKEN_PRICES`. That is queued there and is the check that would have caught the May bump — the tool moved `grok-4.1-fast (retired) -> grok-4.3` in mech-predict #282 on 2026-05-18 and this table was never updated to match. Agreed too that a tracking issue beats a version pointer in a comment.

Deferred to an immediate follow-up PR rather than pushed here: this PR is the price fix that a `mech` release is blocking mech-predict work on, and everything raised is test/comment-level with no effect on the shipped value. Resolving so the merge can proceed — not because it is dismissed. The follow-up is being opened now and I will link it here.

packages/valory/skills/task_execution/tests/utils/test_benchmarks.py
jmoreira-valory: This is the sharpest of the five and the mutation result settles it — reverting to `0.0002`/`0.0005` leaving all 13 tests green means the one number this PR exists to get right is unasserted. Given the PR's own history (`8a0a2a0c` shipped exactly that 6.25x/5x underpricing, `cbe5590c` corrected it), that is the bug class this suite most needs to cover.

Taking your shape, and keeping the membership check alongside as you suggest rather than replacing it — they cover different things: yours pins the value, the membership check pins the jury set.

Deferred to an immediate follow-up PR rather than pushed here: this PR is the price fix that a `mech` release is blocking mech-predict work on, and everything raised is test/comment-level with no effect on the shipped value. Resolving so the merge can proceed — not because it is dismissed. The follow-up is being opened now and I will link it here.

jmoreira-valory:
Follow-up opened: https://github.com/valory-xyz/mech/pull/492

It carries all five threads from this PR — the `__call__` + `total_cost` assertion (@dagacha, @OjusWiZard), the comment removal and the one-line docstring (@OjusWiZard). Mutation-verified: reverting the entry to the original wrong `0.0002`/`0.0005` now fails the suite, which it did not before.

The cross-repo rename guard stays out of `mech` as agreed — it is queued on the mech-predict side where the constants actually live.

PR #492, opened 2026-09-14
test: assert the grok price is accrued, not merely present

jmoreira-valory:
Follow-up to #491, addressing all five review threads on it. Deferred from that PR because a `mech` release carrying its price fix was blocking mech-predict work, and everything raised was test- or comment-level with no effect on the shipped value.

## The problem: the price was unasserted

@OjusWiZard demonstrated by mutation that reverting the entry to the original wrong `0.0002`/`0.0005` left **all 13 tests in the file green**. `in TOKEN_PRICES` is a key-presence check — it would equally pass for an entry missing `"output"`.

Given #491's own history — `8a0a2a0c` shipped exactly that 6.25×/5× underpricing and `cbe5590c` corrected it — a wrong price is the bug class this suite most needed to cover, and was the one it didn't.

## The fix, and proof it works

`test_grok_voter_model_accrues_its_listed_price` — the shape @dagacha and @OjusWiZard converged on independently: call `__call__`, assert `total_cost`.

```python
cb("x-ai/grok-4.3:online", _dummy_counter, input_tokens=1000, output_tokens=500)
assert cb.cost_dict["total_cost"] == pytest.approx(0.00125 + 0.0025 * 0.5)
```

Verified both directions rather than assumed:

| | result |
|---|---|
| correct price | **971 passed** |
| mutated back to `0.0002`/`0.0005` | **1 failed** — `test_grok_voter_model_accrues_its_listed_price`, on the `total_cost` assertion |

The membership check **stays alongside** rather than being replaced, per @OjusWiZard — they cover different things. The new test pins the **value**; the membership check pins the jury **set**.

## Also in this round

- **Dropped the four comment lines** above the grok entry (@OjusWiZard). The distinction drawn there is right: a published list price needs no in-line rationale because `git blame` reaches the PR, whereas the `olas-predict-r1-14b` comment documents a *derived* number that exists nowhere else. This is the one change touching a **non-test** file, which is why it was batched here — it alters file content and so triggers the full CID cascade, which now happens once in a PR already relocking instead of twice.
- **Docstring back to a one-line summary** (@OjusWiZard), suggestion taken verbatim.

## Not addressed here, by agreement

Neither test can catch a voter **rename**. `VOTER_MODEL_GROK` lives in mech-predict and is invisible from this repo — grepping confirms `VOTER_MODEL` appears nowhere in `mech` outside this test. Both reviewers raised this and both agreed it cannot be solved from here.

That guard belongs in mech-predict, walking each tool's declared models against the **vendored** `TOKEN_PRICES` (the vendored pin is what the deployed agent runs, so checking `mech`'s main would give false greens). It is queued there, and it is the check that would have caught the drift behind all of this: mech-predict #282 moved the jury from `grok-4.1-fast` (retired) → `grok-4.3` on 2026-05-18, and this table was never updated to match. Four months, no signal.

971 tests pass; third-party hashes verified unchanged against `origin/main`.


dagacha (approved):
Reviewed with **Omp** (LLM: `xai-oauth/grok-4.6`, effort: Medium).

This is the regression #491 was missing. `__call__` + hardcoded `total_cost` (`0.00125 + 0.0025 * 0.5`) is the right pin: it fails if the key is gone, if `output` is missing, or if the rate is mutated back to `0.0002`/`0.0005`. Membership check staying alongside is correct — it covers the jury *set*, this covers grok's *value*. Comment drop on a published list price is right.

Math matches `token_to_cost` (`PRICE_NUM_TOKENS = 1000`). Hash cascade is the expected CID lock after touching `benchmarks.py`. CI green, including `lock_check`.

No further changes.

OjusWiZard (approved):

dagacha (approved):

PR #493, opened 2026-09-15
Settle only delivered tasks; add off-chain Prometheus metrics

DIvyaNautiyal07:
## Why

Off-chain requests are served to the user over HTTP and then paid for later, when the task is batched into the next Safe transaction as `deliverMarketplaceWithSignatures`. Two gaps:

1. **Silent unpaid work.** If the deliver simulation failed at tx-prep, the task was skipped, but `PostTxSettlementRound` handed off *every* pooled `done_task` as submitted. Next period the skipped task was pruned as if settled, reported to predict-api as delivered under a tx hash it was not in, and never retried. Off-chain tasks have no on-chain record to recover from, so the result went out and the mech was never paid.
2. **No off-chain observability.** Propel already scrapes every agent pod on `:9000`, but nothing distinguished on-chain from off-chain, and nothing counted HTTP ingress outcomes, deliveries, failures, latency, or settlement outcomes.

## What

**task_submission_abci**
- The tx-prep vote is now a JSON envelope `{tx_hash, included_request_ids}`. `TransactionPreparationRound` writes `most_voted_tx_hash` (unchanged value) plus a new `tx_included_request_ids` key; `PostTxSettlementRound` hands off only those ids.
- Skipped off-chain tasks stay in `done_tasks` and are retried for up to 3 periods; after that the executing agent drops the task locally, releases the requester's settling nonce, and counts `dropped`.
- predict-api delivered events are filtered to the same id list.
- New `mech_settlement_total{outcome,source,chain,mech_address}` (settled / sim_failed / dropped / contract_error), counted only by the agent that executed the task so a multi-agent service counts each task once. New `mech_offchain_unsettled_delivered` and `mech_offchain_unsettled_oldest_age_seconds` gauges (result served, not yet paid). `tool_delivery_time` gains a `source` label.

**task_execution**
- Per-tool task series gain `source` (onchain / offchain). `mech_tasks_failed_total{reason}` becomes bounded constants instead of free text with request ids and prices.
- New: `mech_offchain_requests_total{outcome}` and `mech_offchain_fetches_total{outcome}` at every terminal branch of the two HTTP routes, `mech_offchain_deliveries_total`, `mech_offchain_failures_total{reason}`, `mech_offchain_delivery_latency_seconds` (receive to fetchable), `mech_offchain_pending_requests`, `mech_offchain_pending_oldest_age_seconds`. Queue gauges refresh before the in-flight early return.
- `_discard_settling_nonce`, the settling-set counterpart of the accepted-set discard.

No env vars, ports, FSM rounds, or Propel changes. Metrics land in Grafana through the existing pod scrape.

## Rollout note

Old and new agents cannot agree on the tx-prep vote content, so a multi-agent service must have all its agents upgraded together (which is how Propel applies a hash bump). Do not roll one agent at a time.

## Also changed

- `source` on existing series changes their shape; nothing in mech, mech-scanner, propel, or agent-deployments consumes them today.
- Not addressed: the legacy non-marketplace deliver path still drops a task locally on its first failed simulation, as before.

## Verification

- 1041 tests pass across task_execution, task_submission_abci, mech_abci
- black / isort / flake8 / mypy / pylint / darglint, check-abciapp-specs, check-abci-docstrings, check-handlers, check-dependencies, check-doc-hashes, check-third-party-hashes, copyright, bandit, gitleaks (diff), uv lock --check, packages lock --check, check-hash, check-packages
- safety reports only pip / setuptools inside the tox env; liccheck could not run locally (tomte strategy file), no dependencies added


packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -302,6 +367,139 @@ def observe_histogram(self, metric: Histogram, value: float, **labels: Any) -> N
...
+        :param outcome: one of the ``SETTLEMENT_OUTCOME_*`` constants.
+        :param source: ``SOURCE_OFFCHAIN`` or ``SOURCE_ONCHAIN``.
+        :param mech_address: the mech the deliver targets.
+        :param request_ids: the task ids sharing this outcome.
+        """
+        self.count_settlement(
+            outcome, source, mech_address, len(self.local_request_ids(request_ids))
+        )
+
+    def note_offchain_settlement_skipped(
```
OjusWiZard: **Critical — the 3-period retry cap can burn inside a single period, dropping an unpaid off-chain task.**

`TransactionPreparationRound` self-loops on timeout (`rounds.py:549`, `Event.TASK_EXECUTION_ROUND_TIMEOUT: TransactionPreparationRound`, 60s). In `abstract_round_abci/behaviours.py::_process_current_round`, *any* round-height change — a self-loop included — calls `stop()` and instantiates a fresh behaviour, so `get_payload_content` -> `_get_offchain_tasks_deliver_data` -> `note_offchain_settlement_skipped` re-runs from scratch against the same local task dict, with no `db.create` in between.

`_get_offchain_tasks_deliver_data` does one sequential contract simulation per requester, so slow RPC is exactly what trips a 60s round timeout. Three timeout loops in one period take `SETTLEMENT_ATTEMPTS_KEY` 0 -> 3, hit `MAX_OFFCHAIN_SETTLEMENT_ATTEMPTS`, release the settling nonce and drop the task permanently. The design intent is "retried for up to 3 **periods**"; under transient RPC slowness this becomes 3 restarts of one round. The user has the result, the mech is never paid — the exact failure class this PR exists to close.

Suggested fix: commit the attempt increment from `TransactionPreparationRound.end_block` (fires once per settled period), or key the stamp by round height so a re-run at the same height is a no-op.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1976,6 +2289,33 @@ def async_act(self) -> Generator:
             yield from self.wait_until_round_end()
             self.set_done()
 
+    def count_settled_from_synced_data(self) -> None:
```
OjusWiZard, line 2323: **`mech_settlement_total{outcome="settled"}` double-counts on round restart.**

`PostTxSettlementRound` self-loops on both `NO_MAJORITY` and `ROUND_TIMEOUT` (`rounds.py:558-560`), deliberately. By the same behaviour-restart mechanism as the tx-prep round, `async_act` re-runs and calls this again against unchanged `tx_included_request_ids` / `done_tasks`, with no re-entry guard — so the counter increments once per loop instead of once per settled task, contradicting this method's own docstring ("counted exactly once fleet-wide, by its executor").

Worth noting the comment directly above that transition says *"The predict-api write is idempotent, so re-entry is safe"* — this PR adds a non-idempotent call right next to it in `async_act`, which invalidates that invariant.

Scoped to the Prometheus counter; consensus state and pruning are unaffected. Gating on something stable across a restart (e.g. `final_tx_hash` recorded in `shared_state`) would fix it.

OjusWiZard (commented):
Reviewed this across six passes (correctness/consensus, silent-failure, tests, type design, comments, simplification) and verified each finding against the code — CI is green and the test suite is 985 passing locally.

**The core design holds up.** I traced `tx_included_request_ids` through write (`TransactionPreparationRound.end_block`), reset (`TaskPoolingRound`, `_error_state`), consumption (`PostTxSettlementRound`) and the `db.create` boundary — correctly absent from `cross_period_persisted_keys`. Predict-api events are deferred rather than lost, since `PREDICT_API_EVENTS` is pruned only via `submitted_request_ids`. Locking around `shared_state[DONE_TASKS]` is right, `_discard_settling_nonce` correctly mirrors `_discard_outstanding_nonce`, a mixed old/new fleet stalls on `NO_MAJORITY` rather than corrupting state, and there are no Prometheus metric-name collisions across the two skills.

The issues are one family: **non-idempotent side effects added to behaviours the FSM can legitimately re-run within a single period.**

1. **Critical** — the off-chain retry cap can burn in ~3 minutes instead of 3 periods, dropping an unpaid task (`behaviours.py:441`).
2. **Important** — `settled` double-counts on `PostTxSettlementRound` self-loop (`behaviours.py:2292`).
3. **Important** — `contract_error` double-counts ids already counted `sim_failed` (`behaviours.py:2033`, `:2216`).
4. **Important** — the legacy non-marketplace path labels a permanent drop as `sim_failed` (`behaviours.py:1732`).
5. **Important** — no test exercises intra-period round re-entry, which is what hides 1 and 2.

Plus some advisory notes on label typing, a duplicated constant, and comment density.

(1) is the one I'd block on: it reintroduces the "off-chain work delivered but never paid for" outcome this PR was written to eliminate, under nothing more exotic than a slow RPC.

Rest is solid work — the envelope approach is the right fix and the test suite around it is unusually thorough.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1758,15 +2021,30 @@ def _get_offchain_tasks_deliver_data(
...
+                    # Every off-chain task in this period's batch is
+                    # abandoned (including groups already appended to
+                    # ``tx_list``, since the caller discards it on None).
+                    self.count_local_settlement(
+                        SETTLEMENT_OUTCOME_CONTRACT_ERROR,
+                        SOURCE_OFFCHAIN,
+                        self.metrics_mech_label(),
+                        extract_request_ids(offchain_done_tasks_list),
+                    )
+                    return None, []
```
OjusWiZard, line 2057: **`contract_error` double-counts ids already counted `sim_failed` in the same batch.**

The loop counts a failing group as `sim_failed` then `continue`s. If a *later* group's contract call errors, this counts the entire `offchain_done_tasks_list` as `contract_error` — including groups already counted under a different outcome, which are still in local `DONE_TASKS` and so still pass the `local_request_ids` filter.

Scenario: sender A fails simulation (counted `sim_failed`), sender B's `get_offchain_deliver_data` then returns a non-`STATE` performative -> A's ids are counted a second time as `contract_error`. Inflated `contract_error` reads as an RPC/infra problem when the real cause for those ids was a simulation failure.

Accumulate the already-counted ids and subtract them here.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1913,15 +2203,37 @@ def _num_to_bytes(value: int) -> bytes:
...
+                        failed_mech,
+                        failed_details,
+                    ) in marketplace_deliver_by_mech.items():
+                        self.count_local_settlement(
+                            SETTLEMENT_OUTCOME_CONTRACT_ERROR,
+                            SOURCE_ONCHAIN,
+                            failed_mech,
+                            failed_details[REQUEST_IDS_KEY],
+                        )
+                    return None, []
```
OjusWiZard, line 2249: Same double-count as the off-chain path: this iterates **all** of `marketplace_deliver_by_mech`, including groups already counted `sim_failed` earlier in the loop before a later group's contract call failed. Exclude ids already counted.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1487,11 +1729,18 @@ def get_payload_content(self) -> Generator[None, None, str]:
                 self.context.logger.warning(
                     f"Deliver tx simulation failed for task {task}. Skipping this deliver."
                 )
+                self.count_local_settlement(
```
OjusWiZard: **This labels a permanent drop as `sim_failed`, whose documented meaning is "retryable".**

Two lines below, `self.remove_tasks([task])` deletes the task on the *first* failed simulation — no cap, no retry (pre-existing, and the PR body calls it out as not-addressed, which is fine). What's new is the label: `SETTLEMENT_OUTCOME_SIM_FAILED` is defined in this PR (`behaviours.py:177-194`) as "the task stays in `done_tasks` for a retry next period."

So a permanent unpaid loss on the legacy non-marketplace path is now indistinguishable on a dashboard from a transient retried one, and invisible in `outcome="dropped"` — the metric built to surface exactly this. Either give this path the retry treatment, or give it a distinct outcome value.

packages/valory/skills/task_submission_abci/tests/test_behaviours.py
```diff
@@ -1075,6 +1089,390 @@ def test_end_to_end_prune_using_real_synchronized_data(self) -> None:
...
+        assert local[0][SETTLEMENT_ATTEMPTS_KEY] == 1
+        assert b.context.shared_state[SETTLING_NONCES_BY_SENDER] == {"0xSENDER": {7, 8}}
+        mock_count.assert_any_call(
+            SETTLEMENT_OUTCOME_SIM_FAILED, SOURCE_OFFCHAIN, "0xMECH", 1
+        )
+        mock_count.assert_any_call(
+            SETTLEMENT_OUTCOME_DROPPED, SOURCE_OFFCHAIN, "0xMECH", 0
+        )
+
+    def test_reaching_the_cap_drops_task_and_releases_settling_nonce(self) -> None:
```
OjusWiZard, line 1427: **Test gap: nothing exercises intra-period round re-entry.**

Coverage in this PR is otherwise genuinely strong — the cap boundary on both sides, the drop path, selective nonce release, malformed-envelope parametrization, the multi-agent `task_executor_address` invariant, and `_validated_str_list` degradation are all pinned with real assertions rather than call-count checks.

But every test here calls `note_offchain_settlement_skipped` exactly once per period, which bakes in the one-call-per-period assumption that the `TransactionPreparationRound` self-loop breaks (see my comment on `behaviours.py:441`). A test that runs the behaviour twice at the same round height and asserts the attempt counter advanced only once would cover both that and the `count_settled_from_synced_data` double-count.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -1113,6 +1155,34 @@ def ipfs_tasks(self) -> List[Dict[str, Any]]:
...
+        """Return the ``chain`` / ``mech_address`` labels for the off-chain counters.
+
+        :return: the label kwargs.
+        """
+        return {
+            "chain": str(self.params.default_chain_id),
+            "mech_address": metrics_mech_address(self.params),
+        }
+
+    def _count_request_outcome(self, outcome: str) -> None:
```
OjusWiZard: `outcome: str` here (and on `_count_fetch_outcome`, `_send_rejection_response`, and `_record_offchain_failure`'s `metric_reason`) gives mypy nothing to enforce on a labelled Prometheus counter.

Every current caller passes the right `REQUEST_OUTCOME_*` / `FETCH_OUTCOME_*` / `OFFCHAIN_FAILURE_*` constant, so there's no live bug — but this PR introduced `SettlementOutcome` and `Source` as `Literal`s precisely so "a misspelt outcome fails mypy instead of silently minting a new label value," and those four families didn't get the same treatment. Mirroring the pattern is cheap.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -164,6 +173,62 @@
...
+# Per-task local bookkeeping key: how many tx-prep rounds have skipped
+# this off-chain task because its deliver simulation failed. Stamped on
+# the executing agent's ``shared_state[DONE_TASKS]`` entry only (other
+# agents do not hold the task locally), never on the synced copy.
+SETTLEMENT_ATTEMPTS_KEY = "settlement_attempts"
+MAX_OFFCHAIN_SETTLEMENT_ATTEMPTS = 3
+# Wall-clock receive stamp set by ``task_execution.handlers`` on every
+# off-chain pending task; survives into ``done_tasks`` via the
+# ``**executing_task`` spread in ``_finalize_done_task``.
+ENQUEUED_AT_LOCAL = "enqueued_at_local"
```
OjusWiZard: `ENQUEUED_AT_LOCAL` is already defined at `task_execution/behaviours.py:119`, and this module's import block a few lines up already pulls `Source`, `SOURCE_ONCHAIN`, `SOURCE_OFFCHAIN` and `_discard_settling_nonce` from that same module specifically so the two skills can't drift. Adding it to that import instead of re-declaring the literal would keep the convention — nothing would catch the two copies diverging, and the failure mode is a silent `0` gauge.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -893,8 +1008,53 @@ def _ensure_deadline(self) -> None:
                 f"Deadline set to {self._request_handling_deadline} for task {self._executing_task}."
             )
 
+    def _offchain_metric_labels(self) -> Dict[str, str]:
```
OjusWiZard, line 1020: This helper is byte-identical to `handlers.py:1158`, and `task_submission_abci/behaviours.py:548` hand-rolls the same dict a third time. A single `offchain_metric_labels(params)` next to `metrics_mech_address` in `models.py` would cover all three.

packages/valory/skills/task_execution/models.py
```diff
@@ -59,6 +59,27 @@ class RequestParams:
     )
 
 
+def metrics_mech_address(params: Any) -> str:
```
OjusWiZard: `params: Any` erases a concretely typed `Params` at every production call site (`self.params` is already `cast(Params, ...)` in both skills) purely so tests can pass a `SimpleNamespace`. A `Protocol` gets both — `SimpleNamespace` still satisfies it structurally, and mypy would catch a rename of `mech_to_config` / `is_marketplace_mech`.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -164,6 +173,62 @@
...
+#   itself failed, the whole batch for that path was abandoned this
+#   period).
+# * ``source``: ``"offchain"`` / ``"onchain"`` (``task_execution.behaviours.Source``).
+#
+# Granularity caveat: the mech simulates one call per requester (off-chain)
+# or per mech (on-chain marketplace), so a ``sim_failed`` burst of N is
+# usually one failing group, not N independent failures.
+#
+# ``SettlementOutcome`` is a ``Literal`` so a misspelt outcome fails mypy
+# instead of silently minting a new label value.
```
OjusWiZard, line 183: Nit, and a house-style call rather than a defect: this block, plus the ones on `SETTLEMENT_ATTEMPTS_KEY`, `ENQUEUED_AT_LOCAL`, `REQUEST_IDS_KEY`, `mech_offchain_unsettled_delivered` and `update_offchain_unsettled_gauges`, is rationale the PR description and `git blame` already carry — it'll rot the first time someone edits a label. The label list and the granularity caveat would survive better in the metric's `documentation=` string or the PR body.

The consensus-safety notes elsewhere in the diff (why `_validated_str_list` degrades instead of raising, why a malformed envelope becomes `ERROR`) do earn their place — that reasoning genuinely isn't recoverable from the code.

packages/valory/skills/task_submission_abci/rounds.py
```diff
@@ -60,6 +62,108 @@ def extract_request_ids(tasks: List[Dict[str, Any]]) -> List[str]:
     ]
 
 
+# ``TransactionPreparationRound`` payload format. The behaviour used to
+# vote on the bare multisend hex; it now votes on a JSON envelope that
+# also carries the request ids that made it into the multisend, so
+# ``PostTxSettlementRound`` can record exactly those (and not every
+# ``done_task``) as submitted. Tasks skipped at tx-prep time (deliver
+# simulation failed) stay in ``done_tasks`` and are re-pooled next
+# period instead of being pruned as though they had settled.
```
OjusWiZard: Same note: the "used to vote on the bare multisend hex, now votes on a JSON envelope" history is what the commit and PR description are for. The `TypedDict` plus `encode`/`decode` pair already says what the format is; this paragraph only restates why it changed, and will read as stale once the change is a year old.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1913,15 +2203,37 @@ def _num_to_bytes(value: int) -> bytes:
...
+                            failed_mech,
+                            failed_details[REQUEST_IDS_KEY],
+                        )
+                    return None, []
 
                 data_ = cast(bytes, contract_api_msg.state.body["data"])
                 simulation_ok = cast(bool, contract_api_msg.state.body["simulation_ok"])
+                group_request_ids = details[REQUEST_IDS_KEY]
                 if not simulation_ok:
+                    # On-chain requests stay undelivered on the contract and
```
jmoreira-valory: 🟠 **F1 · HIGH — On-chain marketplace skip has no cap, so one bad request stops the mech**

Before this PR, `PostTxSettlementRound` handed off `extract_request_ids(done_tasks)`, so the next `handle_submitted_tasks` pruned every pooled task out of `shared_state[DONE_TASKS]`, including ones whose deliver simulation had failed. Now only `tx_included_request_ids` is handed off, and this branch neither removes the task (the legacy branch at `behaviours.py:1739` does) nor stamps an attempt counter (the off-chain branch at line 2043 does). Nothing else deletes from `DONE_TASKS`: the writers are `_finalize_done_task`'s append, `remove_tasks_by_id` driven by `submitted_request_ids`, `remove_tasks` on the legacy path, and `note_offchain_settlement_skipped`'s drop.

`get_marketplace_deliver_data` is called once per mech with the whole group's `requestIds` array, so one request that reverts in simulation sets `simulation_ok` False for every task grouped under that mech. The group is rebuilt from `synchronized_data.done_tasks` every period, so once a permanently failing id is in it, on-chain delivery for that mech stops for good. A plausible way in: the settlement app takes `FailedRound` but the tx lands anyway, and the request is already delivered on the marketplace from then on. The only signal is an INFO log and a rising `mech_settlement_total{outcome="sim_failed"}`.

The comment's rationale does not hold either: `_handle_get_undelivered_reqs` sets `self.from_block = max(req["block_number"] for req in reqs) + 1` on every read, so a request below the cursor is never re-fetched.

Give this branch the same treatment as the off-chain one: stamp `SETTLEMENT_ATTEMPTS_KEY` on the local task and drop it after `MAX_OFFCHAIN_SETTLEMENT_ATTEMPTS`, counted as `dropped` with `source="onchain"`.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1758,15 +2021,30 @@ def _get_offchain_tasks_deliver_data(
...
                 data_ = cast(bytes, contract_api_msg.state.body["data"])
                 simulation_ok = cast(bool, contract_api_msg.state.body["simulation_ok"])
+                group_request_ids = details[REQUEST_IDS_KEY]
                 if not simulation_ok:
                     self.context.logger.info(
-                        f"Simulation failed for offchain dropping the transaction: {contract_data}"
+                        f"Simulation failed for offchain deliver of request_ids="
+                        f"{group_request_ids}; skipping this period: {contract_data}"
+                    )
+                    self.note_offchain_settlement_skipped(
```
jmoreira-valory: 🟡 **F2 · MEDIUM — Retry cap is charged to the whole sender group, not the failing request**

`offchain_list_by_sender` groups by requester and `get_offchain_deliver_data` simulates the group in one call, so `simulation_ok=False` says nothing about which request caused it. `note_offchain_settlement_skipped` then charges an attempt to every id in `group_request_ids`, so after three periods the whole group is dropped together and counted `dropped`. With one unrecoverable request and four that would settle on their own, the mech loses payment for all five, and `mech_settlement_total{outcome="dropped"}` cannot separate the cause from the collateral.

Some causes really are group-wide, for instance the requester's escrow draining between accept and settle, and dropping the group is right for those. Per-request causes such as a nonce the chain already consumed are not.

Cheapest improvement: on reaching the cap, drop one task rather than the whole group and leave the rest at their current attempt count, so a poisoned group recovers within `len(group)` periods instead of losing everything. Whichever way it goes, please record the blast radius next to `MAX_OFFCHAIN_SETTLEMENT_ATTEMPTS`, so an operator reading `dropped` knows it counts collateral.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -1697,12 +1956,16 @@ def _get_offchain_tasks_deliver_data(
                 lambda: {
                     OffchainKeys.DELIVER_WITH_SIGNATURES.value: [],
                     OffchainKeys.DELIVERY_RATES.value: [],
+                    REQUEST_IDS_KEY: [],
                 }
             )
+            # uses the first mech in config as marketplace mech
+            mech_address = self.mech_addresses[0]
```
jmoreira-valory, line 1982: 🔵 **F3 · LOW — Off-chain metrics label a different mech than the deliver targets**

The deliver is built against `self.mech_addresses[0]`, the first key of `mech_to_config`, while both metric calls on this path (the `contract_error` count at line 2036 and `note_offchain_settlement_skipped` at line 2043) label with `self.metrics_mech_label()`, which returns the first entry flagged `is_marketplace_mech` and only falls back to `agent_mech_contract_address`. When the first configured mech is not the marketplace mech, the two disagree and `mech_settlement_total{source="offchain"}` points at a mech the transaction never touched.

The behaviour-preserving fix is to label with the address the call actually used: pass `mech_address` instead of `self.metrics_mech_label()` at both sites. If `[0]` is instead the wrong contract target, that is a larger change and the `# uses the first mech in config as marketplace mech` assumption is what needs revisiting.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -164,6 +173,62 @@
...
+)
+mech_offchain_unsettled_oldest_age_seconds = Gauge(
+    "mech_offchain_unsettled_oldest_age_seconds",
+    "Age of the oldest off-chain result still awaiting on-chain settlement",
+    labelnames=["chain", "mech_address"],
+)
+# Per-task local bookkeeping key: how many tx-prep rounds have skipped
+# this off-chain task because its deliver simulation failed. Stamped on
+# the executing agent's ``shared_state[DONE_TASKS]`` entry only (other
+# agents do not hold the task locally), never on the synced copy.
```
jmoreira-valory: 🔵 **F4 · LOW — settlement_attempts does reach the synced copy, unlike what the comment says**

`get_done_tasks` returns `self.done_tasks`, which is `deepcopy(shared_state[DONE_TASKS])`, and `get_payload_content` JSON-dumps that straight into the `TaskPoolingPayload`. So once `note_offchain_settlement_skipped` stamps `settlement_attempts` on a local task, the key rides into `synchronized_data.done_tasks` from the next period onward.

No impact today: the counter is only read back off the local list, and `TaskPoolingRound` merges payloads by request id rather than requiring identical bytes, so consensus is unaffected. But the comment states the opposite of what the code does, and the next person to add a synced-copy consumer will read it and believe it.

```suggestion
# agents do not hold the task locally). It does ride into the synced
# ``done_tasks`` via the TaskPooling payload, but nothing reads it there.
```

jmoreira-valory (changes requested):

**What this does.** Off-chain mech results are served to the user over HTTP and paid for later, when the deliver is batched into a Safe multisend. Until now the post-settlement round handed off every pooled done_task as submitted, so a task whose deliver simulation failed at tx-prep was pruned as if it had settled, reported to predict-api under a tx hash it was not in, and never retried. This PR makes the tx-prep round vote on a JSON envelope of {tx_hash, included_request_ids} so only the ids actually in the multisend are handed off and pruned, keeps skipped off-chain tasks queued for up to 3 periods before dropping them and releasing the requester's settling nonce, and adds a set of Prometheus series for the off-chain HTTP routes, deliveries, latency and settlement outcomes.

CI is green at this HEAD (22 check runs, all success) and the SHA still matches the briefing.

I checked the parts the diff does not show. The FSM spec, transition function and AbciApp docstring are unchanged and mutually consistent, and all four final states (`FinishedTaskPoolingRound`, `FinishedTaskExecutionWithErrorRound`, `FinishedWithoutTasksRound`, `FinishedPostTxSettlementRound`) are still mapped in `mech_abci/composition.py`. `tx_included_request_ids` is correctly absent from `cross_period_persisted_keys` and is zeroed on all three reset paths (`TaskPoolingRound.end_block`, and `_error_state` on both ERROR and NO_MAJORITY). `REQUEST_IDS_KEY = "request_ids"` does not collide with `MarketplaceKeys.REQUEST_IDS = "requestIds"` in either grouping dict. Both skills' `Params` expose `mech_to_config`, `agent_mech_contract_address` and `default_chain_id`, so `metrics_mech_address` resolves on both sides rather than degrading to an empty label. Every call site of the five relabelled series passes the new `source` label, and I walked every return in `_handle_signed_requests_impl`: each one increments exactly one `mech_offchain_requests_total` outcome, including the two that route through `_send_internal_error`.

The delivery-rate settlement cannot pick up a stale `tx_included_request_ids`, since `UpdateDeliveryRateRound` is only reachable straight out of registration, by which point the db has been recreated.

@OjusWiZard's twelve comments cover the round-restart double counting, the `contract_error` overlap on both deliver paths, the `sim_failed` label on the legacy permanent drop, and the typing and duplication points. None of those are re-raised here; the four below are separate.

**Findings** — 4 · 🟠 1 High · 🟡 1 Medium · 🔵 2 Low

| F1 | 🟠 HIGH | [packages/valory/skills/task_submission_abci/behaviours.py:2222](https://github.com/valory-xyz/mech/blob/73b5d8e7651f5c8d87e67f3f19414a21e304656d/packages/valory/skills/task_submission_abci/behaviours.py#L2222) | **On-chain marketplace skip has no cap, so one bad request stops the mech** |
| F2 | 🟡 MEDIUM | [packages/valory/skills/task_submission_abci/behaviours.py:2043](https://github.com/valory-xyz/mech/blob/73b5d8e7651f5c8d87e67f3f19414a21e304656d/packages/valory/skills/task_submission_abci/behaviours.py#L2043) | **Retry cap is charged to the whole sender group, not the failing request** |
| F3 | 🔵 LOW | [packages/valory/skills/task_submission_abci/behaviours.py:1963](https://github.com/valory-xyz/mech/blob/73b5d8e7651f5c8d87e67f3f19414a21e304656d/packages/valory/skills/task_submission_abci/behaviours.py#L1963) | **Off-chain metrics label a different mech than the deliver targets** |
| F4 | 🔵 LOW | [packages/valory/skills/task_submission_abci/behaviours.py:221](https://github.com/valory-xyz/mech/blob/73b5d8e7651f5c8d87e67f3f19414a21e304656d/packages/valory/skills/task_submission_abci/behaviours.py#L221) | **settlement_attempts does reach the synced copy, unlike what the comment says** |

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Good catch, confirmed the self-loop rebuilds the behaviour. The helper now takes the synced period_count and charges one attempt per period, so a re-run at the same height is a no-op and counts nothing. Test test_cap_burns_only_across_periods_never_within_one runs it MAX+1 times in one period and asserts the task is still queued at attempt 1. Fixed in e6650ba1.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in e6650ba1. count_settled_from_synced_data records the final_tx_hash it counted in shared_state and returns early on the same hash, so the NO_MAJORITY / timeout loop counts once. Covered by test_round_re_entry_counts_the_same_tx_once and test_a_new_tx_hash_is_counted_again.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed in e6650ba1. Both builders keep the ids already charged sim_failed and exclude them from the contract_error count. test_contract_error_after_sim_failure_does_not_recount_those_ids covers the sender A fails sim, sender B errors case.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Same fix as the off-chain path, in e6650ba1.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Agreed. The legacy path now goes through the same note_settlement_skipped helper as the other two, so it is retried up to the cap and counted dropped when removed, instead of a first-failure removal labelled sim_failed. In e6650ba1.

packages/valory/skills/task_submission_abci/tests/test_behaviours.py
DIvyaNautiyal07: Added in e6650ba1: test_re_entry_in_the_same_period_charges_nothing and test_cap_burns_only_across_periods_never_within_one for the attempt counter, and test_round_re_entry_counts_the_same_tx_once for the settled counter.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Done in e6650ba1: RequestOutcome, FetchOutcome, FailureReason and OffchainFailureReason are Literals and the four signatures use them.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Done, ENQUEUED_AT_LOCAL is imported from task_execution.behaviours now.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Done, offchain_metric_labels(params) lives next to metrics_mech_address in models.py and all three sites call it.

packages/valory/skills/task_execution/models.py
DIvyaNautiyal07: Done, MetricsParams Protocol with the three attributes. The tests cast their SimpleNamespace stubs to it.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Trimmed. The label semantics and the granularity caveat moved into the metric documentation strings, and the remaining comments are one-liners.

packages/valory/skills/task_submission_abci/rounds.py
DIvyaNautiyal07: Removed, the TypedDict plus encode/decode say what the format is.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: You are right, and thanks for the from_block cursor detail, that killed my assumption that the chain re-fetch would recover these. The marketplace path now uses the same stamp-and-drop helper as the off-chain one, counted dropped with source=onchain. In e6650ba1.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Went with the cheap version: at the cap only one task per group is dropped per period, lowest wire nonce first (or first by position for on-chain groups), and the rest keep their count and are retried. The blast radius note sits next to MAX_SETTLEMENT_ATTEMPTS and in the counter's documentation string. test_only_the_lowest_nonce_at_cap_is_dropped_per_period covers it. In e6650ba1.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed, both count sites on the off-chain path now label with the mech_address the deliver call actually used.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Thanks, the comment was wrong. Rewritten to say the keys ride into the synced copy through the pooling payload and nothing reads them there.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -302,6 +363,165 @@ def observe_histogram(self, metric: Histogram, value: float, **labels: Any) -> N
...
+        :param mech_address: the mech the deliver targets.
+        :param amount: number of tasks sharing this outcome.
+        """
+        if amount <= 0:
+            return
+        mech_settlement_total.labels(
+            outcome=outcome,
+            source=source,
+            chain=self.params.default_chain_id,
+            mech_address=mech_address,
```
jmoreira-valory: 🔵 **F5 · LOW — mech_address label case differs between the two skills**

`task_execution.Params` lowercases the `mech_to_config` keys (`key.lower()` in `task_execution/models.py`), while `task_submission_abci.Params` keeps them as configured. The README's `MECH_TO_CONFIG` example uses a checksummed address (`0xFf82123dFB52ab75C417195c5fDB87630145ae81`). With a config like that, the off-chain series disagree on the label value:

- `mech_settlement_total{source="offchain",outcome="settled"}` takes `task["mech_address"]`, which `_handle_done_task` fills from task_execution's lowercased marketplace key.
- `sim_failed`, `dropped` and `contract_error` on the same path take `self.mech_addresses[0]` from task_submission_abci, in the configured case.
- `mech_offchain_unsettled_delivered` and `mech_offchain_unsettled_oldest_age_seconds` go through `offchain_metric_labels` with task_submission_abci's params (configured case), while the task_execution off-chain series get the lowercased key.

A `sum by (mech_address)` then shows one mech as two rows, and a settle-rate panel that divides `settled` by the failure outcomes finds no matching series. The new `offchain_metric_labels` helper shares the function between the skills, but the input still differs. The split was already there at `73b5d8e7` and I missed it in round 1; the shared helper is now the one place to close it.

Fix: lowercase the label in `count_settlement` here, and in `metrics_mech_address` (`return str(mech).lower()` and `return str(params.agent_mech_contract_address).lower()`), so the settlement counter and the gauges match the task_execution series.

```suggestion
            mech_address=mech_address.lower(),
```

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -302,6 +363,165 @@ def observe_histogram(self, metric: Histogram, value: float, **labels: Any) -> N
...
+            self.context.logger.warning(
+                "%s request_id=%s (sender=%s nonce=%s tool=%s) dropped after %d "
+                "periods of failed deliver simulations; the delivery will not be "
+                "settled on-chain.",
+                source,
+                dropped.get("request_id"),
+                dropped.get(SENDER),
+                dropped.get(NONCE),
+                dropped.get("tool"),
+                MAX_SETTLEMENT_ATTEMPTS,
```
jmoreira-valory: ⚪ **F6 · NIT — Drop log reports the cap, not the task's attempt count**

With one drop per group per period, a task can stay at the cap for several periods before its turn: in `test_only_the_lowest_nonce_at_cap_is_dropped_per_period`, `r-late` survives at `MAX_SETTLEMENT_ATTEMPTS` and is charged again next period. The warning still prints `MAX_SETTLEMENT_ATTEMPTS`, so a task dropped after 6 failed periods logs "after 3 periods". Printing the task's own count keeps the log accurate.

```suggestion
                dropped.get(SETTLEMENT_ATTEMPTS_KEY),
```

jmoreira-valory (dismissed):

**What this does.** Off-chain mech results are served over HTTP and paid for later, when the deliver lands in a Safe multisend. This PR makes the tx-prep round vote on the multisend plus the request ids it actually delivers, so a task whose deliver simulation failed is no longer pruned as if it had settled. Skipped tasks (off-chain, on-chain marketplace and legacy) stay queued, and once a task has failed in 3 periods it can be dropped, at most one per group per period. It also adds Prometheus series for the off-chain HTTP routes, deliveries, latency and settlement outcomes.

The push since round 1 is one commit, `e6650ba1`, on top of `73b5d8e7` (compare status `ahead`, 14 files). CI at HEAD is 22 check runs, all success.

All four round 1 findings are fixed at HEAD. The on-chain marketplace skip now goes through `note_settlement_skipped` with `SOURCE_ONCHAIN` (F1). At the cap only one task per group is dropped per call, lowest wire nonce first, and the blast radius is written next to `MAX_SETTLEMENT_ATTEMPTS` and in the `mech_settlement_total` help text (F2). Both off-chain count sites label with the `mech_address` the deliver call used (F3), and the `SETTLEMENT_ATTEMPTS_KEY` comment now says the keys ride into the synced `done_tasks` (F4).

On the re-entry fixes from @OjusWiZard's review: the attempt charge is keyed on `synchronized_data.period_count`, which is `db.reset_index` in open-autonomy's `abstract_round_abci/base.py` and only advances on `AbciAppDB.create()`, so the tx-prep timeout self-loop sees the same period and charges nothing twice. `count_settled_from_synced_data` is keyed on `final_tx_hash`, and both `contract_error` sites now skip ids already charged `sim_failed`. A task dropped locally also leaves the synced group the next period, because `TaskPoolingRound.end_block` rebuilds `done_tasks` from that period's payloads and `done_tasks` is not in `cross_period_persisted_keys`.

`metrics_mech_address` lost its `getattr` fallbacks. Both skills' `Params` always set `mech_to_config` and `agent_mech_contract_address` (an empty config raises `ValueError` at setup), so the strict attribute access does not add a crash path. `_discard_settling_nonce` returns early when the task has no sender or nonce, so routing on-chain drops through it is safe. The only change to `rounds.py` in this push is a removed comment, and the four final states are still mapped in `mech_abci/composition.py`.

The tests named in the replies are present at HEAD (`test_re_entry_in_the_same_period_charges_nothing`, `test_cap_burns_only_across_periods_never_within_one`, `test_only_the_lowest_nonce_at_cap_is_dropped_per_period`, `test_round_re_entry_counts_the_same_tx_once`, `test_contract_error_after_sim_failure_does_not_recount_those_ids`). I read them but did not run them locally.

**Findings** — 2 · 🔵 1 Low · ⚪ 1 Nit

| F5 | 🔵 LOW | [packages/valory/skills/task_submission_abci/behaviours.py:386](https://github.com/valory-xyz/mech/blob/e6650ba1951d64910540c96e08e871e0136d9e1c/packages/valory/skills/task_submission_abci/behaviours.py#L386) | **mech_address label case differs between the two skills** |
| F6 | ⚪ NIT | [packages/valory/skills/task_submission_abci/behaviours.py:504](https://github.com/valory-xyz/mech/blob/e6650ba1951d64910540c96e08e871e0136d9e1c/packages/valory/skills/task_submission_abci/behaviours.py#L504) | **Drop log reports the cap, not the task's attempt count** |

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Good catch, confirmed task_submission keeps the configured case. metrics_mech_address now lower-cases both the marketplace key and the fallback, and count_settlement lower-cases the label it is handed, so every series lands on one value. test_metrics_mech_address_is_lower_cased_regardless_of_config_case pins it with a checksummed address. In c5abda02.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Done in c5abda02, the warning prints the task's own attempt count.

jmoreira-valory (dismissed):

**What this does.** Off-chain mech results are served over HTTP and paid for later in a Safe multisend. This PR makes the tx-prep vote carry the request ids it actually delivers, so a task whose deliver simulation failed is retried (and eventually dropped, with its settling nonce released) instead of being pruned as settled, and it adds Prometheus metrics that split on-chain from off-chain work. The latest push only addresses the two points left from round 2.

Delta since e6650ba1 checked at c5abda02. F5: `metrics_mech_address` now lower-cases both the marketplace key and the fallback, and `count_settlement` lower-cases the label it receives. The other label sites (`offchain_metric_labels` in handlers.py and task_execution/behaviours.py, the unsettled gauges, and the legacy-path `task.get(MECH_ADDRESS) or metrics_mech_label()` fallback) all go through one of those two, so one mech gets one label value. Contract calls still use the configured address. The new `test_metrics_mech_address_is_lower_cased_regardless_of_config_case` covers a checksummed key and the fallback. F6: the drop warning now prints the task's own `settlement_attempts`. The rest of the delta is the package hash cascade (packages.json, aea-config.yaml, service.yaml, skill.yaml). CI at HEAD: 22/22 check runs succeeded.

**No findings.**

jmoreira-valory (approved):

**What this does.** This PR makes the mech settle only the off-chain tasks that actually landed in the multisend, instead of pruning every pooled task as if it had been paid, and it adds Prometheus metrics that split on-chain from off-chain work so Grafana can see HTTP ingress, delivery, and settlement outcomes separately.

Re-review at HEAD 14b6c1d6. My round-2 approval (c5abda02) was dismissed only because the branch moved; the delta since then is a small, self-contained addition: a `mech_terms_url` param and an RFC 8288 `Link: <url>; rel="terms-of-service"` header (plus a `termsUrl` field in the 402 body) surfaced on the off-chain 402 and 200 responses, defaulting to the Valory Mech Terms URL and blank-disableable.

Verified against source at HEAD: `_build_terms_link_header` returns either `""` or a newline-terminated line, matching the `extra_headers` newline-guard contract both call sites rely on (`handlers.py:2564`, `2617`); the `_build_402_challenge` dict is now correctly annotated `Dict[str, Any]` so mypy accepts the conditional `termsUrl` key added after the literal; `Params.mech_terms_url` strips whitespace and defaults to `""` on `None`/absent, matching the tests. `mech_terms_url` is skill.yaml-only (no service.yaml env override), but that mirrors the existing `use_offchain`-style convention for this skill rather than introducing a new gap.

CI is green (22/22 check-runs), HEAD matches the briefing, and none of the fleet-wide footgun classes (receipt handling, Safe execTransaction, TypedDict total=False, chain-keyed dict lookups, EIP-150/191, FSM sync) apply to this delta — it touches no FSM, contract, or signature code.

**No findings.**

OjusWiZard (approved):
LGTM

PR #495, opened 2026-09-16
feat(task_execution): advertise the operator's terms on off-chain responses

DIvyaNautiyal07:
## Proposed changes

Part of the Mech Terms engineering work (identification and notice). The off-chain HTTP server now tells a requester where the operator's terms are, at the two points where the requester acts:

- **402 payment challenge**: `termsUrl` added to the structured challenge body next to the deposit instructions, and a `Link: <url>; rel="terms-of-service"` header (RFC 8288 relation) alongside `WWW-Authenticate`.
- **200 accept response**: the same `Link` header next to `Payment-Receipt`.

**Where the link comes from.** A new `task_execution` param, `mech_terms_url`, defaulted in `skill.yaml` to `https://www.valory.xyz/terms/mechs`.

It is defaulted in the skill rather than set per service on purpose. Carrying it in `service.yaml` would mean a line in every service definition built on this skill: the demo service here, plus `mech_predict` and `mech_agents_fun`, which are what actually run. Any one of them missing the line would silently advertise nothing. With the skill default, a service picks it up on its next version bump with no config change and no env var in agent-deployments. An operator running their own mech points the param at their own terms, or blanks it to advertise nothing.

Operators keep the value equal to the `termsUrl` in the mech's published metadata; the agent-deployments CI already pins that field.

I considered reading the link from the mech's own metadata hash instead, so there is one source of truth. That hash lives in the `mech_abci` params, not `task_execution`, and reading it means an IPFS round trip at startup with its own failure modes for an informational header. A param keeps the server path free of network calls; the metadata guard in agent-deployments keeps the two in sync.

**Not the delivered result.** The doc's separate item, the terms link inside every delivered result document, was dropped as after the fact. Both responses here come before the work: the 402 arrives while the requester still has to deposit, and the 200 is the acceptance of the request.

**Off path.** When the param is blank, no header is emitted and no `termsUrl` key is written; a test pins both, so a refactor that always emits the header with an empty URL fails.

## Fixes

Part of the Mech Terms engineering work (identification and notice).

## Types of changes

- [x] Non-breaking feature (non-breaking change which adds functionality)

## Checklist

- [x] Lint and unit tests pass locally with my changes (973 tests across the three skills; black, isort, flake8, mypy, pylint, darglint, check-hash, check-packages, check-abciapp-specs, check-abci-docstrings, check-dependencies, check-doc-hashes, check-third-party-hashes)
- [x] I have added tests that prove my fix is effective or that my feature works

## Further comments

Package hashes relocked with `autonomy packages lock`. Deploying this needs a mech release and a redeploy of the mechs; the header only appears on off-chain requests.


packages/valory/skills/task_execution/models.py
```diff
@@ -144,6 +144,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         # Default False = on-chain + IPFS only, no analytics writes,
         # unchanged for legacy deployments. Set True to enrol the mech.
         self.use_offchain: bool = bool(kwargs.get("use_offchain", False))
+        # Terms link the off-chain HTTP server advertises: as ``termsUrl`` in
+        # the 402 challenge body and as a ``Link: <url>; rel="terms-of-service"``
+        # header on the 402 and the 200. The skill default is the Valory Mech
+        # Terms; an operator running their own mech sets it to their own terms,
+        # or blank to advertise nothing. Keep it equal to the ``termsUrl`` in
+        # the mech's published metadata.
+        self.mech_terms_url: str = str(kwargs.get("mech_terms_url", "") or "").strip()
```
jmoreira-valory, line 153: ⚪ **F1 · NIT — mech_terms_url only strips the ends; inner CR/LF breaks header framing**

`.strip()` only removes whitespace at the start and end. A value with an embedded `\r`/`\n` (a YAML multi-line scalar pasted by mistake, or an env override once one exists) passes the guard, and `_build_terms_link_header` then writes a second, malformed header line. The senders only check the *trailing* newline, so the mistake would show up on every 402 and 200 at runtime, and a 200 is only sent after the task is already enqueued. The value is operator-controlled, so this is hardening, not a security issue. A startup check would catch it:

```python
if any(c in self.mech_terms_url for c in "\r\n<>"):
    raise ValueError("mech_terms_url must be a single-line URL without '<' or '>'")
```

jmoreira-valory (approved):
### Review round 1 — ✅ Approve

**What this does.** The mech's off-chain HTTP server now tells requesters where the operator's terms live. The 402 payment challenge gets a `termsUrl` field in its JSON body plus a `Link: <url>; rel="terms-of-service"` header, and the 200 accept response gets the same header. The URL comes from a new `task_execution` param, `mech_terms_url`, which defaults in skill.yaml to the Valory Mech Terms page. Setting it blank turns both off.

Checked at HEAD: both call sites keep the senders' newline-terminated `extra_headers` contract (`_send_ok_response` / `_send_rejection_response` still raise on a missing trailing newline, and `_build_terms_link_header` returns `""` or a `\n`-terminated line). The 402 builder call stays inside the existing try/except that answers 500, and on the 200 path the new builder cannot raise, so the enqueued-then-500 window is no wider than before. The 402 body is persisted with `record_response=True`, so `termsUrl` also shows up through the polling endpoint. `rel="terms-of-service"` is the IANA-registered relation from RFC 6903.

Config parity: neither `agents/mech/aea-config.yaml` nor `services/mech/service.yaml` overrides `mech_terms_url`, so the skill default applies after the agent/service hash bump, as the PR body says. Because no `${MECH_TERMS_URL:str:...}` override exists, an operator who wants different terms has to add a line to their service.yaml (`use_offchain` works the same way today). CI: 22/22 checks green at 498e1ff.

Tests cover the header and body on both 402 and 200, the blank/whitespace path (no `Link`, no `termsUrl`), and the Params stripping cases, including `None` and a missing key.

**Findings** — 1 · ⚪ 1 Nit

| F1 | ⚪ NIT | [packages/valory/skills/task_execution/models.py:153](https://github.com/valory-xyz/mech/blob/498e1ffc9fd3fccd5d88f2f22d62086f1c534e90/packages/valory/skills/task_execution/models.py#L153) | **mech_terms_url only strips the ends; inner CR/LF breaks header framing** |

Adamantios (approved):

PR #496, opened 2026-09-21
docs: add the Mech Terms and off-chain sections, document SERVICE_ENDPOINT_BASE

DIvyaNautiyal07:
## Proposed changes

README only.

- **New Terms section.** It explains three things:
  - Requests to Mechs operated by Valory fall under Valory AG's Mech Terms, with the notice in the wording legal approved.
  - How to check whether Valory operates a Mech: resolve its name under `mech.valory.xyz`, which is the Mech address without `0x`, a hyphen, then the chain id. There is a `dig` example.
  - Where a Mech gives its operator's terms link: `termsUrl` in its metadata, and on off-chain responses, the `Link: rel="terms-of-service"` header and `termsUrl` in the 402 body. These come from #495. The section also tells operators running their own Mech to set `mech_terms_url`, which defaults to Valory's Mech Terms.
- **New Off-chain requests section.** The README had nothing on the off-chain path. The new section covers:
  - How to turn it on: `use_offchain`, `SERVICE_ENDPOINT_BASE`, the metadata `url`, and sending traffic to one agent instance.
  - The endpoints, the fields a request carries and the four checks the Mech makes.
  - Every response status, including the `402` deposit challenge and the `Payment-Receipt` header.
  - How requesters pay, and how to read a result.
  - How delivered requests are settled with `deliverMarketplaceWithSignatures`, and that failed ones are never charged.
  - The `mech_offchain_` metrics.
- **`SERVICE_ENDPOINT_BASE` added to the environment variables table.** The mech's HTTP server answers only requests addressed to this host, a Propel host or localhost.

Nothing is stated about the terms of Mechs that Valory does not operate.

## Fixes

Part of the Mech Terms engineering work (identification and notice).

## Types of changes

- [x] Non-breaking fix (non-breaking change which fixes an issue)

## Checklist

- [x] Lint and unit tests pass locally with my changes (docs only; `tomte check-doc-links` and `check-doc-hashes` pass)
- [ ] I have added tests that prove my fix is effective or that my feature works (n/a)

## Further comments

The approved notice sits in a text block because the repo's link checker reads the sentence's final period as part of the URL.


README.md
```diff
@@ -46,6 +46,80 @@ The project consists of three components:
...
+```
+
+The zone has no wildcard record, so a Mech that Valory does not operate has no name there.
+
+**Publishing your terms.** A Mech gives its operator's terms link in two places:
+
+- The `termsUrl` field of the Mech's metadata.
+- Its off-chain responses. An accepted request and a payment-required (402) response both carry a `Link` header with `rel="terms-of-service"`, and the 402 body carries `termsUrl` next to the deposit instructions.
+
+The off-chain link comes from the `mech_terms_url` parameter of the `task_execution` skill, which defaults to Valory's Mech Terms. If you run your own Mech, set `mech_terms_url` to your own terms, or leave it empty to send no link.
```
OjusWiZard: **This contradicts the code, in the direction that matters most.**

`task_execution/models.py:198` defaults the parameter to empty:

```python
self.mech_terms_url: str = str(kwargs.get("mech_terms_url", "") or "").strip()
```

and two tests assert exactly that — `test_params_mech_terms_url_is_stripped_and_defaults_to_empty` and `test_params_mech_terms_url_absent_means_none_advertised` (`== ""`). The Valory URL is the *packaged* `skill.yaml:108` value, not the code default.

That distinction is load-bearing: an operator who deploys the packaged skill unmodified advertises **Valory's terms as their own**, while one who overrides the config and omits the key advertises none. "Defaults to Valory's Mech Terms" reads like a benign fallback and hides the first case. Suggest saying the packaged default is Valory's URL, so a non-Valory operator **must** change it — which is also the stronger reason to read the sentence that follows.

README.md
```diff
@@ -46,6 +46,80 @@ The project consists of three components:
...
 _Note that Mechs which were deployed before the Mech Marketplace contracts (called legacy Mechs) receive request and deliver services directly via their Mech contract._
 
+## Off-chain requests
+
+Besides on-chain requests through the Mech Marketplace, a Mech can take requests over HTTP. The requester signs the request instead of sending a transaction, and pays from a balance deposited in advance. The Mech settles each delivered request on-chain afterwards. [mech-client](https://github.com/valory-xyz/mech-client) handles the client side. This section describes what the Mech does.
+
+**Turning it on.** The off-chain path is off by default. To turn it on:
+
+- Set the `use_offchain` parameter to `true`. Services built on this Mech, such as mech-predict, expose it as the `USE_OFFCHAIN` environment variable.
+- Serve the Mech at a public URL and set `SERVICE_ENDPOINT_BASE` to it. The HTTP server only answers requests addressed to that host, a Propel host or localhost.
```
OjusWiZard: This states a guarantee the code doesn't quite provide. `mech_abci/handlers.py:116-128` builds:

```python
hostname_regex = rf".*({service_endpoint_base}|{propel_uri_base_hostname}|localhost|127.0.0.1|0.0.0.0)(:\d+)?"
```

Two gaps: it's unanchored on the left, so a host merely *containing* the configured name or `localhost` matches; and `service_endpoint_base` is interpolated unescaped, so the dots in the hostname act as regex wildcards.

The code is pre-existing — this PR is just the first place it's described as a restriction, which is why it's worth wording carefully. "Answers requests whose host matches..." would be accurate without promising more than the regex delivers.

README.md
```diff
@@ -46,6 +46,80 @@ The project consists of three components:
...
+- `ipfs_data` hashes to `ipfs_hash`.
+- `nonce` is the requester's next nonce. At most 64 requests per requester can be in flight at once.
+- The requester's prepaid balance in the Mech's balance tracker covers `delivery_rate`.
+
+| Status | Meaning |
+| ------ | ------- |
+| `200`  | Accepted. The `Payment-Receipt` header carries the accepted amount, and settlement is `pending`. Sending an accepted request again returns `200` with `"already accepted"`. |
+| `400`  | The request is malformed. |
+| `401`  | The signature is invalid, or the nonce was already used. |
+| `402`  | The balance is too low. The `WWW-Authenticate: Payment scheme="olas-prepay"` header and the body say how to deposit: `payTo`, `asset`, `chainId`, `currentBalance`, `required`, `depositInstructions` and `termsUrl`. |
```
OjusWiZard: The field list reads as exhaustive but omits two that are always present. `_build_402_challenge` (`task_execution/handlers.py:2704-2722`) also sends `scheme` and `error`, both pinned by tests (`test_handlers.py:803` asserts `payload["scheme"] == "olas-prepay"`, `:813` asserts `payload["error"]`).

Conversely `termsUrl` is *conditional* — it's only added when `mech_terms_url` is set, which the line-121 note implies but this row presents as unconditional.

Either add `scheme` and `error` and mark `termsUrl` optional, or soften to "includes ... among other fields".

README.md
```diff
@@ -46,6 +46,80 @@ The project consists of three components:
...
+| `503`  | Try again later. The off-chain path is off, the nonce is ahead of the next one, the requester has too many requests in flight, or a chain read failed. |
+
+Accepted and `402` responses also carry the operator's terms link. See [Terms](#terms).
+
+**Paying.** The requester deposits by calling `depositFor(requester, amount)` on the balance tracker given as `payTo` in the `402` response. Accepting a request charges nothing. The balance is charged when the Mech settles the delivered request on-chain.
+
+**Reading the result.** Call `/fetch_offchain_info` with a form-encoded `request_id`. The answer is always `200`:
```
OjusWiZard: Not always `200`. `_handle_offchain_request_info` (`task_execution/handlers.py:2457-2465`) wraps the body parse in `try/except` and routes a missing or malformed `request_id` — or an oversize body past `MAX_HTTP_BODY_BYTES` — to `_handle_bad_request`, which replies `400`. There's a test named for it: `test_handlers.py:1660`, "returns 400 when request_id is absent from body".

Suggest "for a well-formed request, the answer is always `200`", or list `400` alongside.

README.md
```diff
@@ -46,6 +46,80 @@ The project consists of three components:
...
+
+Each Mech is run by its own operator.
+
+**Mechs operated by Valory.** Requests to a Mech operated by Valory fall under Valory AG's Mech Terms. The notice for these Mechs reads:
+
+```text
+By submitting a request to this Mech, you agree to be bound by Valory AG's Mech Terms (v1.0), available at https://www.valory.xyz/terms/mechs.
+```
+
+**Checking who operates a Mech.** Valory creates one DNS record under `mech.valory.xyz` for each Mech it operates. The name is the Mech address without `0x`, a hyphen, then the chain id. If the name resolves, the Mech is operated by Valory:
```
OjusWiZard: "If the name resolves, the Mech is operated by Valory" is stated as proof, and this section omits the reader-side check that mech-interact#118's version of the same text includes: if a name you make up at random also resolves, the zone or your resolver is answering everything and the result tells you nothing.

Since the two READMEs document the same mechanism, they should agree — and this is the more authoritative of the two. Worth adding that caveat here (line 114's "the zone has no wildcard record" covers the zone side, but not a resolver that rewrites unknown names).

README.md
```diff
@@ -137,6 +211,7 @@ You may customize the agent's behaviour by setting these environment variables.
...
 | `MECH_TO_CONFIG`           | `dict` | `{"0xFf82123dFB52ab75C417195c5fDB87630145ae81":{"use_dynamic_pricing":false,"is_marketplace_mech":false}}`                                                                                                                                                          | Tracks mech's config.                                                  |
 | `PROFIT_SPLIT_BALANCE`     | `int`  | 1000000000000000000                                                                                                                                                                                                                                                 | Minimun mech balance to trigger the profit split functionality.        |
+| `SERVICE_ENDPOINT_BASE`    | `str`  | `"https://my-mech.example.com"`                                                                                                                                                                                                                                     | Public URL of the Mech. Its HTTP server answers only requests addressed to this host, a Propel host or localhost. |
```
OjusWiZard: Two issues in this row.

The **Default** column shows `"https://my-mech.example.com"`, which is an example rather than a default — the packaged value is `https://dummy_service.autonolas.tech/` (`mech_abci/skill.yaml:152`, and the same in `task_submission_abci` and `delivery_rate_abci`). Every other row in this table shows a real default.

Second, `SERVICE_ENDPOINT_BASE` appears nowhere in this repo outside the README — the parameter is `service_endpoint_base`. If the env var is supplied by the service package's override rather than read here, that's fine, but worth confirming the exact spelling since a reader will set it verbatim.

(The host-matching wording here has the same issue as line 56.)

OjusWiZard (commented):
Documentation-only, so I reviewed it as factual verification: every new claim checked against the code on this branch. Most of it holds up — inline comments mark the six that don't.

**Verified correct**, so you know what was actually checked: the endpoint table (paths and GET/HEAD vs POST wiring), the request field list and form-encoding, all four acceptance checks including `MAX_ACCEPTED_PER_SENDER = 64`, the status-code table including the duplicate-`200` "already accepted" path and the `WWW-Authenticate: Payment scheme="olas-prepay"` header, the `Payment-Receipt` header, all four `/fetch_offchain_info` response shapes, settlement via `deliverMarketplaceWithSignatures` one-call-per-requester in the Safe multisend with `MAX_SETTLEMENT_ATTEMPTS = 3`, `depositFor(requester, amount)`, the `mech_offchain_`-prefixed metric set, the `Link: rel="terms-of-service"` header on both `200` and `402`, the multi-agent startup warning, and `use_offchain` defaulting off in both code and packaged `skill.yaml`.

I also ran the `dig` example this README gives: it resolves (`34.147.134.129`), and a random name in the zone returns nothing — so the no-wildcard claim holds live.

Two things outside this repo: the `USE_OFFCHAIN` env-var claim is about mech-predict and isn't checkable here (nothing contradicts it). And the quoted Valory notice is byte-identical to mech-client#260 and valory-xyz/connect#73 — good. Worth knowing that valory-xyz/mech-interact#118 ships a *different* wording of the same notice; I've raised there whether that variant carries its own approval.

README.md
```diff
@@ -46,6 +46,80 @@ The project consists of three components:
...
+By submitting a request to this Mech, you agree to be bound by Valory AG's Mech Terms (v1.0), available at https://www.valory.xyz/terms/mechs.
+```
+
+**Checking who operates a Mech.** Valory creates one DNS record under `mech.valory.xyz` for each Mech it operates. The name is the Mech address without `0x`, a hyphen, then the chain id. If the name resolves, the Mech is operated by Valory:
+
+```bash
+dig +short c05e7412439bd7e91730a6880e18d5d5873f632c-100.mech.valory.xyz
+```
+
+The zone has no wildcard record, so a Mech that Valory does not operate has no name there.
```
bennyjo: I think we should say just:
"If the name does not resolve, the Mech is not operated by Valory.

bennyjo (commented):
Quick comment on phrasing.

README.md
DIvyaNautiyal07: Fixed in the latest commit.

README.md
DIvyaNautiyal07: Fixed in the latest commit.

README.md
DIvyaNautiyal07: Fixed in the latest commit.

README.md
DIvyaNautiyal07: Fixed in the latest commit.

README.md
DIvyaNautiyal07: Fixed in the latest commit.

README.md
DIvyaNautiyal07: That column is Sample Value, so the example is intended. `SERVICE_ENDPOINT_BASE` is the env var `packages/valory/services/mech/service.yaml` maps to `service_endpoint_base`. The host-matching wording is fixed.

README.md
DIvyaNautiyal07: Added.

bennyjo (approved):
Matches the merged code and the terms. The failing link check is GitHub returning 504s for two pre-existing links; a re-run should clear it.

README.md
```diff
@@ -46,6 +46,80 @@ The project consists of three components:
...
+By submitting a request to this Mech, you agree to be bound by Valory AG's Mech Terms (v1.0), available at https://www.valory.xyz/terms/mechs.
+```
+
+**Checking who operates a Mech.** Valory creates one DNS record under `mech.valory.xyz` for each Mech it operates, and for no other Mech. The name is the Mech address without `0x`, a hyphen, then the chain id. If the name resolves to an address in the public DNS, the Mech is operated by Valory:
+
+```bash
+dig +short c05e7412439bd7e91730a6880e18d5d5873f632c-100.mech.valory.xyz
+```
+
+If the name does not resolve, the Mech is not operated by Valory. The zone has no wildcard record, so a Mech that Valory does not operate has no name there. If a name you make up at random also resolves, your resolver is answering every name and the result tells you nothing. The lookup is not DNSSEC-validated, so it trusts your resolver to return the public answer.
```
OjusWiZard, line 114: One gap left, and it's the mirror of what I raised on mech-interact#118: "If the name does not resolve, the Mech is not operated by Valory" treats a name that doesn't exist and a lookup that failed as the same answer. A resolver outage, a timeout or no network all produce "does not resolve" without saying anything about who operates the Mech — and this section is addressed to a human running `dig`, who is exactly the reader who'll hit that case and draw the wrong conclusion.

Both sibling docs already distinguish it. mech-interact#118's README:

> A lookup that fails outright, with no network or an unreachable resolver, tells you nothing either way, so try again.

and mech-client's:

> A lookup that fails or times out is also a no, and `mechx` logs a warning so it is not mistaken for a real answer.

One sentence after this line settles it. Worth doing here specifically because this is the most authoritative of the three docs and currently the least precise — mech-client carries all three caveats, and this one and mech-interact are each missing a different one.

OjusWiZard (approved):
Approving.

**First, a correction to my own review: my `SERVICE_ENDPOINT_BASE` comment was wrong on both counts, and you were right to push back.** The column is headed "Sample Value" (README.md:204), not Default, so the example is exactly what belongs there — and the env var is real: `packages/valory/services/mech/service.yaml:53,102,138` maps `service_endpoint_base: ${SERVICE_ENDPOINT_BASE:str:https://dummy_service.autonolas.tech/}`. My earlier grep for it outside the README came back empty and I trusted that instead of checking the service package. Sorry for the noise.

The other five all check out against the code:

- **`mech_terms_url`** is now stated the way that matters — the packaged `skill.yaml` sets Valory's URL, so a Mech run by anyone else **must change it**. That's stronger than what I asked for, and it's the framing that actually prevents a third-party operator advertising Valory's terms by accident.
- **The host claim** became "answers requests whose host matches", which is honest about a regex match rather than promising an exact restriction.
- **The 402 body** now lists `scheme` and `error`, with `termsUrl` marked conditional.
- **`/fetch_offchain_info`** documents the `400`.
- **The Terms section** gained both the random-name caveat and the resolver-trust note.

One inline note below, not blocking.

Also worth knowing: CI was red earlier on the doc-link check, but it was two GitHub `504`s against `mech-tools-dev` and `mech-server` — links that already exist on `main` and that this PR never touches. Both return `200` on a direct check, and the re-run went green, so there was never anything here to fix.

PR #497, opened 2026-09-22
feat(task_execution): replay off-chain follow-up steps from the preimage store after a restart

DIvyaNautiyal07:
## Summary

Off-chain requests have three follow-up steps after the tool answer is computed: the batched on-chain deliver, serving the answer on `/fetch_offchain_info`, and the predict-api row. All three ran from in-process state only, so a restart between the answer and the last step left them unfinished with nothing to resume from. The durable preimage buffer (#455) was write-only, and its connection wiring had been dropped from the agent config by the #457 merge.

This makes the preimage store the source of truth for those steps and adds a drainer that finishes them after a restart. Design notes: the "Final plan" section of the internal kv-store durability note (not in the repo).

## What changes

**Agent config**
- Restore the `valory/kv_store` connection, protocol and `store_path` override (`/data/mech_preimages`) that #457 dropped.
- Declare the preimage params. `preimage_retention_enabled` now defaults to **true**; an env override still opts a mech out.
- `service.yaml` mounts `./data` at `/data`, gives each agent its own `store_path` under it (`STORE_PATH_0` to `STORE_PATH_3`), and aea-config exposes `STORE_PATH`. Keys are also namespaced by the agent's own address, so agents on one shared volume never read, replay or prune each other's rows.

**Preimage record** (`task_execution/utils/preimage.py`)
- Carries the consensus-ready `done_task` and the built predict-api event, JSON-detached from the live dicts.
- Per-step stamps: `settled_tx_hash`, `fetched_at` (visibility only), `posted_at`, `abandoned_at`. A delivered row is *complete* once settled and, when the predict-api write is actually configured (the settlement skill publishes `use_offchain` + a non-empty events URL into shared state at setup), posted. Rejected, never-answered and abandoned rows are complete by construction, and so is a delivered row with nothing left to replay (no `done_task` while unsettled, no event while unposted), which covers rows written before this change.
- Stamps only target this agent's off-chain ids. Stamps for a row not in memory are parked, timestamped, applied when the sweep hydrates the row, and pruned past the retention window if no row ever claims them.

**Retention**
- Complete rows expire on the 24h window as before.
- Incomplete delivered rows are held up to `preimage_incomplete_cap_seconds` (default 7 days) and deleted past it with a WARNING.
- `mech_preimage_rows` / `mech_preimage_incomplete_rows` gauges published on every full sweep.

**Drainer** (`KvStoreHandler`, on every LIST page of the sweep: first tick after boot, then hourly)
- Restores answered rows from the last 24h into the in-memory fetch map (never overwrites a live entry).
- Re-queues unsettled rows into `DONE_TASKS` with the wire nonce re-registered in the settling set; `TaskPoolingRound` already de-duplicates by request id, so a task the FSM still holds settles once.
- Restores their predict-api events so the normal post-settlement POST picks them up.
- Rows settled before the restart but never posted are re-sent by `PostTxSettlementBehaviour` as a third, isolated `replay` batch carrying their own settlement tx hash, at most `predict_api_replay_batch_size` (50) per round, oldest first. A 4xx on that batch shrinks the next replay to one event, and a 4xx on a one-event replay marks that row abandoned, so a single refused row cannot block newer ones. In-memory rows past the incomplete cap are evicted at the end of each sweep.

**Stamps from the settlement skill**
- `count_settled_from_synced_data` stamps the tx hash in the round that confirms the tx, the tx-prep retry-cap drop stamps abandoned, a 2xx on the delivered or replay batch stamps posted.

**Drain loop**
- Writes flush ahead of sweep deletes so settlement-critical state is never held behind a delete backlog.
- No-reply circuit breaker: after `preimage_max_write_attempts` consecutive watchdog timeouts with no kv_store reply ever seen, retention is switched off for the process with a WARNING and the queues are cleared, so an agent whose config lacks the connection cannot grow the write queue. Any reply, including one the late-reply guard discards, counts as proof the store is wired. Reads, writes and stamps are enqueue-only on the agent thread; SQLite work stays in the connection's thread pool, so the request path does not get slower.

## Not covered

- A restart between the deliver tx confirming and the settled stamp reaching disk (same round now) re-queues that deliver; the simulation fails on the consumed nonce and the existing retry cap drops it, now with an abandoned stamp. A chain read before re-queue was deliberately left out.
- The replay batch only fires in `PostTxSettlementBehaviour`, so on a mech with no settlements those rows wait for the next settled round.
- Deployment: the mech-predict agent config needs the same connection wiring and volume; that repo is separate.

## Tests

- `tests/utils/test_preimage.py`: stamps, completeness, expiry with the incomplete cap, fetch payload rebuild, hydrate (retention window, parked stamps, in-memory skip, cap skip, counters), replayable events.
- `tests/test_preimage_buffer.py`: LIST page rehydration (fetch map, DONE_TASKS + nonce, event cache, settled-unposted left for replay, de-dup, detachment, gauges), SUCCESS keeps a delivered row until complete, writes-before-deletes ordering, `_buffer_replay_inputs`.
- `test_handlers.py` / `test_behaviours.py`: fetched stamp on first poll only, finalize attaches replay inputs, retention default on / opt-out.
- `task_submission_abci`: settled stamp (in-memory and parked), abandoned stamp at the cap, posted stamp on 2xx for delivered/replay only, replay batch ordering and isolation, event id resolution.

All workflow checks pass locally (lint set, packaging and ABCI checks, doc hashes, copyright, bandit, gitleaks, uv lock). `safety` and `liccheck` fail identically on `main` today (pip advisory in the tox env, cffi 2.1.1 licence classifier). 1142 tests pass across the three skills. Packages pushed with `autonomy push-all`.


packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -627,6 +671,13 @@ def handle_submitted_tasks(self) -> Generator[None, None, None]:
         if isinstance(events_by_id, dict):
             for rid in submitted_ids:
                 events_by_id.pop(rid, None)
+        # Stamp the on-chain settlement on the durable preimage record so a
+        # restart never replays a deliver that already landed (no-op unless
+        # retention is on; parked if the row is not in memory yet).
+        for rid in submitted_ids:
+            preimage_buffer.record_stamp(
```
OjusWiZard: **[important] `PREIMAGE_PENDING_STAMPS` grows forever.** This stamps every submitted id, on-chain requests included. On-chain requests never get a preimage row, so `record_stamp` parks the stamp, and only `hydrate()` (preimage.py L578) ever removes a parked entry, and only for a row that shows up in this agent's store. Result: one permanent entry per on-chain settlement, and per off-chain settlement on every agent that isn't the executor. Retention is now on by default, so this affects every mech. Only stamping `is_offchain` ids here (the `shared_done_tasks_by_id` snapshot has the flag), or putting a TTL on parked entries, would fix it.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -493,6 +501,13 @@ def note_settlement_skipped(
                 ]
         if dropped is not None:
             _discard_settling_nonce(self.context.shared_state, dropped)
+            # Mark the durable record so the drainer does not re-queue a
+            # deliver the cap already gave up on (no-op unless retention on).
+            preimage_buffer.record_stamp(
```
OjusWiZard: **[important]** Same leak as the settled stamp at L678: an on-chain task that hits the retry cap parks an `abandoned_at` that nothing ever removes. `dropped` has `is_offchain` available to gate on.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -257,22 +336,333 @@ def record_settlement(
...
+        if parked:
+            _apply_stamps(record, parked)
+        complete = is_complete(record, require_posted)
+        if not complete:
+            incomplete_rows += 1
+        if age is not None and age <= retention_seconds:
+            payload = fetch_payload(record)
+            if payload is not None:
+                fetch_payloads.append(payload)
+        if complete or request_id in records:
```
OjusWiZard: **[important] The merged stamp is dropped when it's the thing that makes the row complete.** `pending.pop` removes the parked stamp and `_apply_stamps` changes only this local `record`. If that makes `is_complete` true, this `continue` skips `enqueue_write`, so the merge never reaches disk. On the next sweep the stored row still looks incomplete, so it's loaded and replayed: the already-settled deliver retries until the retry cap abandons it, and the predict-api event is sent again. This is the normal restart path, because the FSM settles from consensus-backed state long before the hourly sweep reaches the row. Writing the row back whenever `parked` was applied, before the completeness check, would fix it.

packages/valory/skills/task_execution/tests/utils/test_preimage.py
```diff
@@ -159,3 +161,407 @@ def test_expired_keys_skips_unparseable_and_timestampless() -> None:
...
+    data = {
+        "k/done": _row("1", 10, settled_tx_hash="0x1", posted_at=1),
+        "k/stuck": _row("2", CAP + 1),
+    }
+    _, replay = _hydrate(state, data)
+    assert replay == []
+    assert state[preimage.PREIMAGE_RECORDS] == {}
+
+
+def test_hydrate_applies_parked_stamps_before_deciding() -> None:
```
OjusWiZard, line 570: **[important]** This only checks the row that stays incomplete ("2"). Row "1", the one the parked stamp completes, is where the merge is lost (see the comment on preimage.py L588). Asserting that it is written back would catch that.

packages/valory/skills/task_submission_abci/tests/test_behaviours.py
```diff
@@ -4637,3 +4637,279 @@ def test_complementary_service_metadata_get_update_hash_tx_data_returns_bytes(
...
+        b = self._make_b()
+        _delivered_preimage_state(b.context.shared_state)
+        self._run(b, [STAMP_REQUEST_ID])
+        record = b.context.shared_state[preimage.PREIMAGE_RECORDS][STAMP_REQUEST_ID]
+        assert record[preimage.FIELD_SETTLED_TX_HASH] == STAMP_TX_HASH
+        assert b.context.shared_state[preimage.PREIMAGE_WRITE_QUEUE] == [
+            STAMP_REQUEST_ID
+        ]
+
+    def test_unknown_id_is_parked_for_hydrate(self) -> None:
```
OjusWiZard: **[minor]** This only checks that an unknown id gets parked. It doesn't cover the on-chain id that is never hydrated and so stays parked forever (see the comment on behaviours.py L678).

packages/valory/skills/task_execution/handlers.py
```diff
@@ -997,10 +1015,14 @@ def handle(self, message: Message) -> None:
 
         if performative == KvStoreMessage.Performative.LIST_RESPONSE:
             now = time.time()
+            page = dict(kv_msg.data)
+            require_posted = bool(self.params.use_offchain)
```
OjusWiZard: **[important]** `posted_at` is only ever stamped when `predict_api_events_url` is set, but this makes it required whenever `use_offchain` is on. `task_submission_abci/behaviours.py` L2452 already warns about `use_offchain` with an empty URL, so that setup is known to happen. In it, every delivered row is permanently incomplete: held for 7 days instead of 24h, the incomplete-rows gauge stays inflated, and each row is deleted with the "follow-up step still missing" warning. `require_posted` should reflect whether the predict-api write is actually configured.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -257,22 +336,333 @@ def record_settlement(
...
+    :param record: the parsed preimage record.
+    :param require_posted: whether the predict-api write is enabled, i.e.
+        whether ``posted_at`` is required for completeness.
+    :return: ``True`` when nothing is left to replay.
+    """
+    if record.get("settlement_status") != STATUS_DELIVERED:
+        return True
+    if record.get(FIELD_ABANDONED_AT) is not None:
+        return True
+    if record.get(FIELD_SETTLED_TX_HASH) is None:
```
OjusWiZard, line 503: **[important]** Delivered rows written before this PR (#455 format) have no `settled_tx_hash`, so after the upgrade they all count as incomplete. Replay skips them safely, since they have no `done_task`. But they are held for 7 days, they inflate `mech_preimage_incomplete_rows` right after the deploy, and each is deleted with a misleading warning. Treating a delivered row with no replay inputs as complete would avoid this.

packages/valory/agents/mech/aea-config.yaml
```diff
@@ -280,3 +286,9 @@ config:
   target_skill_id: valory/mech_abci:0.1.0
   port: ${int:8000}
 is_abstract: false
+---
+public_id: valory/kv_store:0.1.0
+type: connection
+config:
+  store_path: ${str:/data/mech_preimages}
```
OjusWiZard: **[important]** Retention is now on by default, but `services/mech/service.yaml` declares no `deployment.agent.volumes` for `/data` and has no `valory/kv_store` override like the other connections. On a normal redeploy that recreates the container, the store is lost with no warning or error, which defeats the point of the PR. The description lists the volume as outside this repo, but `service.yaml` is in this repo. At minimum, add a `${VAR:str:...}` override here and in `service.yaml` so operators can point it at a mounted path.

packages/valory/skills/task_execution/behaviours.py
```diff
@@ -2515,14 +2529,57 @@ def _buffer_preimage_settlement(
             time.time(),
         )
 
+    def _buffer_replay_inputs(
+        self,
+        req_id: str,
+        done_task: Dict[str, Any],
+        predict_api_event: Optional[Dict[str, Any]],
+    ) -> None:
+        """Attach JSON copies of the done_task and event to the preimage record.
```
OjusWiZard, line 2538: **[minor]** The summary line is enough; the paragraph below it is rationale.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2545,16 +2597,32 @@ def _do_predict_api_write_best_effort(self) -> Generator:
                 mech_address=mech_address,
                 verifying_contract=verifying_contract,
                 predict_api_url=predict_api_url,
-                batch_label="delivered",
+                batch_label=BATCH_LABEL_DELIVERED,
                 swept_request_ids=None,  # delivered has no queue-side state
             )
+        if replay_events:
+            # Third independent batch: rows the drainer hydrated from the
```
OjusWiZard: **[minor]** Seven lines of rationale. One line, like "rows settled before a restart, isolated like the other two batches", would do.

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -2831,6 +2912,36 @@ def _post_predict_api_batch(  # noqa: C901
             batch_label=batch_label, outcome="http_error"
         ).inc()
 
+    def _replay_events_from_preimage(self) -> List[Dict[str, Any]]:
+        """Return the drainer's settled-but-unposted events, tx hash stamped.
```
OjusWiZard, line 2942: **[minor]** The paragraph after the summary line explains the implementation in prose; it can be trimmed to the summary and `:return:`.

OjusWiZard (commented):
The design looks sound. I found no double-delivery path: `TaskPoolingRound` de-duplicates by request id, and `handle_submitted_tasks` prunes re-queued settled tasks using the consensus-backed `submitted_request_ids`. The write-before-delete ordering also holds. The inline comments are about how preimage rows are classified and a memory leak in parked stamps. I couldn't run the suite locally (it needs `autonomy packages sync`), so these come from reading the code. The `hydrate()` merge loss was also reproduced by running the function directly.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Good catch, that would have grown one entry per settlement on every agent. Fixed in b259de55: the settled stamp now only goes to off-chain ids this agent executed, read from the local snapshot first and the synced done_tasks as fallback for the post-restart case. Parked entries are also timestamped now and pruned at the end of each sweep once they are older than the retention window, so anything that still slips through is bounded. Added tests for the on-chain id and the other-agent id, both park nothing.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Fixed, the abandoned stamp is gated on the dropped task's is_offchain. Test added for the on-chain drop.

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: You're right, and thanks for reproducing it. hydrate now loads the row and queues the write whenever a parked stamp changed it, before the completeness check, so a stamp that completes the row still reaches disk. The SUCCESS handler then pops it as a complete record. The parked test now asserts row 1 is written back with the merged stamp, and there is a second test that a following sweep leaves it alone.

packages/valory/skills/task_execution/tests/utils/test_preimage.py
DIvyaNautiyal07: Extended, it now asserts row 1 is written back with posted_at merged and that both parked entries are cleared.

packages/valory/skills/task_submission_abci/tests/test_behaviours.py
DIvyaNautiyal07: Rewritten. The parked case now uses an off-chain id this agent executed that is only present in the synced done_tasks, which is the restart shape, and a parametrized test covers the on-chain id and another agent's off-chain id, neither of which parks anything.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Agreed. task_submission_abci's SharedState.setup now publishes whether the write is actually configured (its use_offchain plus a non-empty events URL) into shared_state, and the sweeper and the SUCCESS handler read that instead of task_execution's use_offchain. Falls back to the old flag only before that setup has run. Tests for both directions.

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: Fixed. is_complete now treats an unsettled delivered row with no done_task as complete, and a settled row with no event as complete, so anything written before this PR retires on the normal window and never inflates the gauge. Test added for the legacy shape.

packages/valory/agents/mech/aea-config.yaml
DIvyaNautiyal07: Done. service.yaml now mounts ./data at /data and carries the kv_store override with STORE_PATH, and aea-config uses the same STORE_PATH variable, matching how optimus does it.

packages/valory/skills/task_execution/behaviours.py
DIvyaNautiyal07: Trimmed to the summary line.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Trimmed to one line.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Trimmed to the summary and the return.

DIvyaNautiyal07:
Thanks for the review. All eight important points and the three trims are addressed in b259de55, replies on each thread. The one behaviour change worth a second look: rows with nothing left to replay now count as complete, so legacy rows retire on the normal window instead of sitting at the cap. Packages re-pushed, CI running.

packages/valory/services/mech/service.yaml
```diff
@@ -7,10 +7,12 @@ license: Apache-2.0
 fingerprint:
   README.md: bafybeihppqns2lwn5vehtqh2wu4tjyf76dbzxc26nlfws4sc5hcbjbalma
 fingerprint_ignore_patterns: []
-agent: valory/mech:0.1.0:bafybeibljonaa5icszus5u4k7lzyyl6vuvrymr6jt54jvpcgmjjcmkg33e
+agent: valory/mech:0.1.0:bafybeiea327tqjt5v4yydytmyzoc3bfi5tnrqqymidtvwz5bu6auclrtki
 number_of_agents: 4
 deployment:
   agent:
+    volumes:
+      ./data: /data
```
bennyjo: All 4 agents mount this same `./data`, so they share one preimage database and each agent's sweep re-queues and re-sends the others' unfinished work; could the store path or key prefix be per agent?

bennyjo (commented):
A few findings on the restart drainer.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -257,22 +345,394 @@ def record_settlement(
...
+    shared_state[PREIMAGE_SWEEP_ROW_COUNT] = (
+        shared_state.get(PREIMAGE_SWEEP_ROW_COUNT, 0) + rows
+    )
+    shared_state[PREIMAGE_SWEEP_INCOMPLETE_COUNT] = (
+        shared_state.get(PREIMAGE_SWEEP_INCOMPLETE_COUNT, 0) + incomplete_rows
+    )
+    return fetch_payloads, replay
+
+
+def replayable_events(
```
bennyjo: Unposted records stay in memory and all go out in one batch, so a long predict-api outage can grow this until it never sends; could the replay go out in chunks, with records past the 7-day cap dropped from memory?

packages/valory/skills/task_submission_abci/behaviours.py
```diff
@@ -627,6 +700,15 @@ def handle_submitted_tasks(self) -> Generator[None, None, None]:
...
             for rid in submitted_ids:
                 events_by_id.pop(rid, None)
+        # Stamp the on-chain settlement on the durable preimage record so a
+        # restart never replays a deliver that already landed. Only this
+        # agent's off-chain ids have a row in this agent's store; stamping
+        # anything else would park an entry no row ever claims.
+        if self.context.shared_state.get(preimage_buffer.PREIMAGE_RETENTION_ACTIVE):
+            for rid in self._own_offchain_ids(submitted_ids, shared_snapshot):
+                preimage_buffer.record_stamp(
+                    self.context.shared_state, rid, settled_tx_hash=tx_hash
```
bennyjo: This stamp lands a full period after the delivery, so a restart in between replays a delivery that already landed; could `count_settled_from_synced_data` stamp it instead, since it already has the tx hash?

packages/valory/skills/task_execution/models.py
```diff
@@ -196,14 +196,16 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        # (request, response) pair is mirrored into the kv_store together
+        # with what a restart needs to replay the follow-up steps (on-chain
+        # settlement, requester fetch, predict-api row), and a background
+        # sweeper prunes entries past the retention window. Only meaningful
+        # alongside use_offchain (on-chain deliveries are already public on
+        # IPFS); the buffer is inert on a mech that takes no off-chain
+        # requests. Set False to run without the durable buffer.
         self.preimage_retention_enabled: bool = kwargs.get(
-            "preimage_retention_enabled", False
+            "preimage_retention_enabled", True
```
bennyjo: Default-on makes sense, but an agent without the kv_store connection will wait out the 5s timeout on every write and grow the write queue without limit; could the skill check for the connection at startup and turn retention off with a warning if it's missing?

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -257,22 +345,394 @@ def record_settlement(
...
+    if record.get("settlement_status") != STATUS_DELIVERED:
+        return True
+    if record.get(FIELD_ABANDONED_AT) is not None:
+        return True
+    if record.get(FIELD_SETTLED_TX_HASH) is None:
+        # Unsettled: replayable only with a done_task to re-queue. A row
+        # without one (older schema, or replay inputs that failed to
+        # attach) can never be settled from here, so it is retired on
+        # the normal window rather than held to the cap.
+        return record.get(FIELD_DONE_TASK) is None
```
OjusWiZard: **[minor]** This also treats a new row as complete when its `done_task` failed the JSON round-trip in `_buffer_replay_inputs` (warning logged, nothing attached). That row now retires after 24h, the same as a legacy row. Probably fine, but worth confirming it's intended; a separate marker would keep the two cases apart.

OjusWiZard (commented):
Re-review: all eleven earlier comments are addressed (off-chain gating with the synced fallback, hydrate write-back, `require_posted` from the published setting, legacy rows, volume + `STORE_PATH`, trimmed comments). Only small things left.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -257,22 +345,394 @@ def record_settlement(
...
+    if _apply_stamps(record, stamps):
+        enqueue_write(shared_state, request_id)
+
+
+def prune_parked_stamps(
+    shared_state: Dict[str, Any], now: float, max_age_seconds: float
+) -> int:
+    """Drop parked stamps older than ``max_age_seconds``.
+
+    A parked stamp is claimed by :func:`hydrate` when its row turns up in
```
OjusWiZard: **[minor]** L440-445 are rationale; the summary line covers it.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -1143,6 +1191,153 @@ def handle(self, message: Message) -> None:
...
+            self.context.shared_state.get(
+                preimage_buffer.PREDICT_API_WRITE_CONFIGURED,
+                self.params.use_offchain,
+            )
+        )
+
+    def _publish_sweep_row_counts(self) -> None:
+        """Publish the row counters of the walk that just finished, then reset.
+
+        Also drops parked stamps that no row claimed within the retention
```
OjusWiZard: **[minor]** The second sentence is rationale for the gate, which is already explained where the stamps are gated.

packages/valory/services/mech/service.yaml
DIvyaNautiyal07: Confirmed, the compose generator maps the same host directory into every agent. Two fixes pushed: keys and the sweep LIST are now namespaced by the agent's own address, so agents on a shared volume never see each other's rows, and service.yaml gives each agent its own store_path under /data (STORE_PATH_0 to STORE_PATH_3) so they don't share a SQLite file either. The namespace is in code, so it also protects a deployment that only sets one path.

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: Done. The replay batch is capped by a new predict_api_replay_batch_size (default 50), oldest settled first, so an outage drains over successive settled rounds. At the end of each sweep, in-memory rows older than the incomplete cap are evicted with a warning, mirroring what the sweeper does on disk. Tests for both.

packages/valory/skills/task_submission_abci/behaviours.py
DIvyaNautiyal07: Good idea, moved. count_settled_from_synced_data stamps it now, in the round that confirms the tx, for this agent's off-chain tasks in the included set. That also removed the handle_submitted_tasks stamp and its gating helper, since the loop there already filters by executor and is_offchain. Tests moved with it.

packages/valory/skills/task_execution/models.py
DIvyaNautiyal07: I couldn't find a clean way for a skill to ask the multiplexer which connections are registered, so it's a runtime breaker instead: the watchdog counts consecutive timeouts, the handler marks the store reachable on the first reply of any kind, and after preimage_max_write_attempts timeouts with no reply ever, retention is switched off for the process with a WARNING naming the missing connection and the queues are cleared. A store that answered once and stalls later stays on the normal per-record retry path. Covered by tests for both cases.

DIvyaNautiyal07:
Second round addressed in b139aaa5, replies on each thread: per-agent key namespace plus per-agent store paths in service.yaml, replay batch capped at 50 per round with in-memory eviction at the cap, settled stamp moved into the confirming round, and a no-reply circuit breaker that turns retention off with a warning when the kv_store connection is missing. The earlier red job was the doc-link checker hitting a pypi 503, unrelated to the diff.

packages/valory/skills/task_execution/handlers.py
```diff
@@ -994,13 +1012,21 @@ def handle(self, message: Message) -> None:
             )
             return
         performative = kv_msg.performative
+        # Any matched reply proves the connection is wired; clear the
+        # no-reply circuit breaker's counters.
+        shared_state[preimage_buffer.PREIMAGE_KV_REPLY_SEEN] = True
```
OjusWiZard: **[important]** The late-reply guard at L1006-1013 returns before this line, so a reply that arrives after its 5s watchdog never sets `PREIMAGE_KV_REPLY_SEEN`, even though it proves the connection is wired. A store whose first five ops each take more than 5s (the boot-time LIST, or requests sent while the connection is still coming up) trips `_disable_retention_unreachable_store`. Retention then stays off for the life of the process, because no further requests are sent. Setting `PREIMAGE_KV_REPLY_SEEN = True` before the guard fixes it, and a test for a late reply would cover it.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -257,22 +354,404 @@ def record_settlement(
...
+        pending.pop(request_id, None)
+        pending_at.pop(request_id, None)
+    return len(stale)
+
+
+def _apply_stamps(record: Dict[str, Any], stamps: Dict[str, Any]) -> bool:
+    """Apply write-once stamps to ``record`` in place.
+
+    :param record: the preimage record.
+    :param stamps: ``field=value`` pairs.
```
OjusWiZard: **[minor]** This and the `_disable_retention_unreachable_store` docstring explain the same breaker rationale at length; one of them can be a single line.

OjusWiZard (dismissed):
All four points from the last round are fixed: per-agent key prefix and store path, the capped and oldest-first replay batch with eviction at the cap, the settled stamp moved to `count_settled_from_synced_data`, and the missing-connection breaker. Approving, with one follow-up on the breaker below.

packages/valory/skills/task_execution/utils/preimage.py
```diff
@@ -257,22 +354,404 @@ def record_settlement(
...
+    predict-api outage drains in chunks over successive rounds rather
+    than in one ever-growing request; oldest settled rows go first.
+
+    :param shared_state: the skill's shared state dict.
+    :param limit: at most this many events; ``None`` for all of them.
+    :return: ``(request_id, event, settled_tx_hash)`` triples.
+    """
+    out: List[Tuple[str, Dict[str, Any], str]] = []
+    records = shared_state.get(PREIMAGE_RECORDS, {})
+    ordered = sorted(
```
bennyjo: If predict-api keeps rejecting the oldest 50 (say one malformed event), every round resends the same batch and newer replays wait up to 7 days; could a 4xx on the replay batch stamp or drop those rows the way the sweep batch handles its bad events?

bennyjo (commented):
Thanks for the fixes; one non-blocking question.

packages/valory/skills/task_execution/handlers.py
DIvyaNautiyal07: Right, a late reply is still proof the connection exists. The reply-seen flag and the timeout counter are now cleared before the late-reply guard, so only the op bookkeeping is skipped for a stale reply. Test added for exactly that shape: guard ignores the reply, breaker counters still reset. Pushed.

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: Trimmed the method docstring to its summary line; the rationale lives on the constant only.

packages/valory/skills/task_execution/utils/preimage.py
DIvyaNautiyal07: Good point, the batch is all-or-nothing so a 4xx can't tell us which row it refused. Pushed: a 4xx on the replay batch shrinks the next replay to a single event, and a 4xx on a single-event replay marks that row abandoned with a warning naming the request id, so it retires on the normal window and stops blocking newer rows. A 2xx puts the batch size back. 5xx and network errors stay on the plain retry path, and the delivered batch is unaffected since it is never replayed. Tests for each branch.

